PR #12, opened 2018-06-21
Add category to commands shown in palette and scope them to appropriate contexts

RMacfarlane:
Adds "GitHub Pull Requests" category to all contributed commands that appear in the command palette. Limits the command palette contributions to 

- "GitHub Pull Requests: Refresh"

when the user isn't in PR Mode, and

- "GitHub Pull Requests: Close Pull Request"
- "GitHub Pull Requests: Open Pull Request in GitHub"
- "GitHub Pull Requests: View Pull Request Description"
- "GitHub Pull Requests: Refresh"

when in PR Mode that act on the currently checked out PR.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -45,10 +45,25 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: IP
 		});
 	}));
 
-	context.subscriptions.push(vscode.commands.registerCommand('pr.close', async (pr: PRNode) => {
+	function ensurePR(pr?: PRNode | IPullRequestModel): IPullRequestModel {
+		// If the command is called from the command palette, no arguments are passed.
+		if (!pr) {
+			if (!prManager.activePullRequest) {
+				vscode.window.showErrorMessage('Unable to find current pull request.');
+				return;
+			}
+
+			return prManager.activePullRequest;
+		} else {
+			return pr instanceof PRNode ? pr.pullRequestModel : pr;
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


RMacfarlane:
I'm not sure if we should add a "Checkout Pull Request" command that's always available in the command palette or not... dealing with pagination is a bit strange within the quick pick. For a "Load more" button, it would be possible to display another quick pick with a longer list of options. But if all the PRs aren't shown by default, trying to search will only search whatever page is being shown. I tried implementing getting all PRs and displaying them in the quick pick, but there's a pretty noticeable delay between choosing the command and the quick pick showing up with the options (for the VSCode repo with our many PRs). Maybe this is fine though as most repositories will probably have fewer PRs. @rebornix thoughts?

RMacfarlane:
Here's what it looks like trying to fetch all PRs in VSCode:

pullrequests

Maybe another option to the `IPullRequestsPagingOptions` to control page size would help

RMacfarlane:
Or possibly
pullrequests_with_categories

src/commands.ts
```diff
@@ -45,10 +45,25 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: IP
 		});
 	}));
 
-	context.subscriptions.push(vscode.commands.registerCommand('pr.close', async (pr: PRNode) => {
+	function ensurePR(pr?: PRNode | IPullRequestModel): IPullRequestModel {
```
rebornix: We can probably move this function one level up

RMacfarlane:
Ok, that sounds reasonable. I'll add that separately and check in what I have now, which removes the "Checkout Pull Request" command from the palette.

Issue #93, opened 2018-07-13
Discussion: Streamlined GitHub authentication


shana:
## Goal

To streamline authentication for users connecting to GitHub or GitHub Enterprise for the purposes of using a GitHub application. Ideally, the user doesn't have to anything beyond logging into their account and authorizing access to an application (without copy pasting tokens).

## Authentication tokens and scopes:

GitHub APIs almost always require a token to access them. Some don't (because the data is public for eg.), but anonymous calls are severely rate limited, so any app using the API for public data will need a token at some point

The API has a granular permission system, and a token will have scopes attached to it, so that only those endpoints or actions that the scopes allow will be usable.

1. Git operations only need a `repo` scope
2. GitHub API operations can need other scopes - `user`, `discussion`, `gist`, etc, depending on what they're doing.

There are three ways to obtain a token:

1. The user creates a personal access token in their settings page and select the scopes.
	- Pros: The application doesn't need to be registered anywhere or handle authentication.
	- Cons: This is error prone and requires a lot of manual steps for the user. 

2. The application obtains a token on behalf of the user by doing the oauth2 web authentication flow - opening a browser to the server that will show the requested permissions for the user to authorize, and returning the token to the app
	- Pros: The user can see what permissions the application is requesting. If the user has already given this application these permissions before, the step is automatic and the user doesn't have to do anything at all
	- Cons: Requires the application to listen to a webrequest in some way, which can be problematic behind firewalls

3. The application obtains a token on behalf of the user by asking the user for their username and password, their 2FA code, and calling the GitHub OAuth rest API endpoint with this information, along with the desired scopes, and obtaining a token in return.
	- Pros: No need for external browsers or listening to web requests, everything is handled by the application
	- Cons: Doesn't work for SSO-enabled organizations

## The oauth webflow for applications

While the OAUTH Rest API is usable by any desktop app, the webflow is more complicated. The web flow requires that GitHub redirect the user to a known URI at the end of the authentication - this request includes the token in the body. For desktop applications, this is a bit tricky to pull off.

There are strategies around this:

1. Desktop applications that have their own installers can install a protocol handler on the system, and configure the redirect URI to be a particular protocol (like desktop:// or something://).
	- apps using this approach: Desktop 
	- Pros: doesn't require listening to any web requests
	3. Cons: Requires an installer with permissions to install the protocol handler. The handler can be overridden by any app at any time

[...]

PR #94, opened 2018-07-13
Hook up webflow auth during credential checks

shana:
This is a Proof of Concept of implementing #93 with the webflow + localhost approach. I didn't want to rewrite the existing `Configuration` type (and I initially did this as a separate extension), so this has a separate implementation of the configuration class that keeps the token data - which is of course not really a good way to do it, but for this prototype it's probably good enough.

I've also added validating tokens for their required scopes, and triggering a "permissions upgrade" webflow if they don't. That also happens for any tokens that are read from the git credential helper.

This is how it works:

vscode-auth-blur

Some thoughts:

- When and how should the authentication trigger? Right now it triggers automatically when the PRManager gets initialized, which means it will just happen as soon as VSCode opens and popup on a user's face before they have a chance to think.
   - Should we show a UI before we start the authentication? A dialog box of some sort perhaps?
   - If we show a UI to start the auth process in the editor, then we can skip the webpage confirmation page that currently shows up before the GitHub permissions page (see gif above), since the user is already aware that they're going through an authentication process. Should we do that?

- I'm really not partial to storing tokens in the settings, I'd rather store them in the system credential manager. I assume there's some handy node library to do that for us?

I'm really new to both typescript and node applications in general so any feedback on the code as well as the overall design of the process is very much appreciated!

## Tasks

- [x] Implement authentication for the server that the remote is pointing to
- [x] Validate existing credentials, reauthenticate if they're expired/invalid
- [x] Detect whether the server is a github instance, skip if it isn't
- [x] Show "You need to sign in to [server]" UI in VSCode
- [x] Show "You are signed in to [server]" at the end of the login flow or warning/errors in case something failed
- [x] Trigger authentication when expanding a node in the PR view
   - Right now it's triggering at extension load time
- [ ] Save credentials in system credential manager instead of the user settings file


rebornix:
Rachel already left good suggestions/comments, here are my 2 cents.

> When and how should the authentication trigger
> Should we show a UI before we start the authentication?

Having some UI before the authentication is a good idea and we can use a specific tree node in Pull Request tree view, or a statusbar item and together with a registered command to trigger the auth workflow. For example, Live Share uses a status bar item to indicate the auth status

image

The [Azure Account](https://marketplace.visualstudio.com/items?itemName=ms-vscode.azure-account) extension provides a command 

azuresignin

If we use a status bar item, then it can show authenticated user info once the auth part is done. If we show a specific login tree node in Pull Request Tree View then we don't need to anything specific, the login node will be removed and replaced with categories when the user logs in.


> If we show a UI to start the auth process in the editor, then we can skip the webpage confirmation page that currently shows up before the GitHub permissions page

If I remember correctly the oauth/permission page needs a redirect page and it's static (ignore this if I'm wrong), what if the port we use is occupied?

src/extension.ts
```diff
@@ -61,11 +29,37 @@ export async function activate(context: vscode.ExtensionContext) {
...
 	let repositoryInitialized = false;
 	let prManager: PullRequestManager;
+
 	repository.onDidRunGitStatus(async e => {
 		if (repositoryInitialized) {
 			return;
 		}
 
+		let remotes = repository.remotes.filter(remote => remote.host);
+		let remote = remotes.find(remote => remote.remoteName === 'origin');
```
rebornix: We may want to avoid relying on `origin` or probably making this a configuration (for example, `github.defaultRemoteName`). While working on VSCode code base, I use `upstream` for Microsoft/vscode and `rebornix` for my own fork, so it will fail if we check `origin`. 

It's a corner case I believe but all our code logic right now doesn't rely on `origin`.

src/authentication/configuration.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface IConfiguration extends IHostConfiguration {
+    onDidChange: vscode.Event<IConfiguration>;
+}
+
+export class Configuration implements IConfiguration {
+    username: string | undefined;
+    token: string | undefined;
+    onDidChange: vscode.Event<IConfiguration>;
+    private emitter: vscode.EventEmitter<IConfiguration>;
```
rebornix: We prefer to have private members to start with `_`.

src/github/credentials.ts
```diff
@@ -25,26 +27,49 @@ export class CredentialStore {
...
 
-		if (this._configuration.host === remote.host && this._configuration.accessToken) {
-			this._octokits[remote.url] = Octokit({});
-			this._octokits[remote.url].authenticate({
-				type: 'token',
-				token: this._configuration.accessToken
-			});
-			return this._octokits[remote.url];
+		const webflow = new WebFlow(new VSCodeAppConfiguration(), remote.host);
+		let creds = this._configuration as IHostConfiguration;
```
rebornix: It seems we only need `IHostConfiguration` in `CredentialStore`, we can unrestrict the type of `this._configuration` to be `IHostConfiguration`.

src/authentication/configuration.ts
```diff
@@ -0,0 +1,40 @@
+import * as vscode from 'vscode';
+
+export interface IAppConfiguration {
+    clientId: string | undefined;
```
rebornix: Not a big deal but our existing code is using tabs ;)

src/authentication/configuration.ts
```diff
@@ -0,0 +1,40 @@
...
+    token: string | undefined;
+    onDidChange: vscode.Event<IConfiguration>;
+    private emitter: vscode.EventEmitter<IConfiguration>;
+
+    constructor(public host: string) {
+        this.emitter = new vscode.EventEmitter<IConfiguration>();
+        this.onDidChange = this.emitter.event;
+    }
+
+    update(username: string | undefined, token: string | undefined) {
```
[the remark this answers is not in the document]
shana: Thanks for this! I agree, explicit visibility and constness markers make things much easier to navigate. I'm using tslint and tslint-config-prettier locally to guide me through the language and nag me about the things I should/should not be doing, so I can configure it to follow these rules. Is tslint something you use as a team?

src/authentication/vsConfiguration.ts
```diff
@@ -0,0 +1,97 @@
...
+
+		const clientSecret: string | undefined = config.has(CLIENT_SECRET_KEY)
+			? config.get(CLIENT_SECRET_KEY)
+			: process.env.GITHUB_VSCODE_CLIENT_SECRET;
+		super(clientId, clientSecret);
+	}
+}
+
+export class VSCodeConfiguration extends Configuration {
+	private hosts: { [key: string]: any };
```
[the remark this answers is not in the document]
shana: Does `Map` easily convert to an array?

I was struggling to find a type that could give me key access but easily turn into an array for saving into the configuration file (as that takes a simple `[]`).

What I did here is a bit of a hack - it seems the typescript `{ [key: string]: any }` syntax basically just adds an indexer to the underlying object is, so it's basically an array with an extra indexer (so I just store the index of the object to look it up in the array).

I first implemented a Dictionary type, and then discovered this key access syntax and kept two fields, one with the array of all the hosts and another with the keys and indexes to the array, but then discovered I could do everything in one object, which is what is there now. I agree it's really hard to read, I'll see if Map does the trick.

src/authentication/vsConfiguration.ts
```diff
@@ -0,0 +1,97 @@
...
+
+	constructor(public host: string) {
+		super(host);
+		this.hosts = [];
+		this.loadHosts();
+		const config = this.getHost(this.host);
+		super.update(config.username, config.token);
+	}
+
+	listenForVSCodeChanges() {
```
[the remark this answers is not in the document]
shana: Yeah, probably. Not sure where my head was on that one 🤔

shana:
>> If we show a UI to start the auth process in the editor, then we can skip the webpage confirmation page that currently shows up before the GitHub permissions page

> If I remember correctly the oauth/permission page needs a redirect page and it's static (ignore this if I'm wrong), what if the port we use is occupied?

Yeah, that's a problem I go over in #93. I think the ideal way is for the login to be handled by a webapp on a separate server instead of localhost, bypassing the need for local ports entirely - the local app talks to the server via a websocket, and the server and github talk to each other. That also mitigates the problem of distributing client secrets (as the server is the one keeping them)

src/extension.ts
shana: Yes, I agree, hardcoding `origin` is bad. Your logic right now relies on looking for a remote that is pointing to `github.com`; this logic finds either one that's called `origin` or the first remote if `origin` doesn't exist.

This means it will likely find `rebornix` for you before it finds `upstream`, if you cloned your fork first and then added the `upstream` remote. It's highly dependent on the clone/remote order. Definitely not ideal :P

On GHfVS, I think we look for remotes and then check whether the repository they point to is a fork, in which case we look up the fork parent and use that by default for PRs. I think we also let you switch between PR sources for a fork/upstream combo (i.e., you can see the PRs from upstream or you can see PRs from a fork, for long-lived project forks that have their own community)

The reason I don't want to scan for `github.com` is because github enterprise instances will work just as well with these features, and we can't really rely on looking at the name of the server. I probably need to add a check for a specific header that dotcom or enterprise always return, to make sure the server is usable (besides the whole "which remote do I use" thing)

A configuration option is probably good enough to start with to avoid these issues.

src/github/credentials.ts
shana: Oh, yeah, good catch, this was me carefully walking around your existing Configuration implementation by avoiding changing signatures. All the configuration stuff needs cleaning up :P

src/authentication/configuration.ts
shana: Darnit, I noticed that and configured prettier to use tabs but looks like spaces still snuck in. Fixicating...

shana:
>> Should we show a UI before we start the authentication?

> I think this would be a better experience, we could show a notification telling the user they need to authenticate, with a button to do so. Skipping to the GitHub permissions page would make sense then 👍

> There's an example of using the notification UI here: https://github.com/Microsoft/vscode-pull-request-github/blob/master/src/github/pullRequestManager.ts#L290

Oh awesome, thanks, I'll take a look!

> I'm also wondering if we should show another notification in VSCode after authentication. The redirect page already says the status, but it might be nice to have this in the VSCode UI as well?

Yeah, I was thinking how to redirect the user "eye" back to the editor. A notification would probably do it 👍

src/authentication/configuration.ts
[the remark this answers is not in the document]
RMacfarlane: Yup, we use tslint. In VSCode we have a pretty small set of style rules and a larger number of rules dictating what can be imported where: https://github.com/Microsoft/vscode/blob/master/tslint.json

Adding a `tslint.json` to this project sounds like a good idea to me

src/authentication/vsConfiguration.ts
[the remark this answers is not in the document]
RMacfarlane: Ah, I see! I don't think there's any method that directly gives an array primitive from the map, but there are a couple of ways to iterate through it that could be used to create an array:

```
const flattenedMap = [];
map.forEach(value => flattenedMap.push(value));
```

or

```
Array.from(map.values());
```

src/authentication/configuration.ts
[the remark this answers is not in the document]
rebornix: I can add a basic tslint config file first.

shana:
Almost all the functionality is in (:tada:), I've added a list of tasks done and not done on the issue body, which I'm tackling next. I don't think I forgot anything, if I did let me know!

Right now this is using the websocket approach, but I think that we can actually use a protocol handler with the same model and skip the web socket entirely, if it turns out that web sockets cause problems. Web sockets basically go through the same pipe as https requests, so I feel that if a web socket doesn't work, https wouldn't either and you'd be kinda stuck either way, so not sure I want to change that right now.

src/github/credentials.ts
```diff
@@ -3,48 +3,127 @@
...
+			octokit = this.createOctokit('token', creds);
 		} else {
-			const data = await fill(remote.url);
-			if (!data) {
-				return null;
+
+			// see if the system keychain has something we can use
+			const data = await fill(host);
+			if (data) {
+				const login = await server.validate(data.username, data.password);
```
rebornix: I'm using personal access token for https remotes so this line is being executed as I have username/token saved in git credentials. However this line never returns. Once I commented out this part and it fell back to the oauth login, everything works as expected.

shana: Interesting, I wonder what's going on. It might be throwing somewhere and getting swallowed. I'll try to repro locally.

shana: I *think* I fixed this, there were some very silly bugs in that code.

shana:
I think this is done! Here's how the flow looks now:

authenticationv2

The webbrowser UI is being worked on so it'll look better soon 😄 

The credentials are still being stored in the settings file, but I think I'd rather do the work of moving them to the system credential manager on a separate PR, as there might be some other complications with that and I don't want to delay this more. I'd also like to add a command to trigger the login, probably something that can be done separately.

I'm also unsure whether the sign in notification will work or if it's too subtle. It will pop up for anything that needs to call octokit (because getting an octokit instance triggers the whole credential loading thing). UX is hard!

package.json
```diff
@@ -28,26 +28,13 @@
...
-            "null"
-          ],
-          "default": null,
-          "description": "GitHub access token."
+        "github.hosts": {
+          "type": "array",
+          "default": [],
+          "description": "Host tokens",
+          "items": {
+            "type": "object"
```
RMacfarlane: I think we can also add a `"properties"` property within items, which helps intellisense in the settings editor

something like
```ts
"properties": {
	"host": {
		"type": "string",
		"description": "The host name to access GitHub."
	},
	"username": {
		"type": "string",
		"description": "The host name to access GitHub."
	},
	"token": {
		"type": "string",
		"description": "GitHub access token."
	}
}
```

src/authentication/configuration.ts
```diff
@@ -0,0 +1,53 @@
...
+export interface IHostConfiguration {
+	host: string;
+	username: string | undefined;
+	token: string | undefined;
+}
+
+export const HostHelper = class {
+	public static getApiHost(host: IHostConfiguration | vscode.Uri): vscode.Uri {
+		const hostUri: vscode.Uri = host instanceof vscode.Uri ? host : vscode.Uri.parse(host.host);
+		if (hostUri.authority === 'github.com') {
```
RMacfarlane: is it possible for there to be a case mismatch?

src/github/pullRequestManager.ts
```diff
@@ -14,7 +14,8 @@ import { IPullRequestManager, IPullRequestModel, IPullRequestsPagingOptions, PRT
 import { PullRequestGitHelper } from "./pullRequestGitHelper";
 import { PullRequestModel } from "./pullRequestModel";
 import { parserCommentDiffHunk } from "../common/diffHunk";
-import { Configuration } from "../configuration";
+import { Configuration } from '../authentication/configuration';
+import { GitHubManager } from '../authentication/githubserver';
```
RMacfarlane: `../authentication/githubServer`

package.json
shana: Oh cool, didn't know that, thanks!

src/authentication/configuration.ts
shana: hmmmmm, good question. possibly, yes, I can test that quick

src/github/pullRequestManager.ts
shana: ooof, good catch, that would fail on linux :/

shana:
I went through and normalized all the hostnames to lowercase, there are other places where there could be case mismatches. Tested that out with another repo I have with a bunch of different remotes (to which I added a few more), and it matches things correctly for all of them.

```
another-upstream        https://github.com/octokit/octokit.net.git (fetch)
another-upstream        https://github.com/octokit/octokit.net.git (push)
editor-tools    https://github.com/editor-tools/octokit.net (fetch)
editor-tools    https://github.com/editor-tools/octokit.net (push)
origin  git@github.com:github-for-unity/octokit.net (fetch)
origin  git@github.com:github-for-unity/octokit.net (push)
origin-http     https://GitHub.com/github-For-Unity/octokit.net.git (fetch)
origin-http     https://GitHub.com/github-For-Unity/octokit.net.git (push)
shana   https://github.com/shana/octokit.net (fetch)
shana   https://github.com/shana/octokit.net (push)
upstream        https://GitHub.com/Octokit/Octokit.net (fetch)
upstream        https://GitHub.com/Octokit/Octokit.net (push)
```

PR #179, opened 2018-08-13
Update checkout buttons and state

### LINKED ISSUE #176: "Checked out" is it a button or a status indicator?
image

"Checked out" is not a button but is styled exactly the same as the button next to it.

misolori:
Fixes #176

- Styled the "Checked Out" button to look more like a status indicator
- Disabled checkout buttons when clicked
- Added "Checking Out..." text while checkout procedure happens for a branch

gif

### DIFF
--- a/preview-src/index.css
+++ b/preview-src/index.css
@@ -142,6 +142,16 @@ button:hover:enabled, button:focus:enabled {
 	cursor: pointer;
 }
 
+body button.checkedOut{
+	opacity: 1 !important;
+	background-color: transparent;
+}
+
+body button svg{
+	fill: var(--vscode-button-foreground);
+	margin-right: 6px;
+}
+
 .prIcon {
 	display: flex;
 	border-radius: 10px;
--- a/preview-src/index.css
+++ b/preview-src/index.css
@@ -10,7 +10,7 @@
 	border-bottom: 1px solid var(--vscode-tab-inactiveBackground);
 }
 
-#title:empty{
+#title:empty {
 	border: none;
 }
 
@@ -35,7 +35,7 @@ body .review-comment .avatar-container {
 	margin-right: 12px;
 }
 
-body .review-comment .avatar-container img.avatar{
+body .review-comment .avatar-container img.avatar {
 	margin-right: 0;
 }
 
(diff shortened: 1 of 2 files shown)


preview-src/index.ts
```diff
@@ -131,14 +133,19 @@ function appendComment(comment: any) {
 
 function updateCheckoutButton(isCheckedOut: boolean) {
 	const checkoutButton = (<HTMLButtonElement>document.getElementById(ElementIds.Checkout));
+	const checkoutMasterButton = (<HTMLButtonElement>document.getElementById(ElementIds.CheckoutMaster));
 	checkoutButton.disabled = isCheckedOut;
-	checkoutButton.innerHTML = isCheckedOut ? `Checked Out` : `Checkout`;
+	checkoutMasterButton.disabled = false;
+	const activeIcon = '<svg class="octicon octicon-check" viewBox="0 0 12 16" version="1.1" width="12" height="16" aria-hidden="true"><path fill-rule="evenodd" d="M12 5l-8 8-4-4 1.5-1.5L4 10l6.5-6.5L12 5z"></path></svg>';
```
RMacfarlane: I think we need more handling for light themes, the icon and text color match the background on Light+

preview-src/index.css
```diff
@@ -142,6 +142,16 @@ button:hover:enabled, button:focus:enabled {
 	cursor: pointer;
 }
 
+body button.checkedOut{
+	opacity: 1 !important;
+	background-color: transparent;
+}
+
+body button svg{
```
RMacfarlane: nitpicky comment: could you add a space before the { here and above? 😄

preview-src/index.ts
rebornix: We can probably add 

```css
.vs .octicon {
	color: white;
}
.vs-dark .octicon {
	color: black;
}
```

to the index.css

misolori: Yikes, I missed that. I'll use `var(--vscode-foreground)` for the text color instead. Thanks for catching that!

preview-src/index.css
misolori: @RMacfarlane you got it, updated the rest of the css to make sure there's a space before the `{`

Issue #197, opened 2018-08-15
Startup activation takes 650ms


jrieken:
* Install extension
* Restarting VS Code
* Check the Running Extension view and/or Developer: Startup Performance
* 🐌 notice that the PR extension takes 649ms loading its code and its dependencies. (That's slower than loading *all* of ~6MB `workbench.main.js`)

jrieken:
<img width="1144" alt="screen shot 2018-08-15 at 09 41 47" src="https://user-images.githubusercontent.com/1794099/44137094-7bdda4e4-a06f-11e8-866c-554bc9bb55da.png">

rebornix:
It seems code bundling helps significantly with startup performance.

*Original*

![image](https://user-images.githubusercontent.com/876920/44174094-d55d4c00-a096-11e8-917f-77b59e04e7a5.png)

*Bundled*
![image](https://user-images.githubusercontent.com/876920/44174136-f887fb80-a096-11e8-896d-ea5c54dedd49.png)

jrieken:
👯

rebornix:
Fixed via https://github.com/Microsoft/vscode-pull-request-github/pull/213

PR #213, opened 2018-08-15
bundle extension scripts

rebornix:
Re #197 

We didn't bundle any extension code or node modules and since we don't have any module lazy loading, the extension activation is really slow.

Here is the activation time comparison between unbundled and bundled version

*Original*
image

*Bundled*
image

The activation time decreases by around 80%. Moreover, the extension vsix size before bundling is 7.6MB while the bundled one is 3MB, which is almost half of the original size.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,7 +15,7 @@
 				"${workspaceFolder}/media/*.js"
 			],
 			"sourceMapPathOverrides": {
-				"webpack:///./*": "${workspaceRoot}\\*"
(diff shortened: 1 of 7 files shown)


RMacfarlane:
I ran into a couple of issues when trying to debug the extension. After checking this out I ran `npm install` and then the build task, and tried to debug. The break points I set didn't seem to actually get loaded, and the overview page also didn't load while debugging. Have you run into these? Is there anything I missed in the setup?

rebornix:
@RMacfarlane I pushed a change to ensure that debugging works on both Windows and macOS.

Issue #109, opened 2018-07-31
Feature Request: Add ways to Approve/Reject a PR


ramya-rao-a:
The web view for the PR can use a button to submit the review. Either Approve/Reject/Request changes

PR #224, opened 2018-08-16
Add a way to approve/request changes on a PR

RMacfarlane:
Fixes https://github.com/Microsoft/vscode-pull-request-github/issues/109

Adds buttons at the bottom of the PR for "Request Changes" and "Approve":
screen shot 2018-08-16 at 11 15 06 am

Let me know what you think of this placement - I'm not sure where we will eventually put a "Merge" button. "Request Changes" and "Comment" are the only actions that require text input, for the others, including a message is optional.

### DIFF
--- a/preview-src/index.css
+++ b/preview-src/index.css
@@ -284,6 +284,10 @@ body .comment-form textarea {
 	font-family: var(--vscode-editor-font-family);
 }
 
+.reply-button {
+	margin-left: auto;
+}
+
 body .comment-form textarea:focus {
 	outline: 1px solid var(--vscode-focusBorder);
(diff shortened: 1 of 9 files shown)


RMacfarlane:
I thought about grouping the buttons that way. I think I prefer the current grouping better - the buttons on the left have some kind of state change associated with them, so I think they make sense together.

A secondary button style sounds like a good idea to me. I created this to track it: https://github.com/Microsoft/vscode-pull-request-github/issues/232

I'm going to go ahead and merge this in so that it gets tested, but we can sync up about the button grouping tomorrow

Issue #27, opened 2018-06-26
Do not create tmp files but use the content provider


joaomoreno:
When selecting changes in the `Github Pull Requests` pane, we seem to create temp files... we can actually generate in memory documents with the content provider API.

The current approach goes south when a user clicks the `Open File` action atop the diff editor:

![image](https://user-images.githubusercontent.com/22350/41916693-e47e8344-7958-11e8-9da7-87d843bb3245.png)

alexandrudima:
ugh, this is sooooooo ugly. 

![image](https://user-images.githubusercontent.com/5047891/41966049-5e550896-79fe-11e8-97f3-38b76e2f2b30.png)

@rebornix Using a content provider is trivial and ensures you can open things sooner. i.e. I bet for now you create the `tmp` file (extracting it from the remote or from the git repository), and once you have it, you can open the diff editor. If you were to use a content provider, you could much more easily directly ask vs code to open the diff editor and then resolve the file on the fly.

PR #236, opened 2018-08-20
In memory content provider for pull request tree view.

rebornix:
Re #27 . We now move the inmem content provider for file changes nodes in both pull request tree view and Review Mode's changes in PR. The benefits are somewhat huge

* we don't need to resolve file content when expanding the PR node, we only need to check whether the `mergeBase` commit exist or not.
* we resolve file content only when they are being visited
* no local temp files

### DIFF
--- a/src/common/diffHunk.ts
+++ b/src/common/diffHunk.ts
@@ -251,7 +251,7 @@ export function getGitChangeType(status: string): GitChangeType {
 	}
 }
 
-export async function parseDiff2(reviews: any[], repository: Repository, parentCommit: string): Promise<(InMemFileChange | SlimFileChange)[]> {
+export async function parseDiff(reviews: any[], repository: Repository, parentCommit: string): Promise<(InMemFileChange | SlimFileChange)[]> {
 	let fileChanges: (InMemFileChange | SlimFileChange)[] = [];
 
 	for (let i = 0; i < reviews.length; i++) {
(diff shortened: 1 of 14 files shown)


src/common/uri.ts
```diff
@@ -101,3 +101,20 @@ export function fromFileChangeNodeUri(uri: Uri): FileChangeNodeUriParams {
 		return null;
 	}
 }
+
+export function toInMemUri(uri: Uri, pullRequestModel: IPullRequestModel, commit: string, fileName: string, base: boolean): Uri {
```
RMacfarlane: Looks like `toPRUri` is no longer being used and could be removed

src/common/diffHunk.ts
```diff
@@ -262,61 +251,25 @@ export function getGitChangeType(status: string): GitChangeType {
...
-
-			let contentFilePath = await writeTmpFile(contentArray.join('\n'), path.extname(fileName));
-			let emptyContentFilePath = await writeTmpFile('', path.extname(fileName));
-			let richFileChange = review.status === 'removed' ?
-				new RichFileChange(emptyContentFilePath, contentFilePath, gitChangeType, fileName, diffHunks, false) :
-				new RichFileChange(contentFilePath, emptyContentFilePath, gitChangeType, fileName, diffHunks, false);
-			richFileChange.blobUrl = review.blob_url;
-			fileChanges.push(richFileChange);
-		}
+		let originalFileExist = await repository.checkFileExistence(parentCommit, review.filename);
```
RMacfarlane: I think we need to be able to distinguish files changes that are additions. This sets the `isPartial` flag to true for file additions so the warning gets shown for them

src/common/uri.ts
RMacfarlane: also I think it would be good to rename `fromPRUri` and update the return type for clarity

src/common/diffHunk.ts
rebornix: Good catch.

rebornix:
@RMacfarlane already addressed your comments.

PR #269, opened 2018-08-24
Fixes #265, authentication status bar item added for each remote

### LINKED ISSUE #265: Authentication status bar item added for each remote
Is this intentional? My expectation was that there would be one status bar item overall
screen shot 2018-08-23 at 3 46 36 pm

Or one per authority

RMacfarlane:
This changes the implementation of the status bar item to be per authority instead of per remote, fixing #265

I still need to fix the status bar text when it's first created and the user's credentials are already in settings or the credential manager. Right now the behavior is that it shows "Sign into X" until `hasOctokit` is called when the user interacts with the tree.

### DIFF
--- a/src/github/credentials.ts
+++ b/src/github/credentials.ts
@@ -17,13 +17,19 @@ const TRY_AGAIN = 'Try again?';
 export class CredentialStore {
 	private _octokits: Map<string, Octokit>;
 	private _configuration: VSCodeConfiguration;
+	private _authenticationStatusBarItems: Map<string, vscode.StatusBarItem>;
+
 	constructor(configuration: any) {
 		this._configuration = configuration;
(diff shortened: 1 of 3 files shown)


shana:
> I still need to fix the status bar text when it's first created and the user's credentials are already in settings or the credential manager. Right now the behavior is that it shows "Sign into X" until hasOctokit is called when the user interacts with the tree.

`hasOctokit` can probably be called earlier in `PullRequestManager.updateRepositories` when the list of github repos is getting populated, and then we'll know if the user is "logged in" already or not

RMacfarlane:
👍 I changed it to only add the status bar on demand, when `hasOctokit` is called later, but we can add code in `PullRequestManager.udpateRepositories` for that if we get feedback that it's not intuitive

Issue #228, opened 2018-08-17
Telemetry and metrics for preview release


auchenberg:
We'd like to have a set of basic telemetry in place to measure the success and usage of the extension:

Top of mind:

**Metrics:**
Comments:
- [ ] How many comments are being added from the editor?
- [ ] How many comments are being added from the description page?

Authentication
- [ ] What's the success rate of authentication?  How many people stall?
- [ ] What's the most commons errors we get from Github? Scope issues? 

PR treeview
- [x] What PR categories are used the most? (used = expanding the tree to see PRs)
- [ ] How many times does a PR get checked out from the context menu?

Overview page
- [ ] How often is the overview page opened?
- [ ] How many times does a PR get approved?
- [ ] How many times does a PR get checked out?
- [ ] How many times does a PR get changed requested?
 
Please add more.

auchenberg:
@shana What options do we have to land this with Github tech?

RMacfarlane:
Some other possible metrics:

- [ ] How often is "Open Pull Request in GitHub" used?
- [ ] How often is the action for deleting a local pull request branch used?


- [ ] What errors do users see when trying to check out a pull request?
- [ ] What errors do users see when trying to create or reply to a comment?

shana:
We have infrastructure for submitting metrics from editor extensions and desktop apps, so we can add vscode to that. There's also a telemetry package for this purpose on atom (https://github.com/atom/telemetry) which we probably can leverage here.

auchenberg:
@shana When time allows. How many of these metrics did we get into place?

shana:
@auchenberg As of now, the following metrics are being tracked:

- PR openInGitHub
- PR open description
- PR checkout from context menu
- PR checkout from description
- PR local branch delete
- PR close
- PR request changes
- PR approve
- PR List Expand: Local, Assigned, Review Requested, Mine, All
- Authentication started (user clicked on signin to start the auth process)
- Authentication failed
- Authentication success
- Authentication cancel (user dismissed the signin info box)

These are in master but not all included in 0.22.

We can also track exceptions/errors, but those use a different endpoint/backend so they aren't included right now (but we can track that as well)

auchenberg:
Cool, I think this is a good start.

PR #330, opened 2018-08-29
Add telemetry reporting

shana:
This adds telemetry reporting (with optout) to the extension. Fixes most of https://github.com/Microsoft/vscode-pull-request-github/issues/228, not all of the data points that are listed in the issue are being reported here.

This is using a fork of the telemetry package used in Atom that doesn't require localStorage access - the source of the package is at https://github.com/shana/telemetry right now, to be eventually pulled back in to upstream.

Adding more data points is just a question of adding more overloads to the Telemetry.on method. The naming can be anything, but it is recommended that it be camelCase with enough granularity to be distinguishable from other actions that exist now or in the future (changing metrics names after they're live is annoying)


package.json
```diff
@@ -288,11 +291,12 @@
     "markdown-it": "^8.4.0",
     "markdown-it-checkbox": "^1.1.0",
     "moment": "^2.22.1",
+    "telemetry-github": "https://github.com/shana/telemetry/releases/download/0.1.1/telemetry-github-v0.1.1.tgz",
```
RMacfarlane: Running `vsce package

src/commands.ts
```diff
@@ -34,7 +34,8 @@ function ensurePR(prManager: IPullRequestManager, pr?: PRNode | IPullRequestMode
 	}
 }
 
-export function registerCommands(context: vscode.ExtensionContext, prManager: IPullRequestManager, reviewManager: ReviewManager) {
+export function registerCommands(context: vscode.ExtensionContext, prManager: IPullRequestManager,
+	reviewManager: ReviewManager, telemetry: ITelemetry) {
 	// initialize resources
```
RMacfarlane: should we also log an event on these commands? openPullRequestInGitHub, deleteLocalBranch, approve, requestChanges, openDescription?

src/common/telemetry.ts
```diff
@@ -0,0 +1,111 @@
...
+class MementoDatabase implements IStatsDatabase {
+	constructor(private readonly _context: vscode.ExtensionContext) { }
+	close(): Promise<void> {
+		return Promise.resolve();
+	}
+
+	incrementCounter(counterName: string): Promise<void> {
+		const counters = new Map<string, ICounter>();
+		(this._context.globalState.get<ICounter[]>(TELEMETRY_KEY) || []).forEach(x => counters.set(x.name, x));
+		const counter = counters.get(counterName) || { name: counterName, count: 0 };
```
RMacfarlane: is a timestamp associated with the event anywhere?

src/github/githubRepository.ts
```diff
@@ -75,6 +77,7 @@ export class GitHubRepository implements IGitHubRepository {
 		this._initialized = true;
 		if (!await this._credentialStore.hasOctokit(this.remote)) {
 			this._octokit = await this._credentialStore.login(this.remote);
```
RMacfarlane: could we also log an event for auth failures, or something at the beginning of the login process, so that we get a sense of the auth success rate?

src/commands.ts
```diff
@@ -85,6 +87,11 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: IP
 			title: `Switching to Pull Request #${pullRequestModel.prNumber}`,
 		}, async (progress, token) => {
 			await reviewManager.switch(pullRequestModel);
```
RMacfarlane: could we also log something within this method when switching fails?

package.json
RMacfarlane: my bad, I thought I had run `npm install` first but apparently hadn't 😱 This is fine.

src/common/telemetry.ts
shana: There's a timestamp for when they're sent in, but I just realized that there's a timestamp missing for the report itself, good catch, I'll fix that.

rebornix:
Merge it first for afternoon's dogfood and testing.

PR #371, opened 2018-09-06
Add more linting

RMacfarlane:
Adds some more rules to the `tslint` config and fixes instances of shadowed variables.

@rebornix @shana are there any other `tslint` rules you think we should add?

### DIFF
--- a/src/common/diffPositionMapping.ts
+++ b/src/common/diffPositionMapping.ts
@@ -91,9 +91,9 @@ export function getLastDiffLine(prPatch: string): DiffLine {
 export function getDiffLineByPosition(diffHunks: DiffHunk[], diffLineNumber: number): DiffLine {
 	for (let i = 0; i < diffHunks.length; i++) {
 		let diffHunk = diffHunks[i];
-		for (let i = 0; i < diffHunk.diffLines.length; i++) {
-			if (diffHunk.diffLines[i].positionInHunk === diffLineNumber) {
-				return diffHunk.diffLines[i];
+		for (let j = 0; j < diffHunk.diffLines.length; j++) {
+			if (diffHunk.diffLines[j].positionInHunk === diffLineNumber) {
+				return diffHunk.diffLines[j];
 			}
 		}
 	}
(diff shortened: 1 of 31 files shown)


lumaxis:
There is a nice set of additional Microsoft TSLint rules here: https://github.com/Microsoft/tslint-microsoft-contrib
They also include some recommendations.

src/common/diffPositionMapping.ts
```diff
@@ -91,9 +90,9 @@ export function getLastDiffLine(prPatch: string): DiffLine {
 export function getDiffLineByPosition(diffHunks: DiffHunk[], diffLineNumber: number): DiffLine {
 	for (let i = 0; i < diffHunks.length; i++) {
 		let diffHunk = diffHunks[i];
-		for (let i = 0; i < diffHunk.diffLines.length; i++) {
-			if (diffHunk.diffLines[i].positionInHunk === diffLineNumber) {
-				return diffHunk.diffLines[i];
+		for (let j = 0; j < diffHunk.diffLines.length; j++) {
```
IDisposable: 😲

shana:
I'm not very familiar with tslint so I'm not sure what's best, this list of rules looks to match other projects I've seen and it's pretty comprehensive.

I've seen projects extend the `tslint-config-prettier` configuration when using prettier (might be a thing we want to do to not worry about formatting at all). I've also seen projects extending `tslint:recommended`, dunno if this list matches the recommended one.

All in all, looks good! I'm just wondering about that change @IDisposable pointed out...

src/common/diffPositionMapping.ts
rebornix: Good catch! Even though `let` ensures that the inner `i` shadows the outer one, we should always avoid confusion like this.

jongio: This is a test when clicking the + sign on the line of code that has an existing comment from GitHub.com

jongio: This is a test when replying to an existing comment on GitHub.com

jongio: This is a test when replying to a comment from VS Code extension.

PR #387, opened 2018-09-10
Fixes #375: Handle PR's with more than 30 files

### LINKED ISSUE #375: Handle file change pagination for PRs with more than 30 changed files
- Extension version: 0.1.1

The default page size of results returned from GitHub is 30. We need to add a `Load more` button or automatically fetch additional pages of file changes for PRs that have more than 30 file changes

hartra344:


src/github/pullRequestManager.ts
```diff
@@ -435,12 +435,20 @@ export class PullRequestManager implements IPullRequestManager {
...
 			owner: remote.owner,
 			repo: remote.repositoryName,
-			number: pullRequest.prNumber
+			number: pullRequest.prNumber,
+			per_page: 100
 		});
-
+		let {data} = response;
+		let linkHeader = response.headers.link;
+		while(octokit.hasNextPage(linkHeader)){
```
RMacfarlane: I think `linkHeader` is undefined when there are no additional pages. Can we add a check against that in the `while` before `hasNextPage`?

hartra344: Good catch

PR #448, opened 2018-09-12
Fix name reference in README.md

ruyadorno:
I was following the "How to get started" instructions from the `README` file and it was really misleading the fact that the name won't match the exact name used in the marketplace.


RMacfarlane:
Thanks!

Issue #429, opened 2018-09-11
Refresh after a file change is unnecessary


ramya-rao-a:
Issue Type: <b>Bug</b>

- Expand one of the first level nodes like "All"
- Make a change to any file in the workspace
- Observe that the expanded first level node gets refreshed

Changes in files do not need a PR refersh

Extension version: 0.1.2
VS Code version: Code - Insiders 1.28.0-insider (96854113091424c07b28ba84b82b75a88b8455cb, 2018-09-10T05:18:57.119Z)
OS version: Windows_NT x64 10.0.17134

PR #482, opened 2018-09-14
Reduce calls to `updateState` in review manager

RMacfarlane:
Check if `HEAD` and `remotes` are the same before calling `updateState`. The equals check is a bit ugly, I didn't extract it out into a method because putting it in `Repository` would add another thing for Joao to have to implement

Fixes https://github.com/Microsoft/vscode-pull-request-github/issues/429

### DIFF
--- a/src/view/reviewManager.ts
+++ b/src/view/reviewManager.ts
@@ -100,8 +100,8 @@ export class ReviewManager implements vscode.DecorationProvider {
 
 			if (!sameHead || !sameRemotes) {
 				this._previousRepositoryState = {
-					HEAD: Object.assign(this._repository.HEAD),
-					remotes: this._repository.remotes.slice(0)
+					HEAD: this._repository.HEAD,
+					remotes: this._repository.remotes
 				};
 
 				this.updateState();
--- a/src/view/reviewManager.ts
+++ b/src/view/reviewManager.ts
@@ -13,7 +13,7 @@ import { Comment } from '../common/comment';
 import { GitChangeType, SlimFileChange } from '../common/file';
 import { GitErrorCodes } from '../common/gitError';
 import { IPullRequestModel, IPullRequestManager, ITelemetry } from '../github/interface';
-import { Repository } from '../common/repository';
+import { Repository, Branch } from '../common/repository';
 import { PullRequestChangesTreeDataProvider } from './prChangesTreeDataProvider';
 import { GitContentProvider } from './gitContentProvider';
 import { DiffChangeType } from '../common/diffHunk';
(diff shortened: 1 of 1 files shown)


src/view/reviewManager.ts
```diff
@@ -76,8 +82,30 @@ export class ReviewManager implements vscode.DecorationProvider {
 			vscode.commands.executeCommand('vscode.open', vscode.Uri.file(nodePath.resolve(this._repository.path, params.path)), opts);
 		}));
 		this._disposables.push(_repository.onDidRunGitStatus(e => {
-			// todo, validate state only when state changes.
-			this.updateState();
+			const oldHead = this._previousRepositoryState.HEAD;
+			const newHead = this._repository.HEAD;
+			const sameUpstream = !!oldHead.upstream
```
rebornix: This can probably be simplified to 

```
oldHead.upstream && newHead.upstream && oldHead.upstream.name === newHead.upstream.name && oldHead.upstream.remote === newHead.upstream.remote
```

RMacfarlane: I think that misses the cases where both `oldHead.upstream` and `newHead.upstream` are undefined, which should set this to true

src/view/reviewManager.ts
```diff
@@ -76,8 +82,30 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+				: !!newHead.upstream;
+			const sameHead = oldHead.ahead === newHead.ahead
+				&& oldHead.behind === newHead.behind
+				&& oldHead.commit === newHead.commit
+				&& oldHead.name === newHead.name
+				&& oldHead.remote === newHead.remote
+				&& oldHead.type === newHead.type
+				&& sameUpstream;
+
+			const sameRemotes = this._previousRepositoryState.remotes.length === this._repository.remotes.length
```
rebornix: We may want to filter out remotes created for PR purpose first so a fresh checkout will only lead to one `updateState`, triggered by branch change.

src/view/reviewManager.ts
```diff
@@ -76,8 +82,30 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+				&& this._previousRepositoryState.remotes.every(remote => this._repository.remotes.some(r => remote.equals(r)));
+
+			if (!sameHead || !sameRemotes) {
+				this._previousRepositoryState = {
+					HEAD: Object.assign(this._repository.HEAD),
```
rebornix: Is object duplication necessary here? `this._previousRepositoryState` will keep reference to the old `HEAD` as when `this._repository` status gets updated, we will create new objects for `HEAD` and `remotes`. Correct me if I miss any potential data leak.

src/view/reviewManager.ts
rebornix: Make sense.

src/view/reviewManager.ts
RMacfarlane: Nice, you're right, no need to duplicate

src/view/reviewManager.ts
RMacfarlane: I'm going to hold off on making that change now, but I agree

PR #485, opened 2018-09-14
Implement merge button, fixes #352

### LINKED ISSUE #352: Add support for merging pull request
Trying the extension the first time, I reviewed all changes (even clicked on 'Approve') but didn't find an action to merge the PR.
Closing was there.

Ikuyadeu:
Fixed #352.
Please try new function to this pull request.

screen shot 2018-09-14 at 18 51 58

Current issue
* Only merge (not squash and rebase)
* Close button is remain after merge


rebornix:
Thanks for looking into this issue, the way the merge command/button being added is pretty solid, good work!

I pushed an update to the naming of pr update event for the overview page, to reflect what it is. While playing this branch, the following issues bother me a bit:

* We will create a comment whether the merging succeeds or not. It doesn't align with GitHub website's experience, which will not automatically submit comments when merging a PR.
* We probably want to use modal instead of corner notification if we ask for users' input , like Merge, Close actions. cc @misolori .

src/github/interface.ts
```diff
@@ -1,14 +1,8 @@
-/*---------------------------------------------------------------------------------------------
- *  Copyright (c) Microsoft Corporation. All rights reserved.
- *  Licensed under the MIT License. See License.txt in the project root for license information.
- *--------------------------------------------------------------------------------------------*/
```
RMacfarlane: For
> Close button is remain after merge

I think the problem is that the return of `mergePullRequest` is different than `closePullRequest`. The `Promise<any>` below should be `Promise<IPullRequest>`. The return of merge seems to look like

```ts
data: {
	merged: boolean;
	message: string;
	sha: string;
}
```

I think we should create an interface for this data, and return result.data within the merge method.

src/commands.ts
```diff
@@ -91,6 +91,30 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: IP
...
+				try {
+					let newComment: Comment;
+					if (message) {
+						newComment = await prManager.createIssueComment(pullRequest, message);
+					}
+
+					let newPR = await prManager.mergePullRequest(pullRequest);
+					vscode.commands.executeCommand('pr.refreshList');
+					_onDidUpdatePR.fire(newPR);
+					return newComment;
```
RMacfarlane: Instead of creating and returning a new comment, let's return the result of `mergePullRequest`. Peng pointed out that we don't want to create comments if the action fails, so I think `pr.close` is also wrong, but we can fix it separately.

If it failed, we could display the message with `vscode.window.showErrorMessage`

src/commands.ts
```diff
@@ -91,6 +91,30 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: IP
 		});
 	}));
 
+	context.subscriptions.push(vscode.commands.registerCommand('pr.merge', async (pr?: PRNode, message?: string) => {
+		const pullRequest = ensurePR(prManager, pr);
+		return vscode.window.showWarningMessage(`Are you sure you want to merge this pull request on GitHub?`, 'Yes', 'No').then(async value => {
```
RMacfarlane: From Peng's suggestion to use a modal instead of a notification, we can add `{ modal: true }` after the first argument

src/commands.ts
Ikuyadeu: Fixed only `pr.merge`.
However, `pr.close` still return newComment.

src/commands.ts
Ikuyadeu: Fixed by just add a your suggestion argument.

Ikuyadeu:
@RMacfarlane Thank you for your review!
I just fixed two easy point.

RMacfarlane:
Thanks for the quick update! There was one other comment I had that may have gotten lost. The `onDidUpdatePR` method expects to be called with `IPullRequestData`. It looks like `mergePullRequest` returns

```ts
data: {
	merged: boolean;
	message: string;
	sha: string;
}
```

instead of `IPullRequestData`. I think we should create an interface for this data, and return result.data within the merge method. Then instead of calling `onDidUpdatePR`, let's directly return the result and handle it in `PullRequestOverview`.

Ikuyadeu:
@RMacfarlane Thank you for your review.
Added `MergePullRequest` that has 
```ts
export interface MergePullRequest {
	sha: string;
	merged: boolean;
	message: string;
	documentation_url: string;
}
```
`documentation_url` has a url for error message.

However I could not fixed 
> This will fix the close button still being enabled after merge

Issue #386, opened 2018-09-10
Add support for multi-root workspaces


auchenberg:
Multi-root folder support: Investigate how the active source control provider can control what's visible in the PR viewlets

dannyk81:
Wanted to know if what I experience is actually related to this (missing) enhancement.

VSCode 1.27.1
GitHub Pull Requests 0.1.2

1) I have an empty Workspace
2) I added a folder with a github repo ("repo1")
3) SCM tree view shows "Github Pull Requests" and I can see the PRs for "repo1"
4) I added a second folder with another github repo ("repo2") to the Workspace
5) SCM tree views only shows PRs from "repo1", nothing about "repo2"

Is that the expected result given current implementation?

My understanding of "multi-root folders" is somewhat different, and was actually expecting the above scenario to work, but perhaps I'm wrong and if so - sorry for the noise! and looking forward for this to be implemented 😄

RMacfarlane:
@dannyk81 Yup, that is the expected result currently. A new "Selected source controls" API was recently added to VSCode (https://code.visualstudio.com/updates/v1_27#_scm-selected-source-controls) that we're planning to use to update the SCM tree view

RMacfarlane:
Basic support for this was added in version 0.2.0 of the extension. Changing the selected repository should update the panel and all other PR related UI

![folder_select](https://user-images.githubusercontent.com/3672607/45836568-119d5100-bcc2-11e8-83bb-e009dac24c8c.gif)

Multiselecting isn't supported yet

eamodio:
Awesome! -- Can't wait for multi-select support

rebornix:
Thanks to @RMacfarlane and @joaomoreno we already shipped this feature in 0.2.3.

PR #508, opened 2018-09-20
Add multiroot support

RMacfarlane:
When there are multiple folders in the workspace, changing selection updates the view (and all review state) to be the selected folder.

folder_select

Doesn't add multi-select since the UI for that hasn't been decided (tree view for each selected folder? one tree view with top level nodes being the folders?), just shows the first selected folder in this case.

https://github.com/Microsoft/vscode-pull-request-github/issues/386

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   },
   "enableProposedApi": true,
   "preview": true,
-  "version": "0.1.5",
+  "version": "0.2.0",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
   "version": "0.1.5",
```
rebornix: "0.2.0" ;)

Issue #435, opened 2018-09-11
Add CONTRIBUTING guidelines


auchenberg:
Add CONTRIBUTING guidelines.

Inspiration: https://github.com/Microsoft/vscode/blob/master/CONTRIBUTING.md

RMacfarlane:
Did some editing of the contributing section of the wiki: https://github.com/Microsoft/vscode-pull-request-github/wiki/Contributing

Edit it further if you think anything is missing or could be clearer

PR #513, opened 2018-09-21
Add contributing.md

RMacfarlane:
https://github.com/Microsoft/vscode-pull-request-github/issues/435


rebornix:
I just updated the readme about the contributing part a little bit and added references to the wiki. My only concern is that we probably don't want to have copies in both the source code and wiki, which are likely to be out of sync. Could Wiki be a better place which won't affect our git history and everyone can update no matter when?

RMacfarlane:
Oh whoops, I didn't see that we had a contributing section on the Wiki. I'll go edit that instead

PR #529, opened 2018-10-02
Properly namespace the telemetry opt-out setting

### LINKED ISSUE #497: Telemetry setting isn't namespaced
Noticed that the setting we have to opt out from telemetry isn't namespaced, which makes it look like a built-in setting in VS Code.

screen shot 2018-09-17 at 2 47 25 pm

I suggest renaming it to `"github.pullrequests.telemetry.optout": true`

wyze:
Close #497

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,8 +32,13 @@
       "type": "object",
       "title": "GitHub Pull Requests",
       "properties": {
-        "github.telemetry.optout": {
+        "githubPullRequests.telemetry.optout": {
           "type": "boolean",
+          "description": "Disable usage data and errors to be sent to a GitHub online service"
+        },
+        "telemetry.optout": {
+          "type": "boolean",
+          "deprecationMessage": "The setting `telemetry.optout` has been deprecated in favor of `githubPullRequests.telemetry.optout`.",
           "description": "Disable usage data and errors to be sent to a GitHub online service"
         },
         "github.hosts": {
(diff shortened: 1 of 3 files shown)


wyze:
> This would mean also updating and migrating the hosts setting, if you're up for it.

@RMacfarlane Would you like that done in this PR or a separate PR?

igalic:
while we're at it, perhaps we fix the description

> #### Telemetry: Output
> [ ] Disable usage data and errors to be sent to be sent to a GitHub online service

if we compare this with the application default:

> #### Telemetry: Enable Telemetry
> [ ] Enable usage data and errors to be sent to be sent to a Microsoft online service.

having a negative description of a checkbox is incredibly confusing

RMacfarlane:
Sorry for the delay in circling back to this!

The code changes look good to me. I think @igalic makes a good point that having a negative checkbox is confusing, VSCode avoids settings like this. So let's invert the setting to `githubPullRequests.telemetry.enabled` and then I think this will be ready merge

wyze:
@RMacfarlane no worries. Would this now be an opt-in or should the default setting be enabled?

RMacfarlane:
It should be default enabled 👍

wyze:
Should be good to go now and rebased onto latest master! I added a test before but somehow it wasn't in my previous commit.

PR #538, opened 2018-10-02
Strengthen types and remove duplicated types

wyze:
This is sort of a follow up to my PR in #532, when I was writing the tests I noticed there was a typo in the interface type declaration.

Specifically here: https://github.com/Microsoft/vscode-pull-request-github/blob/1d6a41aebcfe25f10e3f83389242d2d0d3ca57ae/src/github/interface.ts#L70

It was treating it as 2 different properties, `merge_commit_sha` and `boolean`. I fixed the typo and corrected the type, also replaced some of the `any` types with some stronger types.


RMacfarlane:
Thanks for your work! 🎉 

I was wondering if we could push these types upstream to the library we're using to fetch GitHub data (https://github.com/octokit). It looks like someone has already made a change to properly type their responses, so I think a better fix for this would be to bump the version of `@octokit/rest` to the latest and remove our interfaces

wyze:
Oh nice. I didn't realized they shipped type definitions because I didn't see it in the root of their repo, but looks like they generate them!

I'll go ahead and remove the implementations here and use the ones we can from that library.

wyze:
Okay @RMacfarlane. I've upgraded the `@octokit/rest` package and changed the types around to reuse the provided types as much as possible. I also removed duplicate types that were in `preview-src/pullRequestOverviewRenderer.ts`.

preview-src/index.ts
```diff
@@ -3,9 +3,11 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 import './index.css';
-import { renderTimelineEvent, getStatus, renderComment, PullRequestStateEnum, renderReview, TimelineEvent, EventType } from './pullRequestOverviewRenderer';
-import md from './mdRenderer';
 import * as moment from 'moment';
+import { TimelineEvent, EventType } from '../src/common/timelineEvent';
+import { PullRequestStateEnum, IAccount } from '../src/github/interface';
```
RMacfarlane: We have two different tslint config files, one in this directory and one at the root. The one here has `strict: true` and `noImplicitAny: true` which the root level one doesn't. I'm seeing some linter errors from these for this imported file when building. Let's remove those compiler options temporarily so this PR stays focused

src/github/pullRequestOverview.ts
```diff
@@ -189,17 +190,17 @@ export class PullRequestOverviewPanel {
...
 		for (let commitEvent of commitEvents) {
 			const matchingCommits = commits.filter(commit => commit.sha === commitEvent.sha);
 			if (matchingCommits.length === 1) {
 				const author = matchingCommits[0].author;
 				// There is not necessarily a GitHub account associated with the commit.
 				if (author !== null) {
-					commitEvent.author.avatar_url = author.avatar_url;
-					commitEvent.author.login = author.login;
-					commitEvent.author.html_url = author.html_url;
+					commitEvent.authorExtra.avatar_url = author.avatar_url;
```
RMacfarlane: this should be left as author so that we update the existing object, authorExtra is undefined

src/github/interface.ts
```diff
@@ -41,82 +42,30 @@ export interface IAccount {
...
-			email: string;
-		};
-		message: string;
-	};
-	html_url: string;
-	sha: string;
-	parents: any;
-}
-
+export type PullRequest = Pick<
```
RMacfarlane: Nice

src/view/reviewManager.ts
```diff
@@ -297,7 +297,7 @@ export class ReviewManager implements vscode.DecorationProvider {
 
 			const comment = await this._prManager.createCommentReply(this._prManager.activePullRequest, text, thread.threadId);
 			thread.comments.push({
-				commentId: comment.id,
+				commentId: String(comment.id),
```
RMacfarlane: since we're using `toString` elsewhere and this should never be null or undefined, would prefer to use `toString` for consistency

src/github/pullRequestOverview.ts
wyze: I'm now extending the author property properly now, not sure why I didn't do it in the first place. :D

wyze:
Should be good to go with latest changes requested @RMacfarlane. Thanks for all the feedback! :D

wyze:
There was a conflict with latest master, so upon rebasing noticed there was a newer version of `@octokit/rest` so I upgraded that as well.

wyze:
Rebased again and decided to squash it to a single commit since it would look better when merging.

wyze:
@RMacfarlane rebased again. Also, upgraded to another newer version of `@ocktokit/rest`. There were a lot of changes between my last rebase, so I would double check everything.

wyze:
Thanks! :)

Issue #491, opened 2018-09-17
Allow people to increase the height of the text box as much as they want


Dan503:
At the moment the text box at the bottom of the details page does not allow users to resize it and it's default height is very small only being able to show 5.5 lines of text at any one time.

This makes it more difficult to write comments than it needs to be, especially if you need to provide a code example.

Please allow users to resize the text box to whatever height they want.

shatgupt:
I have found out that resize is explicitly disabled for comment form here: https://github.com/Microsoft/vscode-pull-request-github/blob/master/preview-src/index.css#L330

Any specific reason for it?
Is it OK if I send a PR removing it?

It works like this if I remove it:
![2018-10-01-13-57-30](https://user-images.githubusercontent.com/2388644/46315345-0976c880-c582-11e8-8e7e-7e193fb1b935.gif)

Dan503:
`resize: vertical;` sounds good, I vote for that 😊

PR #539, opened 2018-10-02
Allow vertical resizing of comment box

shatgupt:
Fixes https://github.com/Microsoft/vscode-pull-request-github/issues/491
Now looks like:


Dan503:
`resize: vertical` was deleted in the second commit. Why do that? Dont we want that?

Issue #537, opened 2018-10-02
Show the id of the pull request


kieferrm:
Issue Type: <b>Feature Request</b>

It would be helpful to see the ids of the PRs to be able to quickly spot the sorting order of PRs.

Extension version: 0.1.6
VS Code version: Code - Insiders 1.28.0-insider (60db21f08b81f03c15bb5c9395755ad3620a62bd, 2018-10-02T13:19:14.638Z)
OS version: Darwin x64 16.7.0

justinliew:
Are you thinking of the id like (#537) of this one? I can try to take this on.
![image](https://user-images.githubusercontent.com/613567/46390648-ad996600-c68d-11e8-8f6e-eda20401d2c3.png)
So in the above example, the top level tree item would say `(#537) Automate Kit's ability` ?

justinliew:
How does this look?

![image](https://user-images.githubusercontent.com/613567/46391274-d3743a00-c690-11e8-9ce2-43a17507f39e.png)

justinliew:
Added a potential PR to address this. Happy to receive comments, concerns.

PR #540, opened 2018-10-03
Show pull request id in title

justinliew:
Addresses https://github.com/Microsoft/vscode-pull-request-github/issues/537

### DIFF
--- a/src/view/treeNodes/pullRequestNode.ts
+++ b/src/view/treeNodes/pullRequestNode.ts
@@ -299,7 +299,7 @@ export class PRNode extends TreeNode {
 	getTreeItem(): vscode.TreeItem {
 		let currentBranchIsForThisPR = this.pullRequestModel.equals(this._prManager.activePullRequest);
 		return {
-			label: (currentBranchIsForThisPR ? '✓ ' : '') + '(#' + this.pullRequestModel.prNumber.toString() + ') ' + this.pullRequestModel.title,
+			label: (currentBranchIsForThisPR ? '✓ ' : '') + this.pullRequestModel.title + ' (#' + this.pullRequestModel.prNumber.toString() + ')',
 			tooltip: (currentBranchIsForThisPR ? 'Current Branch * ' : '') + this.pullRequestModel.title,
 			collapsibleState: 1,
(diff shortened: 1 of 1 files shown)


justinliew:
@RMacfarlane what's the process for getting PRs looked at? Also I didn't get a CLA bot on this and I think I need to deal with that. Thanks!

RMacfarlane:
Hmm, I'm not sure why the bot didn't kick in. I'll look into it.

There's not really a formalized process for reviews, right now I'm trying to get to them within a day of their creation, but am also balancing other work 😃

justinliew:
> Hmm, I'm not sure why the bot didn't kick in. I'll look into it.
> 
> There's not really a formalized process for reviews, right now I'm trying to get to them within a day of their creation, but am also balancing other work 😃

Ok - thanks for replying so quickly and for all your hard work! I'm new to the open source stuff but Hacktoberfest has really inspired me to try to find tools I use and help improve them :)

RMacfarlane:
Awesome!

I looked more into the CLA bot, documentation on it says it doesn't trigger on small changes. There's a checkmark next to your commit saying that all CLA requirement are met, so we're good there.

This does clearly fix the problem in #537 but it also takes up more horizontal space, which I'm not sure we want. @misolori @rebornix what do you think, is there another way we could show this info? We could show it in the tooltip, but that's not very useful at a glance. I wish the VSCode tree API allowed setting a description or providing a custom render function for tree items.

misolori:
Thanks for taking a look at this!

I do think that adding it to the front of the string adds visual noise when it's intended to be secondary info, so I think I prefer it being added to the end of the string instead:

image


Ideally if we could add decorations on the line, similar to the git/scm decorations, we could add a different color to make them stand out less, but I'm not sure if that's possible.

justinliew:
image
I think that's a good call; that way if stuff is cut off then it's the PR id, rather than part of the text. Plus, these are short PR ids; I'm sure with larger repos there are 4-5 digit ones which would definitely add unnecessary noise.

misolori:
Thanks! I'd also double-check that it gets added to the tooltip, doesn't look like it was added there as well.

PR #550, opened 2018-10-04
Fix timing issue with initialization by relaxing selected repo condition

RMacfarlane:
Small fix - when running out of sources, I noticed the tree view doesn't always get initialized. The `ui.selected` property is initialized to `false` and then gets updated as the workbench load, so there was a data race between that being updated and the extension activating.

Since the initial selected repository is always the first one in the list, or the first one opened, I opted for just removing the the `ui.selected` check.


wyze:
@malwilley and I ran into this today on his machine and we came to the same conclusion about the `ui.selected` property, because we removed the check, and everything else seemed to work as well!

Issue #543, opened 2018-10-03
PR description pane loses scroll position when focus changes


jamesgecko:
- Extension version: 0.1.2
- VSCode Version: Version 1.27.2 (1.27.2)
- OS: MacOS 10.13.6 (17G65)

Steps to Reproduce:

1. Open a PR description with a lot of comments.
2. Press `cmd`+`shift`+`p` and type `split editor`. Press `enter`. A new pane appears.
3. Open a new code file in the new pane. We're editing our code in response to comments in the PR description.
4. Click in the PR description and scroll to the bottom to see the latest comments.
5. Click back to your code pane.

What I expect to happen:

- The code window now has focus, and I can see the comments in the description pane.

What actually happens:

- The code window now has focus, and the PR description window has scrolled all the way to the top. I can no longer see the latest comments.

This issue was mentioned in (but not addressed by) https://github.com/Microsoft/vscode-pull-request-github/issues/337#issuecomment-417383903

PR #551, opened 2018-10-05
Fix description pane losing scroll position

malwilley:
Fixes https://github.com/Microsoft/vscode-pull-request-github/issues/543

Issue was that the webview HTML was being reset every time `update` was called. Now the HTML won't be reset unless the PR has changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -303,6 +303,7 @@
   },
   "devDependencies": {
     "@types/chai": "^4.1.4",
+    "@types/debounce": "^3.0.0",
     "@types/keytar": "^4.0.1",
     "@types/lodash": "^4.14.106",
     "@types/markdown-it": "^0.0.5",
@@ -327,8 +328,9 @@
   },
   "dependencies": {
     "@octokit/rest": "^15.9.5",
-    "es6-promise": "^4.2.5",
     "axios": "^0.18.0",
+    "debounce": "^1.2.0",
+    "es6-promise": "^4.2.5",
     "git-credential-node": "^1.1.0",
     "iconv-lite": "0.4.23",
     "markdown-it": "^8.4.0",
--- a/preview-src/index.ts
+++ b/preview-src/index.ts
@@ -80,6 +80,9 @@ function handleMessage(event: any) {
 		case 'pr.enable-exit':
 			(<HTMLButtonElement>document.getElementById(ElementIds.CheckoutDefaultBranch)).disabled = false;
 			break;
+		case 'set-scroll':
+			window.scrollTo(message.scrollPosition.x, message.scrollPosition.y);
+			break;
 		default:
 			break;
 	}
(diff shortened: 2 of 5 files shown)


RMacfarlane:
Thanks for taking this on!

From my understanding of webviews, when the view becomes hidden and then visible again, the scroll position is not preserved even if the html stays the same. With this change I still see scroll position get reset when switching to a different tab and back to the webview - do you also see this?

I think instead we need to save off this state in the webview - the `pendingCommentText` in `preview-src/index.ts` is a potential example. The markdown extension in VSCode also manages scroll position in its webview

malwilley:
Ah, I see now this was a naive solution! This only fixes the issue listed in the issue (switching between split pane views), not the more general case of switching tabs. I'll work on a more general fix, thanks for the info!

malwilley:
@RMacfarlane I think this should be closer to what we want here. There wasn't an obvious spot for storing the scroll position so I added another private member to `pullRequestOverview.ts` but let me know if that is better kept elsewhere.

src/github/pullRequestOverview.ts
```diff
@@ -224,6 +229,10 @@ export class PullRequestOverviewPanel {
 				return this.checkoutDefaultBranch(message.branch);
 			case 'pr.comment':
 				return this.createComment(message.text);
+			case 'scroll': {
+				this._scrollPosition = message.scrollPosition
```
RMacfarlane: Missing semicolon :)

The linter should run when you do `yarn run watch`, but it's often the top of the output so easy to miss

preview-src/index.ts
```diff
@@ -218,6 +221,16 @@ function addEventListeners(pr: PullRequest): void {
 			branch: pr.repositoryDefaultBranch
 		});
 	});
+
+	window.onscroll = () => {
```
RMacfarlane: This event fires really frequently, so I think we should throttle posting messages.

Maybe something like using `setTimeout` for changes to `pendingCommentText` above

RMacfarlane:
@malwilley This approach looks good! Yeah, I think adding the scroll position to `pullRequestOverview` makes sense. Just had a couple small comments

preview-src/index.ts
malwilley: Good thought! I hope it's okay that I've added `debounce` as a dependency for this purpose.

package.json
```diff
@@ -327,8 +328,9 @@
   },
   "dependencies": {
     "@octokit/rest": "^15.9.5",
-    "es6-promise": "^4.2.5",
     "axios": "^0.18.0",
+    "debounce": "^1.2.0",
```
RMacfarlane: Let's update `ThirdPartyNotices.txt` with the license text for this, and then I think this is ready to merge!

RMacfarlane:
Thank you!

PR #554, opened 2018-10-05
[WIP] A pull request to create pull requests

queerviolet:
This pull request requests that these changes—which create pull requests—be pulled into master, so users can create github pull requests with vscode-pull-request-github.

(However, I request that you not pull these changes—which do not yet actually create github pull requests—quite yet.)

### DIFF
--- a/src/github/pullRequestManager.ts
+++ b/src/github/pullRequestManager.ts
@@ -394,6 +394,12 @@ export class PullRequestManager implements IPullRequestManager {
 		}
 	}
 
+	async createPullRequest(title: string, description: string = '', targetBranch: string = 'master'): Promise<boolean> {
+		Logger.appendLine(`will create PR with title=${title} description=${description} branch=${targetBranch}`);
+		// TODO: That ☝🏽
+		return false;
+	}
+
 	private async changePullRequestState(state: 'open' | 'closed', pullRequest: IPullRequestModel): Promise<any> {
 		const { octokit, remote } = await (pullRequest as PullRequestModel).githubRepository.ensure();
 
--- a/package.json
+++ b/package.json
@@ -76,6 +76,11 @@
     },
     "commands": [
       {
+        "command": "pr.create",
+        "title": "Create Pull Request",
+        "category": "GitHub Pull Requests"
+      },
+      {
         "command": "pr.pick",
         "title": "Checkout Pull Request",
         "category": "GitHub Pull Requests"
@@ -171,6 +176,10 @@
         {
           "command": "pr.close",
           "when": "config.git.enabled && github:inReviewMode"
+        },
+        {
+          "command": "pr.create",
+          "when": "config.git.enabled && !github.inReviewMode"
         },
         {
           "command": "pr.merge",
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -17,6 +17,7 @@ import { GitChangeType } from './common/file';
 import { getDiffLineByPosition, getZeroBased } from './common/diffPositionMapping';
 import { DiffChangeType } from './common/diffHunk';
 import { DescriptionNode } from './view/treeNodes/descriptionNode';
+import Logger from './common/logger';
 
 const _onDidUpdatePR = new vscode.EventEmitter<IPullRequest>();
 export const onDidUpdatePR: vscode.Event<IPullRequest> = _onDidUpdatePR.event;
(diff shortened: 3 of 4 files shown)


rebornix:
I played with current code a bit and it works pretty well with typing in title/desc/target-branch ;)

What I'm not sure about is what's the expected first step for the pull request creation workflow in an editor: providing title/desc or reviewing the file changes first. The latter is the one I want to do first. We can probably create a tree view, similar to "Changes in Pull Request" and it contains file changes and commits. The description node for this tree view is now a pull request submit form where we type in necessary information for the pull request creation. Or we can use Quick Pick UI to do the submition work, like what this PR already implements. @queerviolet what do you think?

queerviolet:
I think what we want to do long term is to immediately create a draft PR. Draft PRs are not pushed to the server—at least not publicly—but they otherwise have the same view as other PRs. We could either create another top-level section in the Github Pull Requests sidebar tree view (Drafts), or we could put them under Created By Me with badges that say *Draft*.

This way, you don't have to learn another interface, and we don't introduce another interface mode. You can edit the draft PR exaclty how you can edit any other PR you create, at a leisurely pace, and once you're ready, you can click a *Publish PR* button that publishes the PR.

I'm working on some mocks of this flow right now that we can discuss. Generally, it will require the following changes:

  1. Enabling editing on the description page (which we want to do anyway)
  2. Either finding somewhere to save draft PR data locally metadata locally, or adding backend support for draft PRs.

Since it's a bit more involved, I wanted to get this feature rolling first with absolutely minimal UX. The evolution I see is:

  1. This PR: Create PR with super minimal UI
  2. Soon: Editing PRs
  3. Draft PRs?

1 & 2 get us 90% of the way to where we need to be. With those two features in, you'll be able to create a [WIP] PR and edit it on the description page. The only thing you won't be able to do is preview a PR before sending it to the server. I think being able to do that is important in terms of building confidence—i.e., overcoming fear of submitting a PR that isn't quite perfect. But it doesn't really enable anything that you couldn't do before, provided you're willing to edit in public a bit.

So, in short: I agree, but think that should be a (near) future evolution of this feature.

queerviolet:
So I've changed my mind. I don't think it's worth it to try for a version of this with a super-minimal UI. Common cases have too much state to work with before submitting, and asking a series of questions with the prompt UI is scary and unwieldy.

Consider a common case:

1. I clone an upstream repo, which I do not have permission to write to. (Perhaps I should have forked, but I wasn't thinking about it).
2. I do some work in a local branch in that repo.
3. I invoke create PR.

Now what?

I think, ideally, we'll create a fork of the upstream and push our local branch there, and then create a branch against the fork. But that's a *lot* of magic to just "have happen". Ideally, I'd want the user to *at least see* what's going to happen on the PR description page before clicking Create PR, which means going ahead and creating that interface.

I'll post some of the mocks we used for UX testing, so we can discuss them. Also, I'm planning on saving draft PR information in git config settings, since we're already saving information there, and those can, it appears, hold arbitrary amounts of data.

sguthals:
You're right about this common workflow @queerviolet.

As of today, the way we handle PRs when you don't have push access is to have users create a Fork. This is a more complex flow that we should definitely open up another issue for it and have that be the next thing we tackle in this flow. 

The basic flow for creating a PR is as such:
1. You’re currently on a branch
2. You have push access to upstream
3. You choose a target branch from the repo
4. You add a title
5. You add a description
6. You click “Create PR” which will immediately create the PR on dotcom and now you have all the PR functionalities in VS Code
And for right now:
7. If you don’t have push access, we fail gracefully

Do you want to open an issue to handle 7 in the above list better?

queerviolet:
Let's just see how this works first.

queerviolet:
Okay, this is ready for review to begin.

I'll probably want to add some tests in a bit, but this has become a bit of a monster, so let's get the ball rolling.

shared/actions.ts
```diff
@@ -0,0 +1,57 @@
+import { Upstream } from './state';
```
RMacfarlane: would it make sense to move the new `shared` and `store` folders into `preview-src` since they're being used by the webview?

RMacfarlane: Nevermind, I see that we import from `store` on the extension side

src/github/githubRepository.ts
```diff
@@ -36,15 +37,20 @@ export class GitHubRepository implements IGitHubRepository {
 	constructor(public remote: Remote, private readonly _credentialStore: CredentialStore) {
 	}
 
-	async resolveRemote(): Promise<void> {
-		try {
-			const { octokit, remote } = await this.ensure();
-			const { data } = await octokit.repos.get({
+	async getMetadata(): Promise<any> {
```
RMacfarlane: can we add a type for this metadata?

src/store/index.ts
```diff
@@ -0,0 +1,53 @@
...
+import { IPullRequestManager } from '../github/interface';
+
+import reducer from './reducers';
+import Logger from '../common/logger';
+
+import newPRMiddleware from './middleware/newPR';
+import { Disposable, Event } from 'vscode';
+import { init } from './actions';
+
+export class Store {
```
RMacfarlane: is this meant to ultimately store all of the extension's state? what's the responsibility of this class?

Centralizing state is fine, but I'm concerned that we now have duplicate sources of truth for some of the state - ex. the PullRequestManager is also tracking `githubRepositories`

src/github/pullRequestManager.ts
```diff
@@ -34,6 +36,39 @@ interface RestError {
 	field: string;
 	resource: string;
 }
+export class NoGitHubReposError extends Error {
```
RMacfarlane: should we extract errors to their own file?

kieferrm:
Pls see #620.

PR #555, opened 2018-10-06
Remove dead code related to PR badge status color, fixes #535

### LINKED ISSUE #535: PR status badge background color
More of a question if this is wanted or not. I noticed there is a function in the code to get the color for a given PR status, but it was not used in the codebase, outside of it's definition.

Should we use it to color the background of the status badges on the overview/description view? If desired, I already have the code working and can send a PR. See a preview of Open and Merged below:

image image

arkon:
Removes dead code previously used for coloring the PR status badge. Fixes #535.


arkon:
Sounds good, updated!

PR #556, opened 2018-10-06
Show labels on PR description

### LINKED ISSUE #464: Add labels to PR description
It would be awesome to have labels visible in the PR description.

wyze:
Close #464 

-----

I'm not sure the design, but it should mimic GitHub's label design. I wen't ahead and not pulled in the color, so it is similar to the status badge. Feels like a lot of wasted space if there is a single label, so definitely open to changing up the design.

Without Labels:

image

With Labels:

screen shot 2018-10-06 at 4 08 39 pm


misolori:
Thanks for taking a look at this! For the design, I'd probably move down into the body and use the `octicon-tag` icon (you can see the commit icon for reference), similar to their mobile view:

image

wyze:
Here is the new look and I fixed a bug where multiple tags would have been rendered incorrectly, but got that fixed up here!

image

wyze:
@misolori Changes made. Used the same PR in your screenshot as my test so should work great now! Again rebased to latest master as well.

Issue #293, opened 2018-08-28
Should commits be more compact?


Tyriar:
https://github.com/Microsoft/vscode/issues/57301

There's a lot of padding between them:

<img width="1015" alt="screen shot 2018-08-28 at 11 03 29 am" src="https://user-images.githubusercontent.com/2193314/44741475-0a639880-aab2-11e8-90a2-be60aa5de047.png">

<img width="877" alt="screen shot 2018-08-28 at 11 04 18 am" src="https://user-images.githubusercontent.com/2193314/44741518-2bc48480-aab2-11e8-8773-5bc1ce7143e3.png">

PR #557, opened 2018-10-07
update css with proper padding for commits

igeligel:
#293

- fixed border to outer container of timeline events
- added improved padding for comment container of commit types

Padding should be `5px` in my opinion because the base height of a `comment-container` with type `data-type="commit"` is `18px`. Based on this and with golden ratio we will get around of margin of `~10px-11px` (`18 * 0.61 = 10.98`). Basically this is a padding of 5px on top and on bottom of the `comment-container` elements.

### DIFF
--- a/preview-src/index.css
+++ b/preview-src/index.css
@@ -22,6 +22,11 @@ a:focus, input:focus, select:focus, textarea:focus {
 	border: none;
 }
 
+#timeline-events {
+	padding: 8px 0;
+	border-bottom: 1px solid var(--border-color);
+}
+
 h2 {
 	margin: 0;
 }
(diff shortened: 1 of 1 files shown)


preview-src/index.css
```diff
@@ -22,6 +22,11 @@ a:focus, input:focus, select:focus, textarea:focus {
 	border: none;
 }
 
+#timeline-events {
+	padding: 8px 0;
```
misolori: The `#timeline-events` is for the entire wrapper for comments, commits, and reviews. So the padding gets applied to the beginning and end. In your PR description you only show a stream of commits, however this breaks when you have variations in the timeline:

image
image


I'd probably remove this and tweak the padding/margin on `.comment-container[data-type="commit"]`

RMacfarlane:
Hey @igeligel, thanks for looking into this! I'm clearing out old PRs and am going to close this since it hasn't had any recent activity.

Issue #273, opened 2018-08-25
Add remote using ssh instead of https, when key exists


roblourens:
If you can detect when this is possible, you should add the remote using the ssh url (`git@github.com:JacksonKearl/vscode.git`) instead of the https url (`https://github.com/JacksonKearl/vscode`).

I can push to the first but not the second without getting a personal access token because 2FA.

PR #561, opened 2018-10-08
Support ssh remotes

rebornix:
Fix https://github.com/Microsoft/vscode-pull-request-github/issues/273 . The logic in this PR is using the same scheme as the origin one (or where the PR is targeting) for the newly created remotes. For example, if the origin remote is using ssh and we try to create a new remote for a pr on the origin remote, we should use ssh instead of https.

### DIFF
--- a/src/common/protocol.ts
+++ b/src/common/protocol.ts
@@ -150,6 +150,54 @@ export class Protocol {
 		}
 	}
 
+	toString(): string {
+		// based on Uri scheme for SSH https://tools.ietf.org/id/draft-salowey-secsh-uri-00.html#anchor1 and heuristics of how GitHub handles ssh url
+		// sshUri        = `ssh:`
+		//    - omitted
+		// hier-part     =  "//" authority path-abempty
+		//    - // is omitted
+		// authority     = [ [ ssh-info ] "@" host ] [ ":" port]
+		//   - ssh-info: git
+		//   - host: ${this.host}
+		//   - port: omitted
+		// path-abempty  = <as specified in [RFC3986]>
+		//   - we use relative path here `${this.owner}/${this.repositoryName}`
+		if (this.type === ProtocolType.SSH) {
+			return `git@${this.host}:${this.owner}/${this.repositoryName}`;
+		}
+
+		if (this.type === ProtocolType.GIT) {
+			return `git://git@${this.host}:${this.owner}/${this.repositoryName}`;
+		}
+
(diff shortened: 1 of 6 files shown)


src/common/protocol.ts
```diff
@@ -150,6 +150,54 @@ export class Protocol {
 		}
 	}
 
+	toString(): string {
+		// based on Uri scheme for SSH https://tools.ietf.org/id/draft-salowey-secsh-uri-00.html#anchor1 and heuristics of how GitHub handles ssh url
```
rebornix: @queerviolet may I have your expertise and review on this, my heuristics of how GitHub generates ssh uri for repos might be wrong.

queerviolet:
Happily :)

src/common/protocol.ts
```diff
@@ -150,6 +150,54 @@ export class Protocol {
 		}
 	}
 
+	toString(): string {
```
RMacfarlane: let's update the tests to also check the `type` of each protocol, and add some unit tests for this method

src/common/protocol.ts
queerviolet: This looks right, to the best of my knowledge. +1 adding some tests on it. 👍🏽

src/common/protocol.ts
```diff
@@ -150,6 +150,54 @@ export class Protocol {
...
+
+		let normalizedUri = this.normalizeUri();
+		if (normalizedUri) {
+			return normalizedUri.toString();
+		}
+
+		return null;
+	}
+
+	with(change: { type?: ProtocolType; host?: string; owner?: string; repositoryName?: string; }): Protocol {
```
queerviolet: I would call this method `update`. `with` suggests to me an operation that returns a new Protocol without mutating this one.

rebornix: that's right. I didn't make up my mind whether we should make it mutable or not but let's change it to `update` for now. Thanks!

src/common/protocol.ts
rebornix: `type` checking is there. The only catch here is we don't have code path for `git://` at this moment.

rebornix:
Added test cases for 

* `type` checking. At this moment, we will see both SSH or SCP like Uris as SSH protocol (which means we don't really have `git://` scheme detection) but it should be fine.
* `update` method for Protocol. It's mutable so the name is more accurate than `with` as @queerviolet pointed out.

queerviolet:
LGTM :)

PR #571, opened 2018-10-11
Change the changelog to be in reverse chronological order

wyze:
It is common that changelogs list the most recent items first so there is less scrolling involved when looking for what recent changes are.


RMacfarlane:
Thanks @wyze! I pushed a commit to fix a merge conflict with master.

wyze:
Awesome. Thank you!

Issue #282, opened 2018-08-28
Load more entry should provide progress feedback


dbaeumer:
Testing https://github.com/Microsoft/vscode/issues/57301

like opening a parent node does.

alexr00:
There is progress feedback in the view.

PR #577, opened 2018-10-14
Show more details in PR tree view tooltip, closes #568.

### LINKED ISSUE #568: Show more details in tree view tooltip
Right now it just re-displays the PR title, it would be nice to also show the author as sometimes it's not obvious (hard to diff the top and bottom avatars):

image

jerrymajewski:
- Show the PR's `author login name` and `creation date` in the `GitHub Pull Requests` tree node tooltips.

**Before:**

screen shot 2018-10-14 at 12 13 14 pm

**After:**

screen shot 2018-10-14 at 12 13 14 pmb

### DIFF
--- a/src/view/treeNodes/pullRequestNode.ts
+++ b/src/view/treeNodes/pullRequestNode.ts
@@ -5,7 +5,6 @@
 
 import * as vscode from 'vscode';
 import * as path from 'path';
-import * as moment from 'moment';
 import { parseDiff, getModifiedContentFromDiffHunk, DiffChangeType } from '../../common/diffHunk';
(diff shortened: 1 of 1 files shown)


rebornix:
I like the idea but I'm a bit reluctant to display time in the PR tooltip, especially when it can be out of date very easily. cc @misolori for review , this is probably the first time we introduce multi line tooltip in this extension.

misolori:
I agree that we should just display the author with no timestamp on a single line.

jerrymajewski:
I updated the tooltip as follows:

screen shot 2018-10-24 at 5 27 59 am

misolori:
We should just append the author to the end of the string after the parenthesis: 
`Closes BLOS-223 (#282) by katwal-dipak`

Issue #573, opened 2018-10-12
vscode:// url redirect does not open vscode. No alternatives offered.


diddledan:
Issue Type: <b>Bug</b>

1. On Ubuntu run `snap install vscode` to install from the snap store
2. Start vscode using the icon in the dashboard
3. Install the GitHub Pull Requests plugin and click the popup to login
4. Login to GitHub and click the "Continue" button
5. If the browser prompts to run `xdg-open` approve the request (this is the Linux mechanism to launch URL handlers
6. VSCode does not receive the URL trigger because snaps don't provide an interface for this feature yet.
7. Try to find an alternative way to trigger vscode with the URL directly, such as on the commandline, but realise there is no documentation on how to do such things :-)
8. That's it.

The vscode cannot be triggered via the URL handling mechanism as mentioned because of snap limitations currently. There is no alternative mechanism to get the URL to vscode that is readily identifiable to a user. It would be handy to provide a command line snippet that will achieve the same as the URL handler for cases where the URL handler fails.

Extension version: 0.2.0
VS Code version: Code 1.28.0 (431ef9da3cf88a7e164f9d33bf62695e07c6c2a9, 2018-10-05T14:52:58.945Z)
OS version: Linux x64 4.15.0-36-generic

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz (8 x 3834)|
|GPU Status|2d_canvas: unavailable_software<br>checker_imaging: disabled_off<br>flash_3d: unavailable_software<br>flash_stage3d: unavailable_software<br>flash_stage3d_baseline: unavailable_software<br>gpu_compositing: unavailable_software<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: unavailable_software<br>video_decode: unavailable_software<br>video_encode: unavailable_software<br>webgl: enabled_readback<br>webgl2: unavailable_off|
|Load (avg)|2, 2, 2|
|Memory (System)|31.30GB (0.40GB free)|
|Process Argv|.|
|Screen Reader|no|
|VM|0%|

</details>

rebornix:
Sorry we didn't mention that anywhere but you can use personal access token for authentication by setting

```
    "github.hosts": [
        {
            "host": "https://github.com",
            "username": "",
            "token": ""
        }
    ]
```

could you give it a try?

[...]

PR #581, opened 2018-10-16
Add a command to manually accept a token response.

queerviolet:
This sets up a command triggered by clicking on the status indicator during auth. The command prompts for the token (really for the URI response), parses it as a URI, and dispatches it.

Fixes #570 #573

### DIFF
--- a/src/authentication/githubServer.ts
+++ b/src/authentication/githubServer.ts
@@ -8,8 +8,7 @@ import axios from 'axios';
 const SCOPES: string = 'read:user user:email repo write:discussion';
 const GHE_OPTIONAL_SCOPES: object = {'write:discussion': true};
 
-// const AUTH_RELAY_SERVER = 'https://vscode-auth.github.com';
-const AUTH_RELAY_SERVER = 'https://client-auth-staging-14a768b.herokuapp.com';
+const AUTH_RELAY_SERVER = 'https://vscode-auth.github.com';
 const EXTENSION_ID = 'GitHub.vscode-pull-request-github';
 const CALLBACK_PATH = '/did-authenticate';
 const CALLBACK_URI = vscode.version.endsWith('-insider')
--- a/src/authentication/githubServer.ts
+++ b/src/authentication/githubServer.ts
@@ -8,7 +8,8 @@ import axios from 'axios';
 const SCOPES: string = 'read:user user:email repo write:discussion';
 const GHE_OPTIONAL_SCOPES: object = {'write:discussion': true};
 
-const AUTH_RELAY_SERVER = 'https://vscode-auth.github.com';
+// const AUTH_RELAY_SERVER = 'https://vscode-auth.github.com';
+const AUTH_RELAY_SERVER = 'https://client-auth-staging-14a768b.herokuapp.com';
 const EXTENSION_ID = 'GitHub.vscode-pull-request-github';
 const CALLBACK_PATH = '/did-authenticate';
 const CALLBACK_URI = vscode.version.endsWith('-insider')
--- a/src/github/credentials.ts
+++ b/src/github/credentials.ts
@@ -194,13 +194,13 @@ export class CredentialStore {
 
 	private willStartLogin(authority: string): void {
 		const status = this._authenticationStatusBarItems.get(authority);
-		status.text = `Signing in to ${authority}...`;
+		status.text = `$(mark-github) Signing in to ${authority}...`;
 		status.command = AUTH_INPUT_TOKEN_CMD;
 	}
 
 	private didEndLogin(authority: string): void {
 		const status = this._authenticationStatusBarItems.get(authority);
-		status.text = `Signed in to ${authority}`;
+		status.text = `$(mark-github) Signed in to ${authority}`;
 		status.command = null;
 	}
 
--- a/src/github/credentials.ts
+++ b/src/github/credentials.ts
@@ -11,9 +11,12 @@ import { Remote } from '../common/remote';
 import { VSCodeConfiguration } from '../authentication/vsConfiguration';
 import Logger from '../common/logger';
 import { ITelemetry } from './interface';
+import { handler as uriHandler } from '../common/uri';
 
 const TRY_AGAIN = 'Try again?';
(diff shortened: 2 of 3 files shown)


src/authentication/githubServer.ts
```diff
@@ -8,7 +8,8 @@ import axios from 'axios';
 const SCOPES: string = 'read:user user:email repo write:discussion';
 const GHE_OPTIONAL_SCOPES: object = {'write:discussion': true};
 
-const AUTH_RELAY_SERVER = 'https://vscode-auth.github.com';
+// const AUTH_RELAY_SERVER = 'https://vscode-auth.github.com';
```
sguthals:
```suggestion
// const AUTH_RELAY_SERVER = 'https://vscode-auth.github.com';
// I'm adding another comment
// And another one
```

src/authentication/githubServer.ts
```diff
@@ -8,7 +8,8 @@ import axios from 'axios';
...
+const AUTH_RELAY_SERVER = 'https://client-auth-staging-14a768b.herokuapp.com';
```
sguthals:
```suggestion
```

queerviolet:
Ping for review on this? We're getting a lot of folks on Linux who can't use the protocol handler :/ #591

src/github/credentials.ts
```diff
@@ -25,6 +28,15 @@ export class CredentialStore {
 		this._configuration = configuration;
 		this._octokits = new Map<string, Octokit>();
 		this._authenticationStatusBarItems = new Map<string, vscode.StatusBarItem>();
+		vscode.commands.registerCommand(AUTH_INPUT_TOKEN_CMD, async () => {
+			const uriStr = await vscode.window.showInputBox({ prompt: 'Token' });
+			if (!uriStr) { return; }
+			const uri = vscode.Uri.parse(uriStr);
```
RMacfarlane: I think we should try-catch this

src/github/credentials.ts
```diff
@@ -89,32 +101,35 @@ export class CredentialStore {
 
 		// the remote url might be http[s]/git/ssh but we always go through https for the api
 		// so use a normalized http[s] url regardless of the original protocol
-		const normalizedUri = remote.gitProtocol.normalizeUri();
-		const host = `${normalizedUri.scheme}://${normalizedUri.authority}`;
+		const {scheme, authority} = remote.gitProtocol.normalizeUri();
```
RMacfarlane: we should add the tslint rule for this, in other places we have whitespace around curlys when deconstructing

src/github/credentials.ts
```diff
@@ -89,32 +101,35 @@ export class CredentialStore {
...
 				}
+			} finally {
+				this.didEndLogin(authority);
 			}
 
 			if (octokit) {
 				retry = false;
 			} else if (retry) {
-				retry = (await vscode.window.showErrorMessage(`Error signing in to ${normalizedUri.authority}`, TRY_AGAIN)) === TRY_AGAIN;
+				retry = (await vscode.window.showErrorMessage(`Error signing in to ${authority}`, TRY_AGAIN)) === TRY_AGAIN;
```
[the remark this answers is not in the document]
queerviolet: Oh, yeah, probably a good idea.

src/github/credentials.ts
```diff
@@ -177,6 +192,18 @@ export class CredentialStore {
 		statusBarItem.command = command;
 	}
 
+	private willStartLogin(authority: string): void {
+		const status = this._authenticationStatusBarItems.get(authority);
+		status.text = `$(mark-github) Signing in to ${authority}...`;
+		status.command = AUTH_INPUT_TOKEN_CMD;
```
RMacfarlane: it seems sort of strange to me to start a different method of completing auth when auth is in flight.

I think I would expect clicking the status bar item to not do anything while it has that text

src/github/credentials.ts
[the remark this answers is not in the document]
queerviolet: (That said, it doesn't really address the issue we're seeing, since auth hangs in a pending state, waiting for the URI handler to get called.)

src/github/credentials.ts
queerviolet: It... is a bit odd, I won't deny that 😅

I initially made it a popup with an `Enter token manually` button. The trouble is, there's no way to destroy those without the user interacting with them. So in the typical case where auth succeeds, it's left lying around, which felt janky to me.

This isn't meant to be particularly discoverable. It's a companion to [this auth server PR](https://github.com/editor-tools/authentication-server/pull/13), which gives folks instructions on what to do if the URI handler doesn't work.

src/github/credentials.ts
```diff
@@ -25,6 +28,15 @@ export class CredentialStore {
 		this._configuration = configuration;
 		this._octokits = new Map<string, Octokit>();
 		this._authenticationStatusBarItems = new Map<string, vscode.StatusBarItem>();
+		vscode.commands.registerCommand(AUTH_INPUT_TOKEN_CMD, async () => {
```
RMacfarlane: I'm not sure if adding it to the `package.json` so that it shows up in the command palette makes sense or not

It would be much more clear to have only one command for sign in, but I think it's also good to have a way to directly invoke this

src/github/credentials.ts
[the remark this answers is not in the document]
RMacfarlane: Ah, got it

src/github/credentials.ts
queerviolet: What would we do with it? Currently, if I paste in a malformed URI (rather hard to do, really) the error bubbles up to the UI:

screen shot 2018-10-19 at 1 00 34 pm

Which seems right?

src/github/credentials.ts
queerviolet: Changed this instance. I'll add the lint rule and clean up anything it finds in a separate CL.

src/github/credentials.ts
queerviolet: Oh, yeah, I thought about doing that, and then didn't do that.

I think it's probably a good idea to be able to invoke it from there. Added.

queerviolet:
Ok, PTAL.

PR #583, opened 2018-10-17
Pull request creation UI. Now with Preact!

queerviolet:
I went ahead and threw in preact and redux, to handle the complicated interactions of state b/t this webview and the extension.


queerviolet:
(I created this for a demo. The real PR is #554. Closing this one ;) )

PR #588, opened 2018-10-18
Match github.com empty comment behavior (Fixes #587)

### LINKED ISSUE #587: Misleading error message when submitting empty comment
I just started playing with this extension and I noticed I am getting a misleading error message when I accidentally tried to add an empty comment. This is a super minor issue but I figured this would be a good reason to download the code and see how it was built. I am sending a PR to fix this.

image

- Extension version: 0.2.1
- VSCode Version:1.28.1
- OS: Windows

Steps to Reproduce:

1. Add a couple new lines in the comment box
2. Click Add

mmanela:
When you submit a comment and only give new lines it gives a misleading error message.

image

This PR (fixes #587) contains a small change to match the error behavior that GitHub.com uses in this case 
image

I also added some types for the error object. I looked in the octokit typings and didn't see something there. This is my first time contributing to this repo so please let me know if I am missing or need to change anything.

Thanks for the awesome extension.

### DIFF
--- a/src/github/pullRequestManager.ts
+++ b/src/github/pullRequestManager.ts
@@ -549,20 +549,22 @@ export class PullRequestManager implements IPullRequestManager {
 
 	private handleError(e: any) {
 		if (e.code && e.code === 422) {
+			let errorObject: RestErrorResult;
 			try {
-				let errorObject: RestErrorResult = e.message && JSON.parse(e.message);
-				const firstError = errorObject && errorObject.errors && errorObject.errors[0];
-				if (firstError && firstError.code === 'missing_field' && firstError.field === 'body') {
-					throw new Error('Body can\'t be blank');
-				} else {
-					throw new Error('There is already a pending review for this pull request on GitHub. Please finish or dismiss this review to be able to leave more comments');
-				}
+				errorObject = e.message && JSON.parse(e.message);
 			}
 			catch {
 				// If we failed to parse the JSON re-throw the original error
 				// since it will have a more useful stack
 				throw e;
 			}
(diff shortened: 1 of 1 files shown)


src/github/pullRequestManager.ts
```diff
@@ -544,6 +547,20 @@ export class PullRequestManager implements IPullRequestManager {
 		return this.repository.checkout(branchName);
 	}
 
+	private handleError(e: any) {
+		if (e.code && e.code === 422) {
+			const errorObject: RestErrorResult = e.message && JSON.parse(e.message);
```
rebornix: JSON.parse might throw exceptions and let's protect that.

rebornix:
It looks to me in overall. I pushed a commit to fix tslint errors but left a tiny comment for JSON.parse. Once that's fixed we can have it merged. Thanks for your help!

src/github/pullRequestManager.ts
mmanela: Thanks updated

src/github/pullRequestManager.ts
```diff
@@ -544,6 +547,28 @@ export class PullRequestManager implements IPullRequestManager {
...
+		if (e.code && e.code === 422) {
+			let errorObject: RestErrorResult;
+			try {
+				errorObject = e.message && JSON.parse(e.message);
+			} catch {
+				// If we failed to parse the JSON re-throw the original error
+				// since it will have a more useful stack
+				throw e;
+			}
+			const firstError = errorObject && errorObject.errors && errorObject.errors[0];
```
rebornix: 👍

rebornix:
LGTM. Thanks again for your help!

Issue #572, opened 2018-10-12
Is there any proxy support ? (with Github Enterprise)


yoh1496:
Nothing is shown in my project cloned from GitHub Enterprise behind a proxy of company. Can you let me know how to resolve this problem?

- Extension version: 0.2.0
- VSCode Version: 1.28.0
- OS: Windows 10 1803

Steps to Reproduce:

1. Write `github.hosts` in setting file.
2. Cloning repository, open the project folder.
3. Nothing is shown.

The output of this extension is below

```
Looking for git repository
Git repository found, initializing review manager and pr tree view.
No GitHub remotes found
No GitHub remotes found
No GitHub remotes found
Review> no matching pull request metadata found for current branch master
``` 

In my environment, a repository cloned from GitHub was shown correctly.

```
Looking for git repository
Git repository found, initializing review manager and pr tree view.
No GitHub remotes found
No GitHub remotes found
Found GitHub remote
```

Reading some sources in this extension, I thought that there was no considering about being used behind proxy. Is there any way to use this extension behind a proxy?

queerviolet:
This should be fixed by #590. Please reopen if there are any issues!

PR #590, opened 2018-10-19
add httpsOverHttp to support proxy

yoh1496:
#572 

- adding proxy support ( using `HTTPS_PROXY` environmental value )

I wrote a little code to use this extension under the environment needs to use a proxy.
Code review and consider to support using a proxy please.

### DIFF
--- a/src/github/credentials.ts
+++ b/src/github/credentials.ts
@@ -147,19 +147,6 @@ export class CredentialStore {
 	}
 
 	private createOctokit(type: string, creds: IHostConfiguration): Octokit {
-		let proxy: object | undefined;
-		try {
-			const proxyURL = new URL(process.env.HTTPS_PROXY);
-			proxy = {
-				host: proxyURL.hostname,
-				port: proxyURL.port,
-				proxyAuth: (proxyURL.username && proxyURL.password) ?
-					`${proxyURL.username}:${proxyURL.password}` : null,
-			};
-		} catch(e) {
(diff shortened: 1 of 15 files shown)


src/github/credentials.ts
```diff
@@ -131,7 +133,14 @@ export class CredentialStore {
 	}
 
 	private createOctokit(type: string, creds: IHostConfiguration): Octokit {
+		const proxySettings = process.env.HTTPS_PROXY ?
+			new URL(process.env.HTTPS_PROXY) : null;
```
queerviolet: Could we catch the possible parsing error here and `showErrorMessage`?

yoh1496:
@queerviolet Thank you for your code review!

I wrote a try-catch block
and I wrote a code for supporting BASIC-Auth proxy ( I forgot to write in this pull request ).

Code review please, again.

queerviolet:
Hi! This now fails in the case when `HTTPS_PROXY` isn't set. Also, I just realized that we have the exact same agent creation code in two places. We don't really need that—we can create the agent once at startup and forget about it.

I made a PR on your PR with that change: https://github.com/yoh1496/vscode-pull-request-github/pull/1

The diff looks huge, because I also rebased atop the most recent master. Rebase your branch off `Microsoft:master` first, and it'll be a lot less shocking.

yoh1496:
@queerviolet hi, I'm sorry for my late response and dirty commits,,,

I modified a little code given from you and pushed it. Please confirm it.

queerviolet:
Looks good! Congrats :)

PR #596, opened 2018-10-21
Setup initial open file diff event from pr description

kyle-rader:
## Summary
When comments on files show up in the PR description, the file has no way of being opened/ opening the full diff of that file.
This PR is to discuss and add a Show Diff / Open File button(s) to the file name bar on comments in the description view.

image

This is a WIP.

### DIFF
--- a/preview-src/index.ts
+++ b/preview-src/index.ts
@@ -231,6 +231,19 @@ function addEventListeners(pr: PullRequest): void {
 			branch: pr.repositoryDefaultBranch
 		});
 	});
+
+	let fileDiffButtons = document.getElementsByClassName("file-diff-btn")!;
+	console.log(`found ${fileDiffButtons.length} file diff buttons`);
+
+	for (let fileDiffButton of Array.from(fileDiffButtons)) {
+		console.log(`Adding click listener to ${fileDiffButton.outerHTML}`);
+		fileDiffButton.addEventListener('click', (e) => {
+			vscode.postMessage({
+				command: 'pr.file-diff',
+				path: (<HTMLElement>e.target).getAttribute('data-path'),
+			});
+		});
+	}
 }
 
 function clearTextArea() {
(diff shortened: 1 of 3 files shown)


kyle-rader:
@RMacfarlane This my first blind attempt at the feature you mentioned in your Github Universe talk to open the file/file-diff from the comment preview in the PR description.

Right now - the command to open the diff view fails, and I'm not sure how to debug further. This approach probably needs other improvements. #IDontKnowWhatI'mDoing :)

rebornix:
@kyle-rader thanks for your help on this. This is an interesting feature but a bit complex. The workflow of opening the diff view in my mind can be 

* Users click on the link of a diff hunk in the description page
* [PullRequestOverviewPanel](https://github.com/microsoft/vscode-pull-request-github/blob/master/src/github/pullRequestOverview.ts#L16) handles the clicking event and analyze which file to open by reading the comment info.
* Based on the comment info, we know which line/column we should scroll to when the editor is being opened.
* Pull Request Tree View should focus on the file change node, which is determined above
* We open the diff editor for the file change node and scroll to the target line/column, which is determined above.

RMacfarlane:
Thanks @kyle-rader, sorry for the delay in looking at this! This looks like a great start, you've correctly added something to the Description page and sending the message back to `pullRequestOverview.ts` is correct.

It looks like we've made things harder for you by having the wrong types on the `pr.openDiffView` command. The first two arguments, `parentFilePath` and `filePath`, should be type `vscode.Uri` instead of string. You can take the file path and convert it into this using the `toPrUri` function.

RMacfarlane:
Hey @kyle-rader, thanks for giving this a shot! We implemented this feature with cfcb1f74f06669fe42087216f2b0b10be0285d65 so I will close this out.

kyle-rader:
@RMacfarlane thanks for the update - sorry I never got back to it. Glad to see it got in though!

PR #600, opened 2018-10-22
Edit/delete comments within the editor

RMacfarlane:
comment_edit

use the new APIs to allow editing and deleting comments. I'm restricting edit/delete to only non-outdated comments that match the logged in user, but we could relax that if needed.

I'll send a separate PR for adding edit/delete actions to comments on the description page

### DIFF
--- a/src/github/credentials.ts
+++ b/src/github/credentials.ts
@@ -20,6 +20,7 @@ const AUTH_INPUT_TOKEN_CMD = 'auth.inputTokenCallback';
 
 export class CredentialStore {
 	private _octokits: Map<string, Octokit>;
+	private _logins: Map<string, string>;
 	private _configuration: VSCodeConfiguration;
 	private _authenticationStatusBarItems: Map<string, vscode.StatusBarItem>;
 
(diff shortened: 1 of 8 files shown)


src/github/pullRequestManager.ts
```diff
@@ -394,6 +394,38 @@ export class PullRequestManager implements IPullRequestManager {
 		}
 	}
 
+	async editComment(pullRequest: IPullRequestModel, commentId: string, text: string): Promise<Comment> {
+		const { octokit, remote } = await (pullRequest as PullRequestModel).githubRepository.ensure();
+
+		const ret = await octokit.pullRequests.editComment({
```
rebornix: We can try/catch like other methods

src/github/credentials.ts
```diff
@@ -20,13 +20,15 @@ const AUTH_INPUT_TOKEN_CMD = 'auth.inputTokenCallback';
...
 	private _octokits: Map<string, Octokit>;
+	private _logins: Map<string, string>;
 	private _configuration: VSCodeConfiguration;
 	private _authenticationStatusBarItems: Map<string, vscode.StatusBarItem>;
 
 	constructor(configuration: any,
 		private readonly _telemetry: ITelemetry) {
 		this._configuration = configuration;
 		this._octokits = new Map<string, Octokit>();
+		this._logins = new Map<string, string>();
```
queerviolet: Octokit actually exposes `currentUser`, it's just not shown in our typings.

queerviolet: Oh, lol. My bad, *I* put it on there in #554 😅

I guess I'd still suggest doing that, mostly bc it really seems like it should be there and we can probably pull it upstream.

RMacfarlane: Thanks!

PR #606, opened 2018-10-24
Refactor renderTimelineEvents to return HTMLElements instead of a string

RMacfarlane:
It's significantly easier to handle the comment edit UX with named HTML elements instead of strings. This is the first part of the work to have edit/delete support on comments there, just refactoring to use `document.createElement`

### DIFF
--- a/preview-src/index.ts
+++ b/preview-src/index.ts
@@ -92,13 +92,22 @@ function handleMessage(event: any) {
 window.addEventListener('message', handleMessage);
 
 function renderPullRequest(pr: PullRequest): void {
-	document.getElementById(ElementIds.TimelineEvents)!.innerHTML = pr.events.map(renderTimelineEvent).join('');
+	renderTimelineEvents(pr);
 	setTitleHTML(pr);
 	setTextArea();
 	updateCheckoutButton(pr.isCurrentlyCheckedOut);
 	updatePullRequestState(pr.state);
 
 	addEventListeners(pr);
+}
+
+function renderTimelineEvents(pr: PullRequest): void {
+	const timelineElement = document.getElementById(ElementIds.TimelineEvents)!;
+	timelineElement.innerHTML = '';
+	pullRequest.events
+		.map(event => renderTimelineEvent(event))
+		.filter(event => event !== undefined)
+		.forEach(renderedEvent => timelineElement.appendChild(renderedEvent as HTMLElement));
 }
 
 function updatePullRequestState(state: PullRequestStateEnum): void {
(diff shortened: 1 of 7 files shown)


preview-src/index.ts
```diff
@@ -101,6 +101,15 @@ function renderPullRequest(pr: PullRequest): void {
 	addEventListeners(pr);
 }
 
+function renderTimelineEvents(pr: PullRequest): void {
+	const timelineElement = document.getElementById(ElementIds.TimelineEvents)!;
```
queerviolet: I think it would be easier to just do this once, at the top of the file. Instead of tracking `ElementIds`, we could just query for the elements and retain references to them.

If we go that route, we should probably make our lives easier by changing [`pullRequestOverview.ts`](src/github/pullRequestOverview.ts) to have the `<script>` after all the `<div>`s.

preview-src/pullRequestOverviewRenderer.ts
```diff
@@ -227,57 +227,104 @@ function groupBy<T>(arr: T[], fn: (el: T) => string): { [key: string]: T[] } {
...
-	</div>
-</div>`;
+
+	const commentContainer = document.createElement('div');
+	commentContainer.classList.add('comment-container', 'commit');
+	const commitMessage = document.createElement('div');
+	commitMessage.className = 'commit-message';
+
+	const commitIcon = document.createElement('span');
+	commitIcon.innerHTML = `<svg class="octicon octicon-git-commit" width="14" height="16" viewBox="0 0 14 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
queerviolet: Maybe this could go in `media`?

preview-src/pullRequestOverviewRenderer.ts
```diff
@@ -227,57 +227,104 @@ function groupBy<T>(arr: T[], fn: (el: T) => string): { [key: string]: T[] } {
...
+
+	const userIcon = renderUserIcon(comment.user.html_url, comment.user.avatar_url);
+	const reviewCommentContainer = document.createElement('div');
+	reviewCommentContainer.className = 'review-comment-container';
+	commentContainer.appendChild(userIcon);
+	commentContainer.appendChild(reviewCommentContainer);
+
+	const commentHeader = document.createElement('div');
+	commentHeader.className = 'review-comment-header';
+	const authorLink = document.createElement('a');
```
queerviolet: I don't know what our style is for these, but it somehow seems nicer to me to cast to the appropriate type ASAP.

```suggestion
	const authorLink = <HTMLAnchorElement>document.createElement('a');
```

(It's also a bummer to me that TypeScript can't handle this more gracefully.)

RMacfarlane: Yeah good point! It's silly to keep casting them everywhere instead of just declaring the correct type up front.

changed all instances of `const <x> = document.createElement` to be explicitly typed

preview-src/pullRequestOverviewRenderer.ts
RMacfarlane: I moved it to `resources` instead, since it looks like `media` is the output directory for webpack

preview-src/index.ts
RMacfarlane: Yeah, I agree. I think to make this webview maintainable we should ultimately break it down into classes for different UI components

Right now I don't think we can pull the getElementByIds up to the top of the file, since some of these elements are added as the PR loads. Let's handle this separately

queerviolet: 👍🏽

queerviolet:
LGTM!

Issue #110, opened 2018-07-31
The date on the comment should be a link to the comment


ramya-rao-a:
Testing https://github.com/Microsoft/vscode/issues/55396 on Ubuntu

The date on the comment should be a link to the comment just like on github

PR #629, opened 2018-10-26
Add hyperlinks to timestamps

JeffreyCA:
Resolves https://github.com/Microsoft/vscode-pull-request-github/issues/110

screen shot 2018-10-26 at 11 25 34 am

I also made all hyperlinks underline on hover. Before, some of the hyperlinks were underlined without hovering and some were. To be consistent I made them all not underlined when not hovered.

### DIFF
--- a/preview-src/index.css
+++ b/preview-src/index.css
@@ -494,6 +494,11 @@ code {
 	overflow: auto;
 }
 
+.timestamp,
+.timestamp:hover {
+	color: inherit;
+}
+
 /** Theming */
 
 .comment-body pre code {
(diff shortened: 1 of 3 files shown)


rebornix:
The change looks good to me but IMHO the link for timestamp should not stand out, we may want to still use gray/white for it and show underline when hover.

JeffreyCA:
Done!

screen shot 2018-10-26 at 03 04 53 pm

Are there any other links that should should not be blue?

rebornix:
@JeffreyCA awesome, the code change is neat ;) Right now let's leave other links as they used to be.

PR #659, opened 2018-11-02
Store tokens in the keychain or extension global state and migrate out all our configuration

queerviolet:
This PR deprecates the `githubPullRequests.hosts` configuration, migrating it to a combination of extension `globalState` and the system keychain (if available). If no keychain is available, we'll store tokens in extension storage.

Existing global config entries will be migrated on startup. Workspace config entries won't be. In principle, we used to support per-workspace auth, and we don't anymore. I have my doubts as to whether anyone was using that feature, though.

**New command: `auth.logout`** Shows a quickpick of all hosts for which we currently have a token, letting the user select one or more tokens to delete. (There's no other mechanism for users to manage their tokens, otherwise).

**Changed behavior: `auth.inputTokenCallback`** If the manually-entered string doesn't parse as a callback URI, we treat it as a token, prompting the user for a hostname and then saving it in the keychain. I'm keeping the callback URI option around, since it makes the flow a little easier (the callback URI contains the hostname, so we don't have to ask for it).

**Misc change: `promiseFromEmitter` is now `promiseFromEvent`** I didn't really understand the primacy of events in VS code, thinking instead of the node EventEmitter model. This small change makes the thing a lot more useful.

Goodbye, numerous config objects.


src/extension.ts
```diff
@@ -27,9 +28,9 @@ let telemetry: ITelemetry;
 async function init(context: vscode.ExtensionContext, git: GitAPI, repository: Repository): Promise<void> {
 	Logger.appendLine('Git repository found, initializing review manager and pr tree view.');
 
-	const configuration = new VSCodeConfiguration();
-	await configuration.loadConfiguration();
-	configuration.onDidChange(async _ => {
+	Keychain.init(context);
+	await migrateConfiguration();
```
rebornix: we may want to migrate before we intialize the keychain, otherwise the very first time I use the new version, it asks me to login again.

src/authentication/vsConfiguration.ts
```diff
@@ -1,174 +1,26 @@
...
-		}), true);
+	// With tokens stored in local storage, we don't really have per-workspace
+	// authentication settings anymore. Only port global settings.
+	const hosts = config.inspect(HOSTS_KEY).globalValue;
+	if (!Array.isArray(hosts)) { return; }
+	for (const { host, token } of hosts) {
+		Logger.appendLine(`Migrating ${host} from ${namespace}.${HOSTS_KEY} to keychain`);
+		await migrateToKeychain(host, token);
 	}
+	await config.update(HOSTS_KEY, undefined, vscode.ConfigurationTarget.Global);
```
rebornix: `config` can be `undefined`.

src/extension.ts
queerviolet: `migrateConfiguration` calls `Keychain.migrateToken`, which requires the keychain to be initialized.

I don't see how initializing the keychain causes us to prompt the user? We just set the default values of two locals.

src/authentication/vsConfiguration.ts
queerviolet: Done.

package.json
```diff
@@ -110,6 +110,11 @@
         "title": "Manually Provide Authentication Response",
         "category": "GitHub Pull Requests"
       },
+      {
+        "command": "auth.logout",
+        "title": "Log out of GitHub",
```
RMacfarlane: I think this should be "Sign out" since its opposite command is "Sign in"

src/extension.ts
rebornix: My assumption is wrong here. The extension asks me very often to sign in and it happens the first time as well when I launch the debugging instance with this change but it seems they are unrelated. We need some logging to see why that happens.

package.json
queerviolet: Done.

nicolo-ribaudo:
How does this work? The auth page doesn't provide any fallback.

Issue #620, opened 2018-10-25
Discussing Create PR


kieferrm:
We just looked again @queerviolet's PR. It's a viable path since it's close to the web UI experience.  But it made us think again about how to keep the PR creation experience as close as possible to the VS Code UX. Below is an attempt to capture our discussions starting with restating some goals.

## Goal
- Ensure users don't have to leave the editor to create a PR. (There are steps users have to take today in order to create a PR such as creating a branch, pushing commits to the branch, ensuring that the branch is in a good shape. All of this can be done in the editor today. So, we can start with assuming the same.)
- Integrate tightly with the markdown editing experience. This is actually not limited to "Create PR" since this is also a challenge for comments in general.
- Fit in the overall VS Code experience; i.e. maintain the perceived lightweight nature of VS Code.

## Proposed Flow for first iteration
1. User is reponsible creating branch, push branch, nesure branch is in good shape. Just as they are if they use the web UI.

2. User chooses `Create Pull Request` from command palette. Same as PR.

3. This opens the QuickPick with 3 steps and a `Create` button, and the user fills in the data
   1. Title of the PR
   2. Target repository (dropdown list)
   3. Target branch (dropdown list)

→ This creates the PR on GH.
→ VS Code switches to the same mode as if the user would have checked out a PR, and it opens the Overview page of the PR.

4. User hits the edit button next to the description of the PR

→ Description switches into edit mode.
→ Text field has an affordance to "maximize/flyout".

5. User hits the `Maximize` button.

→ Markdown editor opens on synthesized document (extension is the document provider). Extension listens to all document changes and live updates the draft of the description (same for comments).

6. User uses the markdown text editor to write the description, they can use the markdown preview to see what the markdown would be rendered.

7. User closes the markdown editor (or switches back to the Overview page)

→ Description is the same as it was/is on the markdown editor.

8. User submits/saves draft.

## What we need
- Ensure draft comments are maintained when switching away from the Overview Page. @rebornix is working on this.
- Ensure we can edit. @rmacfarlane is working on this.
- QuickPick: The steps show the same information as the current PRs (see below) just in a linear fashion. The same data is needed to create the QuickPick items.
- Have the "maximize". Not yet taken. Close to what @queerviolet experimented with.

![image 1](https://user-images.githubusercontent.com/4674940/47469477-f12b4f80-d7b5-11e8-94e8-d2080cee3e82.png)

queerviolet:
[...]

PR #661, opened 2018-11-05
Create pull request with quick pick

rebornix:
This PR reflects the recording shared in https://github.com/Microsoft/vscode-pull-request-github/issues/620#issuecomment-433568713. It derived from https://github.com/Microsoft/vscode-pull-request-github/commit/534dd0015cbd0a99c4c09844f4d8aa1161b4fb9b in @queerviolet 's pull request #554 and the code change is minimal 

* Detect local remotes and their default targets, basically that's the logic in #554 
* Users only choose target remote and branch
* We show a progress bar while creating Pull Request
* Switch to Review Mode immediately and open the description view for that PR. We already merged https://github.com/Microsoft/vscode-pull-request-github/pull/643 so users can update the description of the pull request with ease.

https://github.com/Microsoft/vscode-pull-request-github/pull/554 shows all the possibilities of creating pull request: quick pick (modal), webview and `pullrequest.md`, and it inroduces Preact and Redux, which simplies the webview rendering code and improves the testability significantly. For short, we love both the UX and engineering improvement in #554. Instead of going back and forth there, let's branch out and use this pr to bring in the *first iteration* workflow.

@queerviolet I'd love to hear your ideas on how we can bring remote/branch logic from https://github.com/Microsoft/vscode-pull-request-github/pull/554/files#diff-16d81de09f288c2fbde3485db2c8003f into this pr. We can discuss this on Monday's syncup.


rebornix:
@queerviolet if you see anything missing from #554, feel free to push directly to this branch or leave comments, both ways work.

src/common/git.ts
```diff
@@ -0,0 +1,62 @@
...
+const exec = promisify(execFile);
+
+/**
+ * Push a local branch to a remote.
+ *
+ * @param remote name of remote to push to
+ * @param {string?} localBranch name of local branch to push (default: current branch)
+ * @param {string?} remoteBranch name of remote branch (default: name of local branch)
+ */
+export async function push(
```
RMacfarlane: we should be able to use the git extension API for this, it was added recently. would require bumping engine to 1.29

src/github/pullRequestManager.ts
```diff
@@ -397,6 +441,98 @@ export class PullRequestManager implements IPullRequestManager {
...
+		if (!params.title) {
+			params.title = title;
+		}
+
+		if (!params.body) {
+			params.body = body;
+		}
+
+		// Create PR
+		let { data } = await repo.octokit.pullRequests.create(params);
```
RMacfarlane: we should try-catch this here or in the calling code

src/github/githubRepository.ts
```diff
@@ -36,15 +37,20 @@ export class GitHubRepository implements IGitHubRepository {
 	constructor(public remote: Remote, private readonly _credentialStore: CredentialStore) {
 	}
 
-	async resolveRemote(): Promise<void> {
-		try {
-			const { octokit, remote } = await this.ensure();
-			const { data } = await octokit.repos.get({
+	async getMetadata(): Promise<any> {
```
RMacfarlane: in master there's a change to bump the version of the `octokit/rest` library, which adds return type information, so it should be easy to create an interface that extends that type to have `currentUser`

src/common/git.ts
rebornix: yes we should do this.

src/github/githubRepository.ts
rebornix: agree and ideally we can have `currentUser` in `@octokit/rest` typings as well.

src/view/reviewManager.ts
```diff
@@ -1025,6 +1028,51 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+			ignoreFocusOut: true,
+			placeHolder: 'Choose a remote which you want to send pull request to'
+		}).then(async (selected: RemoteQuickPickItem) => {
+			if (!selected) {
+				return;
+			}
+			let selectedRemote = selected.remote;
+			const base: any = await this._prManager.getMetadata(selectedRemote.remoteName);
+			let targets = `${base.owner.login}:${base.default_branch}`;
+			vscode.window.showQuickPick([targets]).then(target => {
```
RMacfarlane:
```ts
if (!target) {
	return
}
```

src/view/reviewManager.ts
```diff
@@ -1025,6 +1028,51 @@ export class ReviewManager implements vscode.DecorationProvider {
 		await this.validateState();
 	}
 
+	public async createPullRequest(): Promise<void> {
+		const branchName = this._repository.state.HEAD.name;
+		let potentialTargetRemotes = this._prManager.getGitHubRemotes();
+		let picks: vscode.QuickPickItem[] = potentialTargetRemotes.map(remote => new RemoteQuickPickItem(remote));
+		vscode.window.showQuickPick(picks, {
+			ignoreFocusOut: true,
+			placeHolder: 'Choose a remote which you want to send pull request to'
```
RMacfarlane: small typo: `send pull request to` -> `send a pull request to`

src/view/reviewManager.ts
```diff
@@ -1025,6 +1028,51 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+		}).then(async (selected: RemoteQuickPickItem) => {
+			if (!selected) {
+				return;
+			}
+			let selectedRemote = selected.remote;
+			const base: any = await this._prManager.getMetadata(selectedRemote.remoteName);
+			let targets = `${base.owner.login}:${base.default_branch}`;
```
RMacfarlane: should we add something like `Choose a base branch` as the placeholder?

src/view/reviewManager.ts
```diff
@@ -1025,6 +1028,55 @@ export class ReviewManager implements vscode.DecorationProvider {
 		await this.validateState();
 	}
 
+	public async createPullRequest(): Promise<void> {
+		const branchName = this._repository.state.HEAD.name;
```
RMacfarlane: right now if you try to create a PR from local branch that hasn't been pushed to the remote, a notification with just "Validation Failed" is shown. should we push the branch in that case?

I think there may also be a problem when the local branch is named differently than the remote branch

src/view/reviewManager.ts
```diff
@@ -1025,6 +1028,55 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+		let potentialTargetRemotes = this._prManager.getGitHubRemotes();
```
queerviolet: We could call `getPullRequestDefaults` here to get the default values for the quickpicks. That'll handle guessing reasonable initial values.

src/view/reviewManager.ts
queerviolet: It doesn't seem like there's a reason to nest `.then`s here, rather than chaining them or using `await`.

src/view/reviewManager.ts
rebornix: for the first iteration, we can show a warning for that case. We didn't decide yet how we are going to handle the `push` workflow.

rebornix:
Addressed comments except https://github.com/Microsoft/vscode-pull-request-github/pull/661#discussion_r230855623 (we can update this after we discuss how to handle `push` like in #554) and https://github.com/Microsoft/vscode-pull-request-github/pull/661#discussion_r230868193 (which can probably be updated in upstream).

src/common/git.ts
RMacfarlane: we can remove this file, not used anywhere

src/github/pullRequestManager.ts
```diff
@@ -748,3 +894,36 @@ export function getEventType(text: string) {
 			return EventType.Other;
 	}
 }
+
+const ownedByMe: Predicate<GitHubRepository> = repo => {
+	const { currentUser=null } = repo.octokit as any;
+	return repo.remote.owner === currentUser.login;
```
RMacfarlane: let's check that `currentUser` exists before dereferencing `login`

src/github/pullRequestManager.ts
```diff
@@ -748,3 +894,36 @@ export function getEventType(text: string) {
...
+		? head.substr(0, idxSep)
+		: params.owner;
+	return byOwnerAndName(owner, repo);
+};
+
+const byOwnerAndName = (byOwner: string, repo: string): Predicate<GitHubRepository> =>
+	({remote: {owner, repositoryName}}) => byOwner === owner && repo === repositoryName;
+
+const titleAndBodyFrom = (message: string): {title: string, body: string} => {
+	const idxLineBreak = message.indexOf('\n');
```
RMacfarlane: would message ever contain `\r\n` line endings?

src/github/pullRequestManager.ts
rebornix: done

src/view/reviewManager.ts
```diff
@@ -1024,6 +1027,65 @@ export class ReviewManager implements vscode.DecorationProvider {
 		await this.validateState();
 	}
 
+	public async createPullRequest(): Promise<void> {
+		const HEAD = this._repository.state.HEAD;
+		if (!HEAD.upstream) {
```
RMacfarlane: `upstream` seems to always be undefined

rebornix: discussed offline, the local branch needs to be `--set-upstream` to the remote one.

hanif147:
https://www.imdb.com/list/ls095865806/

PR #687, opened 2018-11-13
Publish branch first before creating PR.

rebornix:


src/view/reviewManager.ts
```diff
@@ -1026,13 +1026,83 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+
+		try {
+			// since we are probably pushing a remote branch with a different name, we use the complete synatx
+			// git push -u origin local_branch:remote_branch
+			await this._repository.push(selected.remote.remoteName, `${HEAD.name}:${upstreamBranchName}`, true);
+		} catch (err) {
+			if (err.gitErrorCode === GitErrorCodes.PushRejected) {
+				let choice = await vscode.window.showWarningMessage(`Can't push refs to remote, try run 'git pull' first to integrate with your change, or choose another remote or branch name.`, {
+					modal: true
+				}, 'Retry');
```
rebornix: `Retry` is not a good name here, any suggestion?

src/view/reviewManager.ts
```diff
@@ -1026,13 +1026,83 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+			return;
+		}
+
+		try {
+			// since we are probably pushing a remote branch with a different name, we use the complete synatx
+			// git push -u origin local_branch:remote_branch
+			await this._repository.push(selected.remote.remoteName, `${HEAD.name}:${upstreamBranchName}`, true);
+		} catch (err) {
+			if (err.gitErrorCode === GitErrorCodes.PushRejected) {
+				let choice = await vscode.window.showWarningMessage(`Can't push refs to remote, try run 'git pull' first to integrate with your change, or choose another remote or branch name.`, {
```
RMacfarlane: typo - `try running`

src/view/reviewManager.ts
RMacfarlane: I think `Retry` isn't too bad, I would use it

src/view/reviewManager.ts
```diff
@@ -1026,13 +1026,83 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+		const selected: RemoteQuickPickItem = await vscode.window.showQuickPick<RemoteQuickPickItem>(picks, {
+			ignoreFocusOut: true,
+			placeHolder: `Pick a remote to publish the branch ${HEAD.name} to:`
+		});
+
+		if (!selected) {
+			return;
+		}
+
+		let upstreamBranchName = await vscode.window.showInputBox({
```
queerviolet: I'd suggest dropping this prompt and just pushing the branch with the same name. Pushing with a different name is pretty uncommon IME (but maybe there's a use case I'm not aware of?). And anyway, if the user wants that, they can push it themselves.

src/view/reviewManager.ts
```diff
@@ -1026,13 +1026,83 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+		if (!latestBranch || !latestBranch.upstream) {
+			return;
+		}
+
+		return latestBranch;
+	}
+
 	public async createPullRequest(): Promise<void> {
 		const HEAD = this._repository.state.HEAD;
+		let branchName;
```
queerviolet: Does this case ever happen? I think it's possible for us to end up in a repo with no GitHub remotes, right?

```suggestion
const potentialTargetRemotes = this._prManager.getGitHubRemotes();
if (!potentialTargetRemotes.length) {
  vscode.window.showWarningMessage(`No GitHub remotes found. Add a remote and try again.`)
}
let branchName;
```

rebornix: thanks, I wish I can use *Code Suggestion* for this one ;)

rebornix:
@kieferrm and I discussed offline about how we can simplify create pull request workflow for the common cases and how we can leverage QuickPick's API more.

Firstly, when we try to create a pull request, we will ask for pull request related information first, target remote and target branch. If there is only one remote, we don't show the remote picker anymore

create-pr

Secondly, we check whether the branch has upstream set or not. If there is already one, we don't need to anything. If the branch has no upstream, whether it's because we forget to use `-u` flag or we didn't push yet, we will ask if you want to publish the branch.

To publish a branch, we will need two steps at most. The first step is choosing a remote, if there is only one remote, this step can be skipped. The second one is choosing a name for the branch. By default we fillin the local branch name and validate through GitHub whenever the value changes. Once the validation passes, we will publish the branch.

After the branch is published, we will create a pull request immediately.

create-pr-with-push

PR #715, opened 2018-12-04
Remove token verification step.

### LINKED ISSUE #710: Cannot authenticate via URI handler
Issue Type: <b>Bug</b>

1. Click sign in to github in the status bar
2. Follow the steps to authenticate
3. Let uri handler in vs code open authentication uri

:bug: => stuck signing in. the extension does not handle the uri and store the token

Extension version: 0.333.1
VS Code version: Code - Insiders 1.30.0-insider (cd5b4d1bb3fc7e509fba077095b2b280b8ff7161, 2018-11-29T06:32:54.872Z)
OS version: Windows_NT x64 10.0.17763

queerviolet:
Bring in code from the authentication server to parse the callback state parameter, rather than asking the server for it.

Fixes #710 (at least, some instances of it)

### DIFF
--- a/src/authentication/githubServer.ts
+++ b/src/authentication/githubServer.ts
@@ -132,7 +132,7 @@ const SEPARATOR = '/', SEPARATOR_LEN = SEPARATOR.length
  * @returns {any} decoded JSON data
  * @throws {SyntaxError} if the message was null or could not be parsed as JSON
  */
-export const decode = (signedMessage?: string): any => {
+const decode = (signedMessage?: string): any => {
 	if (!signedMessage) { throw new SyntaxError('Invalid encoding'); }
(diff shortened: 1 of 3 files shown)


src/authentication/githubServer.ts
```diff
@@ -120,18 +121,30 @@ class ResponseExpired extends Error {
...
+/**
+ * Hydrate and verify the signature of a message produced with `encode`
+ *
+ * Returns an object
+ *
+ * @param {string} signedMessage signed message produced by encode
+ * @returns {any} decoded JSON data
+ * @throws {SyntaxError} if the message was null or could not be parsed as JSON
+ */
+export const decode = (signedMessage?: string): any => {
```
shana: Is the export needed here?

queerviolet: Nope!

PR #738, opened 2018-12-06
Add inline actions for file changes.

rebornix:
image

The only difference between file changes in PR tree view and SCM is the way we render File Change Type.


src/view/reviewManager.ts
```diff
@@ -63,8 +63,16 @@ export class ReviewManager implements vscode.DecorationProvider {
 		let gitContentProvider = new GitContentProvider(_repository);
 		gitContentProvider.registerTextDocumentContentFallback(this.provideTextDocumentContent.bind(this));
 		this._disposables.push(vscode.workspace.registerTextDocumentContentProvider('review', gitContentProvider));
-		this._disposables.push(vscode.commands.registerCommand('review.openFile', (uri: vscode.Uri) => {
-			let params = fromReviewUri(uri);
+		this._disposables.push(vscode.commands.registerCommand('review.openFile', (value: GitFileChangeNode | vscode.Uri) => {
+			let params: ReviewUriParams;
```
RMacfarlane: What about files that have been deleted? In the SCM changes view, clicking the "Open File" button on such a change is a no-op, maybe we should do the same

rebornix: Modified the context for file change node to contain GitChangeType info and then we can filter deleted out.

src/view/fileTypeDecorationProvider.ts
```diff
@@ -0,0 +1,80 @@
...
+			case GitChangeType.UNKNOWN:
+				return 'U';
+			case GitChangeType.UNMERGED:
+				return 'C';
+		}
+
+		return '';
+	}
+
+	priority(status: GitChangeType): number {
```
RMacfarlane: what does priority do? is this the priority of the decoration if there are other providers registered?

RMacfarlane:
Looks good, just had one question on the decoration provider

rebornix:
@RMacfarlane good catch! Update the priority for all our decorations provider, now

* file change types all have priority 1
* comment decoration has priority 2

For decorations we add to Tree Item, the file change decoration is always placed on the right

image

PR #746, opened 2018-12-10
optimize workflow for checkout to an existing local branch.

rebornix:


src/github/pullRequestGitHelper.ts
```diff
@@ -54,40 +54,45 @@ export class PullRequestGitHelper {
 
 	static async fetchAndCheckout(repository: Repository, remote: Remote, branchName: string, pullRequest: IPullRequestModel): Promise<void> {
 		let remoteName = remote.remoteName;
-		Logger.debug(`Fetch remote ${remoteName}`, PullRequestGitHelper.ID);
-		await repository.fetch(remoteName);
-
 		let branch: Branch;
 
 		try {
 			branch = await repository.getBranch(branchName);
```
RMacfarlane: This looks good, but I wonder if we can simplify the logic more. The `fetchAndCheckout` and `createAndCheckout` methods repeat a lot of the same operations and when I was first reading them I was confused about when each should be called, based on the name I wasn't sure what would be created in `createAndCheckout`.

rebornix: Thanks for your suggestion. We should be able to do that. The way we switch branches right now

* Validate local branch to see if there is already one, and the return value is simply branch name and remote name, which are strings.
  * Fetch the branch and checkout if it exits
  * Create remtoe and then fetch

It works fine if Git is no cost but unluckily that's not true. Instead of having a `if` condition check, which is a costly git operation, we should just fetch the whole config, find all the information needed, if we find something matched, checkout; if not, go to `createAndCheckout`.

RMacfarlane:
I suggest renaming the `getBranchForPullRequestFromExistingRemotes` to `checkoutExistingPullRequestBranch` as well
```diff
diff --git a/src/github/interface.ts b/src/github/interface.ts
index bcd0e81..1c41dd4 100644
--- a/src/github/interface.ts
+++ b/src/github/interface.ts
@@ -159,10 +159,10 @@ export interface IPullRequestManager {
 
 	resolvePullRequest(owner: string, repositoryName: string, pullReuqestNumber: number): Promise<IPullRequestModel>;
 	getMatchingPullRequestMetadataForBranch();
-	getBranchForPullRequestFromExistingRemotes(pullRequest: IPullRequestModel);
+	checkoutExistingPullRequestBranch(pullRequest: IPullRequestModel): Promise<boolean>;
 	getBranch(remote: Remote, branchName: string): Promise<Branch>;
 	checkout(branchName: string): Promise<void>;
-	fetchAndCheckout(remote: Remote, branchName: string, pullRequest: IPullRequestModel): Promise<void>;
+	fetchAndCheckout(pullRequest: IPullRequestModel): Promise<void>;
 }
 
 export interface ITelemetry {
diff --git a/src/github/pullRequestManager.ts b/src/github/pullRequestManager.ts
index e72071a..aacd13a 100644
--- a/src/github/pullRequestManager.ts
+++ b/src/github/pullRequestManager.ts
@@ -835,11 +835,11 @@ export class PullRequestManager implements IPullRequestManager {
 		return matchingPullRequestMetadata;
 	}
 
-	async getBranchForPullRequestFromExistingRemotes(pullRequest: IPullRequestModel) {
+	async checkoutExistingPullRequestBranch(pullRequest: IPullRequestModel): Promise<boolean> {
 		return await PullRequestGitHelper.checkoutExistingPullRequestBranch(this.repository, this._githubRepositories, pullRequest);
 	}
 
-	async fetchAndCheckout(remote: Remote, branchName: string, pullRequest: IPullRequestModel): Promise<void> {
+	async fetchAndCheckout(pullRequest: IPullRequestModel): Promise<void> {
 		await PullRequestGitHelper.fetchAndCheckout(this.repository, this._githubRepositories, pullRequest);
 	}
 
diff --git a/src/view/reviewManager.ts b/src/view/reviewManager.ts
index b2c5a25..ad9f0cc 100644
--- a/src/view/reviewManager.ts
+++ b/src/view/reviewManager.ts
@@ -1005,10 +1005,10 @@ export class ReviewManager implements vscode.DecorationProvider {
 		this.statusBarItem.show();
 
 		try {
-			let localCheckout = await this._prManager.getBranchForPullRequestFromExistingRemotes(pr);
+			const didLocalCheckout = await this._prManager.checkoutExistingPullRequestBranch(pr);
 
-			if (!localCheckout) {
-				await this._prManager.fetchAndCheckout(localCheckout.remote, localCheckout.branch, pr);
+			if (!didLocalCheckout) {
+				await this._prManager.fetchAndCheckout(pr);
 			}
 		} catch (e) {
 			Logger.appendLine(`Review> checkout failed #${JSON.stringify(e)}`);

```

RMacfarlane:
How about this?
```diff
diff --git a/src/github/pullRequestGitHelper.ts b/src/github/pullRequestGitHelper.ts
index 0a6db95..a79ea85 100644
--- a/src/github/pullRequestGitHelper.ts
+++ b/src/github/pullRequestGitHelper.ts
@@ -53,17 +53,12 @@ export class PullRequestGitHelper {
 	}
 
 	static async fetchAndCheckout(repository: Repository, githubRepositories: GitHubRepository[], pullRequest: IPullRequestModel): Promise<void> {
-		let remote: Remote;
-		let branchName: string;
-		let headRemote = PullRequestGitHelper.getHeadRemoteForPullRequest(repository, githubRepositories, pullRequest);
-		if (headRemote) {
-			// the head of the PR is in this repository (not fork), we can just fetch
-			remote = headRemote;
-			branchName = pullRequest.head.ref;
-		} else {
+		const remote = PullRequestGitHelper.getHeadRemoteForPullRequest(repository, githubRepositories, pullRequest);
+		if (!remote) {
 			return PullRequestGitHelper.createAndCheckout(repository, pullRequest);
 		}
 
+		const branchName = pullRequest.head.ref;
 		let remoteName = remote.remoteName;
 		let branch: Branch;
 

```

rebornix:
I think the code is in a better shape now, a bunch of dead code paths are removed. Previously we tried to protect every case which leads to a lot unnecessary code. Now the code path is simple

* Swtich a pull request branch. Before we finish switching, there will be no `updateState`.
* Check if we already have the pull request branch already, if so, checkout
* Now we are switching to a new pull request. Create the remote if necessary, and then fetch the branch, checkout
* We switch the review mode, trigger a `updateState`.

src/github/pullRequestGitHelper.ts
```diff
@@ -27,121 +27,81 @@ export interface PullRequestMetadata {
 export class PullRequestGitHelper {
 	static ID = 'PullRequestGitHelper';
 	static async createAndCheckout(repository: Repository, pullRequest: IPullRequestModel) {
```
RMacfarlane: how about renaming this something like `checkoutFromFork`?

rebornix: +1. done

src/view/reviewManager.ts
```diff
@@ -1042,6 +1057,8 @@ export class ReviewManager implements vscode.DecorationProvider {
 		}
 
 		this._telemetry.on('pr.checkout');
+		Logger.appendLine('Review> switch to Pull Request #${pr.prNumber} - done', ReviewManager.ID);
+		this.switchingToReviewMode = false;
 		await this._repository.status();
```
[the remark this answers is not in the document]
rebornix: This is used to trigger an `updateState` immediately instead of waiting for the event from `this._repository.status`, whose timing is not guaranteed.

RMacfarlane:
I think this needs to be `branchName`:
```diff
diff --git a/src/github/pullRequestGitHelper.ts b/src/github/pullRequestGitHelper.ts
index f603fd1..26add35 100644
--- a/src/github/pullRequestGitHelper.ts
+++ b/src/github/pullRequestGitHelper.ts
@@ -74,7 +74,7 @@ export class PullRequestGitHelper {
 			Logger.appendLine(`Branch ${remoteName}/${branchName} doesn't exist on local disk yet.`, PullRequestGitHelper.ID);
 			const trackedBranchName = `refs/remotes/${remoteName}/${branchName}`;
 			Logger.appendLine(`Fetch tracked branch ${trackedBranchName}`, PullRequestGitHelper.ID);
-			await repository.fetch(remoteName, trackedBranchName);
+			await repository.fetch(remoteName, branchName);
 			const trackedBranch = await repository.getBranch(trackedBranchName);
 			// create branch
 			await repository.createBranch(branchName, true, trackedBranch.commit);

```

src/github/pullRequestGitHelper.ts
```diff
@@ -26,122 +26,82 @@ export interface PullRequestMetadata {
...
+		let remoteName = await PullRequestGitHelper.createRemote(repository, pullRequest.remote, pullRequest.head.repositoryCloneUrl);
+		// fetch the branch
+		let ref = `${pullRequest.head.ref}:${localBranchName}`;
+		Logger.debug(`Fetch ${remoteName}/${pullRequest.head.ref}:${localBranchName} - start`, PullRequestGitHelper.ID);
+		await repository.fetch(remoteName, ref);
+		Logger.debug(`Fetch ${remoteName}/${pullRequest.head.ref}:${localBranchName} - done`, PullRequestGitHelper.ID);
+		await repository.checkout(localBranchName);
+		// set remote tracking branch for the local branch
+		await repository.setBranchUpstream(localBranchName, `refs/remotes/${remoteName}/${pullRequest.head.ref}`);
 		let prBranchMetadataKey = `branch.${localBranchName}.${PullRequestMetadataKey}`;
```
RMacfarlane: could use `associateBranchWithPullRequest` method

rebornix: nice catch!

rebornix:
@RMacfarlane all comments addressed ;) thanks for the detailed review.

PR #747, opened 2018-12-10
Try to reduce the number of calls when fetching PRs with multiple remotes present

RMacfarlane:


rebornix:
I'm not sure assuming that users will use `upstream` and `origin` for upstream and forks is a good idea or not, for example our team still needs remotes which are neither `upstream` or `origin` for pull requests. I like your idea of having a list remotes that the extension uses for fetching pull requests.

src/github/pullRequestManager.ts
```diff
@@ -119,6 +124,47 @@ export class PullRequestManager implements IPullRequestManager {
 		}
 	}
 
+	private validateRemotesSetting() {
+		const remotesSetting = vscode.workspace.getConfiguration(SETTINGS_NAMESPACE).get<string[]>(REMOTES_SETTING);
+
+		if (remotesSetting) {
+			return remotesSetting.forEach(remote => {
+				if (!this._githubRepositories.some(repo => repo.remote.remoteName === remote)) {
```
rebornix: Since this setting can be non-workspace setting, if I put `upstream` and `origin` in my user setting, it will show warning to most of my repositories. Probably we should fail slightly (but emitting log to output).

src/github/pullRequestManager.ts
```diff
@@ -119,6 +124,47 @@ export class PullRequestManager implements IPullRequestManager {
...
+		if (remotesSetting) {
+			return remotesSetting.forEach(remote => {
+				if (!this._githubRepositories.some(repo => repo.remote.remoteName === remote)) {
+					vscode.window.showWarningMessage(`No remote with name '${remote}' found. Please update your 'githubPullRequests.remotes' setting.`);
+				}
+			});
+		}
+	}
+
+	private getDisplayedGitHubRemotes(): GitHubRepository[] {
```
rebornix: When I see the function name, I though the return results are `Remote`s but not `GitHubRepository`.

package.json
```diff
@@ -69,6 +69,14 @@
           "default": "info",
           "description": "Logging for GitHub Pull Request extension. The log is emitted to the output channel named as GitHub Pull Request."
         },
+        "githubPullRequests.remotes": {
+          "type": "array",
+          "default": null,
+          "items": {
+            "type": "string"
+          },
+          "description": "List of remotes to fetch pull requests from. If no remotes are specified, will default to `upstream` or `origin` if these remotes are present, or allvalid GitHub remotes."
```
rebornix: `allvalid` missing a space.

RMacfarlane:
My preference would be to deprecate the `githubPullRequests.includeRemotes` setting in favor of `githubPullRequests.remotes` and have the default behavior be:

- If there are remotes named `upstream` and `origin`, initialize githubRepository classes only for those, fetching from `upstream` before origin
- If there's a remote named `origin` with any other remotes not named `upstream`, initialize a githubRepository only for it
- If there is neither `upstream` or `origin`, fetch from all remotes not created by the extension

When the remotes list is set, it should always be honored regardless of whether or not the remote was created by the extension.

Having a remote named `origin` or `upstream` that is created by the extension itself is seems very unlikely, since we name the remote we create after the owner of the fork. And since the user can always explicitly set the remotes list, if the default behavior isn't what they want, it should be easy to fix. Maybe if `upstream` or `origin` aren't found, we can even show a one-time notification asking if they want to set `githubPullRequsets.remotes`.

Does that seem reasonable?

PR #750, opened 2018-12-11
Navigation to diff view from description page

rebornix:
nagivation-diff-view-from-description-page


src/commands.ts
```diff
@@ -223,10 +223,10 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: IP
 		return await prManager.requestChanges(pr, message);
 	}));
 
-	context.subscriptions.push(vscode.commands.registerCommand('pr.openDescription', async (pr: IPullRequestModel) => {
-		const pullRequest = ensurePR(prManager, pr);
+	context.subscriptions.push(vscode.commands.registerCommand('pr.openDescription', async (descriptionNode: DescriptionNode) => {
+		const pullRequest = ensurePR(prManager, descriptionNode.pullRequestModel);
```
RMacfarlane: when opening the description from the command palette or the status bar, descriptionNode can be undefined

rebornix: fixed ;)

rebornix:
Added outdated label to comment reviews and we don't support code navigation if the comment is outdated.

image

image

preview-src/pullRequestOverviewRenderer.ts
```diff
@@ -591,11 +591,24 @@ class ReviewNode {
...
+
 					const diffView: HTMLDivElement = document.createElement('div');
 					diffView.className = 'diff';
 					const diffHeader: HTMLDivElement = document.createElement('div');
 					diffHeader.className = 'diffHeader';
-					diffHeader.textContent = comments[0].path;
+					const diffPath: HTMLSpanElement = document.createElement('span');
+					diffPath.className =  outdated ? 'diffPath outdated' : 'diffPath';
+					diffPath.textContent = comments[0].path;
+					diffPath.addEventListener('click', () => this.openDiff(comments[0]));
```
RMacfarlane: should the event listener only be added if the comment isn't outdated?

src/commands.ts
RMacfarlane: nice

src/github/pullRequestOverview.ts
```diff
@@ -268,6 +275,18 @@ export class PullRequestOverviewPanel {
...
 
+	private openDiff(message: IRequestMessage<{ comment: Comment }>): void {
+		try {
+			const comment = message.args.comment;
+			const prContainer = this._descriptionNode.parent;
+
+			if ((prContainer as TreeNode | Revealable<TreeNode>).revealComment) {
+				(prContainer as TreeNode | Revealable<TreeNode>).revealComment(comment);
+			}
+		} catch (e) {
```
RMacfarlane: log something to debug if reveal fails?

preview-src/pullRequestOverviewRenderer.ts
```diff
@@ -591,11 +591,24 @@ class ReviewNode {
 						});
 					}
 
+					let outdated = comments[0].position !== comments[0].original_position;
```
RMacfarlane: in other places we check if position is null to determine if a comment is outdated. I think we should do that here too

rebornix: done

preview-src/pullRequestOverviewRenderer.ts
rebornix: done

PR #758, opened 2018-12-18
Adopt commit compare api for finding merge base

rebornix:
For VSCode repository, the time used for opening a pull request decreases from around 5s to 500ms on my Windows box in campus ;)

It's a huge save for large repository like VSCode as when you review other people's pull requests, most of the time you don't have their code locally (unless you do a checkout). It means we keep fetching pull request branches and master whenever we open a new PR node or do a refresh

### DIFF
--- a/src/github/pullRequestManager.ts
+++ b/src/github/pullRequestManager.ts
@@ -789,15 +789,6 @@ export class PullRequestManager implements IPullRequestManager {
 		Logger.debug(`Fetch file changes, base, head and merge base of PR #${pullRequest.prNumber} - enter`, PullRequestManager.ID);
 		const { octokit, remote } = await (pullRequest as PullRequestModel).githubRepository.ensure();
 
-		const { data } = await octokit.repos.compareCommits({
-			repo: remote.repositoryName,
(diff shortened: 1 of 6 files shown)


src/github/pullRequestManager.ts
```diff
@@ -785,24 +785,30 @@ export class PullRequestManager implements IPullRequestManager {
...
+			head: `${pullRequest.head.repositoryCloneUrl.owner}:${pullRequest.head.ref}`
 		});
-		let { data } = response;
 
-		while (response.headers.link && octokit.hasNextPage(response.headers)) {
-			response = await octokit.getNextPage(response.headers);
-			data = data.concat(response.data);
+		pullRequest.mergeBase = data.merge_base_commit.sha;
+
+		if (!pullRequest.base) {
```
RMacfarlane: when is `pullRequest.base`  undefined? should this be before the `compareCommits` call?

rebornix: thanks, I already pushed a fix for that. `base` and `head` are missing when we use the search api.

Issue #354, opened 2018-09-02
Provide a way to select only code


NavyAdmiral:
![image](https://user-images.githubusercontent.com/40662484/44951484-da165400-ae3b-11e8-9fa9-40a8efe20807.png)

Much like how GitHub is already doing it. The `-`, `+` and `line number` views in the `PR Description` page should be isolated from the code selection.

NavyAdmiral:
@rebornix why isn't this view using the same view from the default vscode diff view?

rebornix:
@NavyAdmiral the description page is a pure web app in a webview and it's contributed by the extension, which means it can't use vscode's own diff view.

We can probably use Monaco editor for this case.

PR #763, opened 2018-12-21
improve the boxes

RMacfarlane:


NavyAdmiral:
@RMacfarlane https://github.com/Microsoft/vscode-pull-request-github/issues/354 Can this be addressed within this PR? I believe it would make the PR page look better.

rebornix:
It looks much better than before, great work!

My only concern is Author info of single comments stand out too much right now

image

since we stop showing separators between single comments, Author info looks similar to comments and as Author avatar is large, it stands out more than the real comments. Probably we can shrink it a little bit like what .com does.

RMacfarlane:
I decreased the size of the author information in review comments, and added the right indent back again:

screen shot 2019-01-04 at 10 05 35 am

@NavyAdmiral I'll take a look into that soon, but I think it's outside the scope of this PR

PR #767, opened 2019-01-02
Add graphql library

RMacfarlane:
Factored out a bit of Ashi's and my work on adding graphql support. This exposes `query` and `mutate` from the `GitHubRepository` class.


src/github/githubRepository.ts
```diff
@@ -4,35 +4,72 @@
...
+			return null;
+		}
+
+		Logger.debug(`Request: ${JSON.stringify(query, null, 2)}`, GRAPHQL_COMPONENT_ID);
+		const rsp = await gql.query<T>(query);
+		Logger.debug(`Response: ${JSON.stringify(rsp, null, 2)}`, GRAPHQL_COMPONENT_ID);
+		return rsp;
+	}
+
+	mutate = async <T = any>(mutation: MutationOptions): Promise<ApolloQueryResult<T>> => {
```
rebornix: any reason that we don't declare these methods as `async mutate<T = any>(): Promise { }` ? They look foreign to all other methods we have.

src/github/githubRepository.ts
```diff
@@ -4,35 +4,72 @@
...
+	supportsGraphQl(): boolean {
+		return !!(this.hub && this._hub.graphql);
+	}
+
+	public get octokit(): Octokit {
+		return this.hub && this._hub.octokit;
+	}
+
+	query = async <T = any>(query: QueryOptions): Promise<ApolloQueryResult<T>> => {
+		const gql = this.hub && this._hub.graphql;
```
rebornix: We can always read `this.hub` instead of mixing `hub` and `_hub`.

src/github/githubRepository.ts
RMacfarlane: in calling code we were using the same syntax as before with ensure, so something like `const { query } = await (pullRequest as PullRequestModel).githubRepository.ensure();`. When using the method like that, the correct `this` is lost unless using the arrow function syntax or using `bind`.

I agree that these look foreign :/ we could change them to normal syntax and in calling code, just call ensure and then `(pullRequest as PullRequestModel).githubRepository.query(...)` instead

PR #768, opened 2019-01-02
Fix #310. Allow openning description to the side.

### LINKED ISSUE #310: Setting to open pull request description to side
https://github.com/Microsoft/vscode/issues/57301

I think it would work well for me if I could set the pull request description to open to the side, that way I can double click Description, Checkout and then review files in the first editor.

rebornix:
The command shows as an inline action for description node.


src/github/pullRequestOverview.ts
```diff
@@ -43,16 +43,19 @@ export class PullRequestOverviewPanel {
 	private _initialized: boolean;
 	private _scrollPosition = { x: 0, y: 0 };
 
-	public static createOrShow(extensionPath: string, pullRequestManager: IPullRequestManager, pullRequestModel: IPullRequestModel) {
-		const column = vscode.window.activeTextEditor ? vscode.window.activeTextEditor.viewColumn : undefined;
+	public static createOrShow(extensionPath: string, pullRequestManager: IPullRequestManager, pullRequestModel: IPullRequestModel, toTheSide: Boolean = false) {
+		let activeColumn = vscode.window.activeTextEditor ? vscode.window.activeTextEditor.viewColumn : undefined;
+		if (toTheSide && activeColumn) {
```
RMacfarlane: this is kind of a weird case, but when I first tried this, I had the welcome page open instead of an editor, and the webview opened as normal instead of to the side of that. should we always set the `activeColumn` to beside when `toTheSide` is set?

RMacfarlane:
Looks good, I just had one question

rebornix:
@RMacfarlane adjusted as suggested, even though the workflow is still a bit strange sometimes as the webview is singleton.

PR #770, opened 2019-01-04
Make PR title element content editable

RMacfarlane:
Instead of needing to click the pencil icon that appears on hover, let users directly click on and edit the title.

Currently the change gets sent on blur of the title, not sure if it would be expected to also support pressing something like enter.

### DIFF
--- a/preview-src/index.ts
+++ b/preview-src/index.ts
@@ -96,28 +96,37 @@ function renderTitle(pr: PullRequest): HTMLElement {
 	const titleHeader = document.createElement('div');
 	titleHeader.classList.add('description-header');
 
-	const titleBody = document.createElement('div');
-	titleBody.innerHTML = `${pr.title} (<a href=${pr.url}>#${pr.number}</a>)`;
+	const title = document.createElement('span');
+	title.classList.add('title-text');
+	title.textContent = pr.title;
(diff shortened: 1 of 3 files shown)


preview-src/index.ts
```diff
@@ -135,28 +135,37 @@ function renderTitle(pr: PullRequest): HTMLElement {
...
-			pr.title = text;
-			updateState({ title: text });
-			titleBody.innerHTML = `${pr.title} (<a href=${pr.url}>#${pr.number}</a>)`;
-		}
-
-		const actionsBar = new ActionsBar(titleContainer, { body: pr.title, id: pr.number.toString() }, titleBody, messageHandler, updateTitle, 'pr.edit-title');
-		const renderedActionsBar = actionsBar.render();
-		actionsBar.registerActionBarListeners();
-		titleHeader.appendChild(renderedActionsBar);
+		title.contentEditable = 'true';
```
rebornix: The catch with `contentEditable` is we can pasting rich content into the DOM node, for example, pasting an image into it. I have a feeling it's very easy to mess up with it without enough meaningful error message.

BTW, since we don't use `contentEditable` in other places, it's probabbly hard to discover this feature.

PR #774, opened 2019-01-04
Remove IPullRequestManager and IPullRequestModel interfaces

RMacfarlane:
There is only one PullRequestManager implementation, and one PullRequestModel. So these don't really add a layer of abstraction but do take work to maintain. This PR removes them and uses the classes directly


rebornix:
Well, it requires conflict resolution already to merge this PR, good luck 🚀

PR #775, opened 2019-01-04
Display mergeability on PR description page

RMacfarlane:
screen shot 2019-01-04 at 11 57 09 am

I accidentally pushed this commit to master, closing my previous PR 😱 I've reverted it in master and am recreating this so that it actually gets reviewed

### DIFF
--- a/src/github/pullRequestManager.ts
+++ b/src/github/pullRequestManager.ts
@@ -708,8 +708,6 @@ export class PullRequestManager implements IPullRequestManager {
 				assignee: data.assignee,
 				created_at: data.created_at,
 				updated_at: data.updated_at,
-				comments: 0,
-				commits: 0,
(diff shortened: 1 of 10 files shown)


src/github/pullRequestManager.ts
```diff
@@ -606,7 +606,8 @@ export class PullRequestManager implements IPullRequestManager {
 				comments: 0,
 				commits: 0,
 				head: data.head,
-				base: data.base
+				base: data.base,
+				mergeable: true
```
rebornix: It seems it's always `true`?

RMacfarlane: It gets updated when the individual pull request is fetched, it isn't returned when listing all PRs. I'll change the property to optional and just not set it here instead

Issue #792, opened 2019-01-07
Disable Merge PR button when there are conflicts


Tyriar:
ext version: 1.7.1

Enabled:

![image](https://user-images.githubusercontent.com/2193314/50788192-82667a00-126d-11e9-8dc3-2fe4db88ae4c.png)

Disabled:

![image](https://user-images.githubusercontent.com/2193314/50788215-927e5980-126d-11e9-9ae2-6c6b5aa45423.png)

Issue #782, opened 2019-01-07
Provide a link to status check details


Tyriar:
ext version: 1.7.1

I want a details link on the CI check so I can check failures easily

![image](https://user-images.githubusercontent.com/2193314/50786711-226dd480-1269-11e9-8768-8ce43ff6a45a.png)

PR #807, opened 2019-01-11
Fix status check link and disable merge button when there are conflicts

RMacfarlane:
A couple of small fixes for status checks
Fixes https://github.com/Microsoft/vscode-pull-request-github/issues/792, disables merge button when there are conflicts

And add details link with correct status check link https://github.com/Microsoft/vscode-pull-request-github/issues/782

### DIFF
--- a/preview-src/pullRequestOverviewRenderer.ts
+++ b/preview-src/pullRequestOverviewRenderer.ts
@@ -101,6 +101,7 @@ export function renderStatusChecks(pr: PullRequest) {
 	statusCheckInformationContainer.appendChild(statusSummary);
 
 	const statusesToggle = document.createElement('a');
+	statusesToggle.href = '#';
 	statusesToggle.setAttribute('aria-role', 'button');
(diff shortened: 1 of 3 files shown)


preview-src/pullRequestOverviewRenderer.ts
```diff
@@ -135,13 +136,18 @@ export function renderStatusChecks(pr: PullRequest) {
 
 		statusElement.appendChild(state);
 
-		const statusIcon = renderUserIcon(s.url, s.avatar_url);
+		const statusIcon = renderUserIcon(s.target_url, s.avatar_url);
```
rebornix: I merged the timeline events PR so you may need to convert the casing.

rebornix:
The code looks good, but we have some conflicts to solve, sorry for that :(

preview-src/pullRequestOverviewRenderer.ts
RMacfarlane: I think it's actually ok right now, only the REST API is being used to get status checks, it isn't part of the timeline events call

PR #816, opened 2019-01-11
Fix timeline review commit id

rebornix:
As we didn't resolve right commit id from timeline events, navigation from description review diff view to file change won't work as expected.

### DIFF
--- a/src/github/pullRequestManager.ts
+++ b/src/github/pullRequestManager.ts
@@ -17,7 +17,7 @@ import { GitHubManager } from '../authentication/githubServer';
 import { formatError, uniqBy, Predicate, groupBy } from '../common/utils';
 import { Repository, RefType, UpstreamRef, Branch } from '../typings/git';
 import Logger from '../common/logger';
-import { convertRESTPullRequestToRawPullRequest, convertPullRequestsGetCommentsResponseItemToComment, convertIssuesCreateCommentResponseToComment, parseGraphQLTimelineEvents, convertRESTTimelineEvents } from './utils';
+import { convertRESTPullRequestToRawPullRequest, convertPullRequestsGetCommentsResponseItemToComment, convertIssuesCreateCommentResponseToComment, parseGraphQLTimelineEvents, convertRESTTimelineEvents, parseGraphQLComment } from './utils';
 const queries = require('./queries.gql');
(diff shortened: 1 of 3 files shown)


src/github/pullRequestManager.ts
```diff
@@ -381,9 +381,9 @@ export class PullRequestManager {
 			});
 
 			const comments = data.repository.pullRequest.reviews.nodes
-				.map(node => node.comments.nodes.map(comment => this.addCommentPermissions(toComment(comment), remote)))
+				.map(node => node.comments.nodes.map(comment => this.addCommentPermissions(parseGraphQLComment(comment), remote)))
```
RMacfarlane: can we consolidate the `parseGraphQLComment` and `toComment` methods into one?

RMacfarlane: As in, I think we should completely replace `toComment` with `parseGraphQLComment`

RMacfarlane: also since we get `viewerCanUpdate` and `viewerCanDelete`, we don't need to call `addCommentPermissions` here

rebornix: > can we consolidate the parseGraphQLComment and toComment methods into one?

yup and we should then always convert REST to GraphQL but not the other way around

> also since we get viewerCanUpdate and viewerCanDelete, we don't need to call addCommentPermissions here

good catch. BTW, do you think we should rename `canEdit` to `viewerCanUpdate`?

RMacfarlane: `viewerCanUpdate` is a little bit clearer, but I think either are ok

rebornix:
@RMacfarlane removed `toComment` and checked `isPending` in `parseGraphQLComment`, others are not changed for now.

src/github/queries.gql
```diff
@@ -112,7 +112,7 @@ fragment ReviewComment on PullRequestReviewComment {
 	state
 	pullRequestReview { databaseId }
```
RMacfarlane: looks like 	`viewerCanDelete` and `viewerCanUpdate` are on `IssueComment`, but also need to be added here

rebornix: resolved.

PR #825, opened 2019-01-16
Add typings for graphql responses

RMacfarlane:
The `nodeId` property tracking the graphql id of comments was missing for comments normalized from graphql responses, which is needed for replying to comments in a pending review. I fixed this and added typings for all of the graphql responses.

This may be a pain to maintain, it would be great if we could automatically generate it from the `gql` file, but I think it's very useful to have these responses be typed.

### DIFF
--- a/src/github/graphql.ts
+++ b/src/github/graphql.ts
@@ -1,0 +1,175 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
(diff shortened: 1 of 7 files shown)


src/github/graphql.ts
```diff
@@ -0,0 +1,175 @@
+/*---------------------------------------------------------------------------------------------
```
queerviolet: Love this file. It would be *great* if we could autogenerate it from our queries, though. Maybe there's a loader that'll do it for us?

rebornix: I think technically we can https://medium.com/graphql-mastery/generate-javascript-static-types-from-graphql-typescript-and-flow-4d28b46b8d13

PR #828, opened 2019-01-18
use file scheme for file in Changes In Pull Request

rebornix:


src/view/reviewManager.ts
```diff
@@ -29,24 +29,24 @@ import { PullRequestModel } from '../github/pullRequestModel';
 export class ReviewManager implements vscode.DecorationProvider {
 	public static ID = 'Review';
 	private static _instance: ReviewManager;
-	private _documentCommentProvider: vscode.Disposable;
-	private _workspaceCommentProvider: vscode.Disposable;
+	private _documentCommentProvider?: vscode.Disposable;
+	private _workspaceCommentProvider?: vscode.Disposable;
```
RMacfarlane: this can be simplified since the change type is known to be delete here

src/view/reviewManager.ts
```diff
@@ -640,15 +640,17 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+				const uri = vscode.Uri.file(filePath);
 				let changedItem = new GitFileChangeNode(
 					this.prFileChangesProvider.view,
 					pr,
 					change.status,
 					change.fileName,
 					change.blobUrl,
-					toReviewUri(uri, null, null, change.status === GitChangeType.DELETE ? '' : pr.head.sha, false, { base: false }),
-					toReviewUri(uri, null, null, change.status === GitChangeType.ADD ? '' : pr.base.sha, false, { base: true }),
+					change.status === GitChangeType.DELETE ?
```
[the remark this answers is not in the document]
RMacfarlane: I'm still seeing issues comment behavior when browsing the "Changes in Pull Request" tree. Comments are now only showing up on the right side where the scheme is "file", not sure if that's the intended behavior? Finding the matched file still fails since filePath.query is checked before parentFilePath.query, but if those are reordered, then comments get duplicated between the left and ride side

src/view/reviewManager.ts
```diff
@@ -640,15 +660,17 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+						toReviewUri(uri, undefined, undefined, change.status === GitChangeType.DELETE ? '' : pr.head.sha, false, { base: false }) : uri,
```
RMacfarlane: don't need to check `change.status` again here

PR #831, opened 2019-01-21
add review, approval info and linkify

rebornix:
Add more timeline event info

*Merge Event*
image

*Review Event Author Association*
image

*Linkify*
image

image

### DIFF
--- a/src/github/githubRepository.ts
+++ b/src/github/githubRepository.ts
@@ -280,14 +280,14 @@ export class GitHubRepository implements IGitHubRepository, vscode.Disposable {
 		try {
 			Logger.debug(`Fetch pull request ${id} - enter`, GitHubRepository.ID);
 			const { octokit, query, remote } = await this.ensure();
-			let { data } = await octokit.pullRequests.get({
+			let prsResult = await octokit.pullRequests.get({
 				owner: remote.owner,
 				repo: remote.repositoryName,
 				number: id
 			});
 			Logger.debug(`Fetch pull request ${id} - done`, GitHubRepository.ID);
 
-			if (!data.head.repo) {
+			if (!prsResult.data.head.repo) {
 				Logger.appendLine('The remote branch for this PR was already deleted.', GitHubRepository.ID);
 				return;
 			}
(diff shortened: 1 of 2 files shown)


preview-src/pullRequestOverviewRenderer.ts
```diff
@@ -563,6 +608,9 @@ class ReviewNode {
 		userLogin.href = this._review.user.url;
 		userLogin.textContent = this._review.user.login;
 
+		const authorAssociation: HTMLSpanElement = document.createElement('span');
+		authorAssociation.textContent = `(${this._review.authorAssociation.toLocaleLowerCase()})`;
```
RMacfarlane: `authorAssociation` seems to be undefined when using the REST API. Also, when the authorAssociation is `none`, I think we shouldn't render it

src/github/utils.ts
```diff
@@ -171,6 +171,41 @@ export function parseGraphQLComment(comment: ReviewComment): Comment {
...
+		url: graphQLPullRequest.url,
+		number: graphQLPullRequest.number,
+		state: graphQLPullRequest.state,
+		body: graphQLPullRequest.body,
+		bodyHTML: graphQLPullRequest.bodyHTML,
+		title: graphQLPullRequest.title,
+		createdAt: graphQLPullRequest.createdAt,
+		updatedAt: graphQLPullRequest.updatedAt,
+		head: {
+			label: graphQLPullRequest.headRef.name,
```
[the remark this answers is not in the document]
rebornix: It can be null based on the schema, I'd love to use the generated typings with strict null checks ;)

src/github/utils.ts
```diff
@@ -171,50 +171,129 @@ export function parseGraphQLComment(comment: ReviewComment): Comment {
...
+				break;
 		}
 	});
 
-	return events;
+	return ret;
 }
 
-export function convertRESTTimelineEvents(events: any[]): TimelineEvent[] {
+export function convertRESTTimelineEvents(events: any[]): Common.TimelineEvent[] {
```
RMacfarlane: if possible let's also replace `any` with a real type here, `avatar_url` isn't being converted to `avatarUrl`

Tyriar:
Do you get the HTML directly from GitHub or do some conversions? Microsoft/vscode#xyz should also work

rebornix:
@Tyriar we didn't do anything more if it's from github.com

Issue #837, opened 2019-01-24
Editing a pending comment fails


RMacfarlane:
Issue Type: <b>Bug</b>

1. Start a review by adding a comment
2. Try to edit the comment

Editing errors, seems the REST update comment API doesn't support changing pending comments

Extension version: 1.24.0
VS Code version: Code - Insiders 1.31.0-insider (50854c0a705745bbf50982fab3c30fd6bd1f7b19, 2019-01-24T09:09:43.349Z)
OS version: Darwin x64 18.0.0

PR #838, opened 2019-01-24
Fix editing pending comments

RMacfarlane:
Fixes https://github.com/Microsoft/vscode-pull-request-github/issues/837

### DIFF
--- a/src/view/treeNodes/pullRequestNode.ts
+++ b/src/view/treeNodes/pullRequestNode.ts
@@ -564,7 +564,12 @@ export class PRNode extends TreeNode {
 
 	private async editComment(document: vscode.TextDocument, comment: vscode.Comment, text: string): Promise<void> {
 		const fileChange = this.findMatchingFileNode(document.uri);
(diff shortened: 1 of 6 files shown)


src/view/treeNodes/pullRequestNode.ts
```diff
@@ -564,7 +564,12 @@ export class PRNode extends TreeNode {
 
 	private async editComment(document: vscode.TextDocument, comment: vscode.Comment, text: string): Promise<void> {
 		const fileChange = this.findMatchingFileNode(document.uri);
-		const rawComment = await this._prManager.editReviewComment(this.pullRequestModel, comment.commentId, text);
+		const existingComment = fileChange.comments.find(c => c.id.toString() === comment.commentId);
```
queerviolet: May as well leverage what the `==` operator is best at ;)

```suggestion
		const existingComment = fileChange.comments.find(c => c.id == comment.commentId);
```

RMacfarlane: Good point! It seems we have a `tslint` rule against `==` though, so I think I'll keep the cast explicit

PR #839, opened 2019-01-25
Update styles and layout

misolori:
Explored a few ways to update the styles so that the colors and spacing could work better (IMO). I had to re-arrange the order of a few classes in `preview-src/index.ts`, let me know if there are any issues in the way I did this. Here's a summary of the proposed changes :

- Moved the avatar in the title to the description header (this matches GitHub's style)
- Moved the created timestamp to the description header
- Updated the comment border and background color
- Fixed a bug in the buttons when re-sized smaller
- Fixed an alignment and color issue in the tag labels
- Made the diff background colors slightly less opaque to make them readable in other themes
- Updated "Outdated" label to be themeable
- Removed borders in diff comments to read better

## Before

image

## After

image

### DIFF
--- a/preview-src/index.css
+++ b/preview-src/index.css
@@ -78,7 +78,6 @@ body .comment-container.review {
 	background-color: var(--vscode-editor-background);
 }
 
-
 .review-comment-container {
 	width: 100%;
 	display: flex;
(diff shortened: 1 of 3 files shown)


misolori:
@RMacfarlane thanks for the feedback, given that you've made a lot of changes to this layout I'll close this one and create a separate PR for the compact commits and outdated label.

Separate from this, we'll need to address the color contrast ratio for the links in the comment header since it doesn't pass in our default dark theme (light theme just passes).

PR #858, opened 2019-01-29
Do not fail if scopes fail to validate

### LINKED ISSUE #662: 503 on vscode-auth.github.com when signing in to GitHub Enterprise
- Extension version: 0.2.3
- VSCode Version: 1.27.2
- OS: macOS 10.14.1 

Steps to Reproduce:

1. Open Code with a GitHub Enterprise repo open
2. Click "sign in" on the extension pop-up
3. Click green "Continue" button
4. Wait for a while (60s?)
5. See a Heroku 503 page (see below)

image

### LINKED ISSUE #777: "you are now signed in" but "Token is no longer valid for host" for GitHub Enterprise
I see the following messages when I enter my token and server name

image

and the following messages in output window

> Authentication> Token is no longer valid for host https://github.example.com.

- Extension version: 0.3.2
- VSCode Version: 1.30.1
- OS: Windows 7

Steps to Reproduce:

1.  click on bottom left "sign in to github.factset.com"
2. enter token and server name as prompted

I'm not sure what is going wrong. I don't use https://vscode-auth.github.com to authenticate because I get the same error as #662

queerviolet:
Also add some logging so we can see what's going on for some GHE users (#662, #777).

I suspect we're hitting a proxy server, but they may also be getting tokens without the scopes we want. Stop failing in this case—a partially working token is better than failing outright all the time, presumably.

Fixes #662
Fixes #777

### DIFF
--- a/src/authentication/githubServer.ts
+++ b/src/authentication/githubServer.ts
@@ -92,15 +92,22 @@ export class GitHubManager {
 
 	public static validateScopes(host: vscode.Uri, scopes: string): boolean {
 		if (!scopes) {
-			return false;
+			Logger.appendLine(`[SKIP] validateScopes(${host.toString()}): No scopes available.`);
+			return true;
 		}
 		const tokenScopes = scopes.split(', ');
-		return this.AppScopes.every(x =>
+		const scopesNotFound = this.AppScopes.filter(x => !(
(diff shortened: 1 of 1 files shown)


rebornix:
I resolved the conflicts and will merge it to master to allow us early dogfood the change.

PR #899, opened 2019-01-30
Test Pull Request

ahchen:
Some description text here

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,6 +1,10 @@
 /*---------------------------------------------------------------------------------------------
+ *
+ *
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -1,6 +1,10 @@
 /*---------------------------------------------------------------------------------------------
+ *
+ *
```
ahchen: Example comment

src/extension.ts
```diff
@@ -95,4 +100,4 @@ export async function deactivate() {
 	if (telemetry) {
 		await telemetry.shutdown();
 	}
-}
```
ahchen: Example comment #2

PR #900, opened 2019-01-31
allow mentions api call to fail partially

rebornix:
I failed to see mentions in VSCode repo, probably because in Microsoft org there are always too many mentionable users and the api call fails unpredictably, and when it fails, its error message is not operational

```
GitHubRepository> Unable to fetch mentionable users: Error: GraphQL error: Something went wrong while executing your query. Please include `4798:2E6F:A16534:C78174:5C523C7A` when reporting this issue.
```

While fetching all pages of mentionable users, we should return mentionable users we receice already instead of return empty result.

Since the mentionable users we get from GitHub is not longer complete, we should now inject Timeline Event related users' alias into the completion list if they are there yet.

### DIFF
--- a/src/github/graphql.ts
+++ b/src/github/graphql.ts
@@ -140,6 +140,7 @@ export interface TimelineEventsResponse {
 			}
 		}
 	};
+	rateLimit: RateLimit;
 }
 
 export interface PendingReviewIdResponse {
@@ -148,6 +149,7 @@ export interface PendingReviewIdResponse {
 			nodes: Review[];
 		}
 	};
+	rateLimit: RateLimit;
 }
 
 export interface PullRequestCommentsResponse {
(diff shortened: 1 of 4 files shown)


src/github/graphql.ts
```diff
@@ -140,6 +140,7 @@ export interface TimelineEventsResponse {
 			}
 		}
 	};
+	rateLimit: RateLimit;
```
queerviolet: We don't seem to be doing anything with this?

rebornix: I emit this out in debug log as both @RMacfarlane and me ran into Rate Limit once this week, we'd like to know this info while debugging.

queerviolet: Ah, ok.

PR #904, opened 2019-01-31
Fix #843. Restore pending comments when refresh

### LINKED ISSUE #843: Refresh clears comment box
Refs: https://github.com/Microsoft/vscode/issues/67294

Type something in the comment box
Click refresh -> the input is lost

refresh

rebornix:


preview-src/cache.ts
```diff
@@ -38,6 +38,14 @@ export function getState(): PullRequest {
 }
 
 export function setState(pullRequest: PullRequest): void {
+	let oldPullRequest = getState();
```
queerviolet: Could we just save the pendingCommentText as the draft for that comment? It looks like then it'll get restored automatically?

rebornix: It's for the content in the bottom textarea, which is not bind to any comment yet.

Issue #914, opened 2019-02-04
"Files" links in Changes in Pull Request view throws an error


Tyriar:
Clicking any entry will throw an error:

<img width="319" alt="screen shot 2019-02-04 at 7 30 49 am" src="https://user-images.githubusercontent.com/2193314/52218122-cfc70e80-284e-11e9-810f-17aa04622184.png">

<img width="452" alt="screen shot 2019-02-04 at 7 29 59 am" src="https://user-images.githubusercontent.com/2193314/52218099-c2aa1f80-284e-11e9-872d-fa645d337740.png">

PR #920, opened 2019-02-05
Handle git show failing when trying to display file content

RMacfarlane:
Fixes https://github.com/Microsoft/vscode-pull-request-github/issues/914

`repository.show` will throw when the commit isn't local, so when the local branch is behind the remote it tracks. Add a try-catch for this and also add auto-pulling logic to switching to an existing branch that is behind its remote

### DIFF
--- a/src/view/gitContentProvider.ts
+++ b/src/view/gitContentProvider.ts
@@ -29,9 +29,13 @@ export class GitContentProvider implements vscode.TextDocumentContentProvider {
 		}
 
 		const absolutePath = pathLib.join(this.repository.rootUri.fsPath, path).replace(/\\/g, '/');
-		let content = await this.repository.show(commit, absolutePath);
-
-		if (!content) {
+		let content: string;
+		try {
+			content = await this.repository.show(commit, absolutePath);
+			if (!content) {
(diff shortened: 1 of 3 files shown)


src/view/gitContentProvider.ts
```diff
@@ -29,9 +29,13 @@ export class GitContentProvider implements vscode.TextDocumentContentProvider {
 		}
 
 		const absolutePath = pathLib.join(this.repository.rootUri.fsPath, path).replace(/\\/g, '/');
-		let content = await this.repository.show(commit, absolutePath);
-
-		if (!content) {
+		let content: string;
+		try {
+			content = await this.repository.show(commit, absolutePath);
+			if (!content) {
```
rebornix: `content` can be `''`, which makes `!content` true.

RMacfarlane: yeah I'm not sure if we handle the case where the file content is actually '' correctly. this should be the same behavior as before, where we use the fallback to see if the file has been deleted or added

RMacfarlane:
cool, the build error should be fixed by the other PR. I'll rebase this one after that is merged

PR #926, opened 2019-02-06
Fix typing for mutate, fixes build error

RMacfarlane:


src/github/pullRequestManager.ts
```diff
@@ -753,14 +753,18 @@ export class PullRequestManager {
...
+		const { data, errors } = await mutate<DeleteReviewResponse>({
 			mutation: queries.DeleteReview,
 			variables: {
 				input: { pullRequestReviewId: pendingReviewId }
 			}
 		});
 
-		const { comments, databaseId } = data.deletePullRequestReview.pullRequestReview;
+		if (errors) {
+			throw new Error(errors.map(err => err.message).join(', '));
```
rebornix: `err` leads a noImplicitAny error.

RMacfarlane: huh, I'm not seeing that when I build

rebornix: My ts server didn't pick the typings, reloading the project fixes the problem. Sorry for the false alarm.

PR #944, opened 2019-02-12
Add setting for views location

RMacfarlane:
views-setting


rebornix:
LGTM. We may show some notification if activity bar is hidden the first time, otherwise it looks like the extension is not working anymore.

PR #953, opened 2019-02-14
add label colors

rebornix:
Add label colors. Since GitHub API doesn't tell us about the right foreground color to use, we do the math ourselves.

### DIFF
--- a/src/github/interface.ts
+++ b/src/github/interface.ts
@@ -56,6 +56,7 @@ export interface IGitHubRef {
 
 export interface ILabel {
 	name: string;
+	color: string;
 }
 
 export interface PullRequest {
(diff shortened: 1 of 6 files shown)


GeorgeWL:
I think that colour shouldn't be a required field though. Colour should be a `color?:string` interface.

That way if someone has no interest in colouring the labels, they can keep on using the current implementation.

src/github/interface.ts
```diff
@@ -56,6 +56,7 @@ export interface IGitHubRef {
 
 export interface ILabel {
 	name: string;
+	color: string;
```
GeorgeWL: I think that colour shouldn't be a required field though. Colour should be a color?:string interface.

That way if someone has no interest in colouring the labels, they can keep on using the current implementation.

src/github/pullRequestModel.ts
```diff
@@ -75,7 +75,7 @@ export class PullRequestModel {
 		this.bodyHTML = prItem.bodyHTML;
 		this.html_url = prItem.url;
 		this.author = prItem.user;
-		this.labels = (prItem.labels || []).map(label => label.name);
+		this.labels = (prItem.labels || []);
```
GeorgeWL: what if they've only set name and not the color?

alexr00:
In the time since this PR was opened we've added label colors.

PR #954, opened 2019-02-14
Add/delete reactions

rebornix:
In next Insider build (as there is no Insider build for today), you should be able to add/delete comment from the comment title action bar or inline

kapture 2019-02-14 at 15 06 13


src/github/pullRequestManager.ts
```diff
@@ -844,6 +844,44 @@ export class PullRequestManager {
 		return parseGraphQLComment(comment);
 	}
 
+	async addCommentReaction(pullRequest: PullRequestModel, graphNodeId: string, reaction: vscode.CommentReaction): Promise<AddReactionResponse> {
+		let reactionEmojiToContent = getReactionGroup().reduce((prev, curr) => {
+			prev[curr.label] = curr.title;
+			return prev;
+		}, {} as { [key: string]: string });
+		const { mutate } = await pullRequest.githubRepository.ensure();
+		const { data } = await mutate<AddReactionResponse>({
```
RMacfarlane: Do we intend to also support reactions for GHE?

src/view/reviewDocumentCommentProvider.ts
```diff
@@ -718,6 +726,57 @@ export class ReviewDocumentCommentProvider implements vscode.DocumentCommentProv
 		});
 	}
 
+	public async addReaction(document: vscode.TextDocument, comment: vscode.Comment, reaction: vscode.CommentReaction) {
+		await this.editReaction(document, comment, reaction, true);
+	}
+
+	public async deleteReaction(document: vscode.TextDocument, comment: vscode.Comment, reaction: vscode.CommentReaction) {
```
RMacfarlane: I keep getting `Deleting the comment reaction failed: Illegal argument: line must be non-negative.` when trying to remove a comment while a PR is checked out

src/github/pullRequestManager.ts
rebornix: We don't. This API won't be called when there is no graphql support.

src/view/reviewDocumentCommentProvider.ts
rebornix: I digged into this bug and turns out to be not related to this pr. It's triggerred by the comments update service, I'll work on another PR to fix this.

rebornix:
@RMacfarlane thanks for the review! The error is triggerred by another bug we introduced before so I'd like to fix that separately.

PR #951, opened 2019-02-13
Display reviewers on description page

RMacfarlane:
reflow-sidebar

Displays reviewers and labels in a sidebar. Will tackle adding new reviewers and labels in a separate PR

Issue #412, opened 2018-09-11
Add reviewers within the description panel of PR


sgyio:
When #283 is done, it would be nice to then add reviewers to the PR.  For now this kind of operation need to be done leaving the editor.

PR #955, opened 2019-02-15
Support adding/deleting reviewers and labels

RMacfarlane:
Builds on top of #951 
https://github.com/Microsoft/vscode-pull-request-github/issues/412

Adding uses the quickpick:
add_reviewer

Icon to remove appears on hover when the view is shown as a sidebar:
delete_reviewer

When the window is less than 900px wide, content looks like, with delete icons always shown:
screen shot 2019-02-15 at 4 08 07 pm


rebornix:
Run into following issues

* Open PR description of https://github.com/auchenberg/pullrequest-demo/pull/45
* Try to add a label
* Labels already assigned are not shown on the list
* Add one (say `question`)
* `question` is added successfully
* Click the + icon again, `question` is still on the list.


* Open PR description of https://github.com/auchenberg/pullrequest-demo/pull/45
* Split the editor
* Click the + icon to add a label, nothing happens (I think the quick pick shows up and dismisses again)
* Repeat the steps again and this time try to remove a label, it works as expected

src/github/pullRequestManager.ts
```diff
@@ -1249,6 +1276,51 @@ export class PullRequestManager {
...
+			owner: remote.owner,
+			repo: remote.repositoryName,
+			number: pullRequest.prNumber,
+			reviewers: [ reviewer ]
+		});
+	}
+
+	async addLabels(pullRequest: PullRequestModel, labels: string[]): Promise<void> {
+		const { octokit, remote } = await pullRequest.githubRepository.ensure();
+		octokit.issues.addLabels({
```
rebornix: do we want to `await` this call?

src/github/pullRequestManager.ts
```diff
@@ -1249,6 +1276,51 @@ export class PullRequestManager {
...
+			owner: remote.owner,
+			repo: remote.repositoryName,
+			number: pullRequest.prNumber,
+			labels
+		});
+	}
+
+	async removeLabel(pullRequest: PullRequestModel, label: string): Promise<void> {
+		const { octokit, remote } = await pullRequest.githubRepository.ensure();
+		octokit.issues.removeLabel({
```
rebornix: same here.

src/github/pullRequestManager.ts
RMacfarlane: yeah, nice catch

RMacfarlane:
Fixed the label filtering problem, the quickpick issue seems to be a problem with the webview getting refocused, so I think it might be an upstream issue

PR #960, opened 2019-02-15
Fix tests

RMacfarlane:
Some libraries were missing from the `tsconfig.json`, which was causing typescript compilation to fail. The `queries.gql` file also wasn't getting copied to the output folder, so when running the tests couldn't be found. And some tests needed correcting after the strict null work. Now all tests run and pass, yay!


rebornix:
Great work! Do you think we can turn on tests on nightly ci?

RMacfarlane:
Yeah, I think that's a good idea

PR #962, opened 2019-02-17
Disable not available merge methods (fixes #946)

### LINKED ISSUE #946: Disable PR merge options are not allowed for repo
This is feature request related to #593 
Repos might remove option to merge commits and the extension now supports squash and rebase merges.

However, it would be nice to hide or disable Merge option if it is not available.

IllusionMH:
This PR adds check for merge methods that are allowed in repository and shows only available methods in dropdown.
If method that user set as their default is disabled for current repo - first available will be selected instead.

Current implementation differs from GitHub which shows in dropdown all options and marks options that are not enabled for repo. Should I change behavior to match GitHub?

### DIFF
--- a/preview-src/cache.ts
+++ b/preview-src/cache.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { vscode } from './message';
-import { PullRequestStateEnum, IAccount, ReviewState, ILabel } from '../src/github/interface';
+import { PullRequestStateEnum, IAccount, ReviewState, ILabel, MergeMethod, MergeMethodsAvailability } from '../src/github/interface';
 import { TimelineEvent } from '../src/common/timelineEvent';
 import { ReposGetCombinedStatusForRefResponse } from '@octokit/rest';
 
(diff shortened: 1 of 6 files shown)


preview-src/cache.ts
```diff
@@ -30,6 +30,7 @@ export interface PullRequest {
 	status: ReposGetCombinedStatusForRefResponse;
 	mergeable: boolean;
 	defaultMergeMethod: string;
+	availableMergeMethods: string[];
```
IllusionMH: Should I create separate union type for available merge methods instead?
If yes - which file is the best to add this type?

RMacfarlane: I think that would be good to have - `src/github/interface.ts` is a good place for that

IllusionMH:
@RMacfarlane thank you for review! Will update PR this week.

IllusionMH:
Rebased branch and rewrote code to show arguments and only disable ones that were not enabled for repository.

IllusionMH:
@RMacfarlane thank you!

StanleyGoldman:
Does this fix #967?
I haven't tested yet..

IllusionMH:
@StanleyGoldman no, PR was created before #967. This change only checks current repo options but not user permissions for the repo.

I was looking into #967, but it had an assignment and would change scope of initial PR. Is if free to pick?

StanleyGoldman:
Ah, I read the title of your pull request and it sounded similar.
I was going to tackle it, but I'm working on something else at the moment...
If you'd like to take a stab at it, please do.

Issue #965, opened 2019-02-19
Explore consolidating "GitHub Pull Requests" and "Changes in Pull Request" tree view


RMacfarlane:
Currently, when you check out a PR, an additional tree view called "Changes in Pull Request" appears, which lists file changes and commits for the PR. The file changes use local commit and file information to display changes, and have an inline action to directly jump to the local file.

Since the "Changes in Pull Request" tree view is collapsed when added, it's hard to discover, and shows somewhat redundant information. When I'm reviewing something, I usually start working through the file changes in the "GitHub Pull Requests" tree, then want to jump to the local file, so expand the "Changes in Pull Request" tree and have to find the same place in the file list.

The idea of having a separate tree was to focus in on a particular PR when its checked out, but since there isn't API for reordering tree views in a viewlet and easily controlling collapse/expand, it's worth seeing if creating a consolidated tree view is a better work flow

RMacfarlane:
The tree could either maintain the "Files" and "Commits" category nodes that are currently shown in the "Changes in Pull Request" tree:
![screen shot 2019-02-19 at 11 03 59 am](https://user-images.githubusercontent.com/3672607/53040383-1e230280-3436-11e9-8384-c5b8926aed97.png)


Or only add a commits node:
![add_commits](https://user-images.githubusercontent.com/3672607/53040622-aef9de00-3436-11e9-9a90-72897682abc3.gif)

RMacfarlane:
After exploring this idea, I think this creates a worse user experience. When returning to VSCode after checking out a pull request, it's harder to resume work, since the state of the pull requests tree is lost. Having a separate tree for the current pull request solves this and helps indicate that the current branch is associated with a pull request. Also, now that the extension defaults to a separate view container, the "Changes in Pull Request" tree gets automatically expanded on checkout, which helps discoverability

PR #966, opened 2019-02-20
Consolidate tree views

RMacfarlane:
https://github.com/Microsoft/vscode-pull-request-github/issues/965

Removes the "Changes in Pull Request" tree view


rebornix:
The code looks good in overall (I'll take detailed review later) but it seems we didn't handle restoring from review mode when first launch. For example, we checkout to a branch and start reviews, and then reload the window, navigate to the GitHub viewlet, all trees are collapsed.  It requries two expands at least (expand category first, find the pr with checkmark if it's not on the second page, expand that pr) to review the code changes again.

src/github/pullRequestManager.ts
```diff
@@ -98,6 +99,9 @@ export const onDidSubmitReview: vscode.Event<Comment[]> = _onDidSubmitReview.eve
 export class PullRequestManager {
 	static ID = 'PullRequestManager';
 	private _activePullRequest?: PullRequestModel;
+	public activeFileChanges?: GitFileChange[];
+	public activeOutdatedFileChanges?: GitFileChange[];
+	public activeComments?: Comment[];
 	private _credentialStore: CredentialStore;
```
rebornix: Not a fan of having view info in PullRequestManager. We can probably keep the information in `prsTreeDataProvider`. The `ReviewManager` takes care of git status change, and modifying `PullRequestManager.activePullRequest`, `prsTreeDataProvider` then listens to `activePullRequest` change event and reveal/refresh the tree when necessary.

PR #972, opened 2019-02-25
New Comments API

rebornix:
The refactoring of registering comment threas for Pull Request list view is done but review mode part is still in progress, following code needs to be improved

* Register comment threads for `pr`, `review` and `file` scheme files
* Update them properly when there are new comments update
* Update commenting range when document is dirtly
* Dispose them properly


package.json
```diff
@@ -241,7 +241,7 @@
         "command": "pr.openDescription",
         "title": "View Pull Request Description",
         "category": "GitHub Pull Requests",
-        "when": "github:inReviewMode"
+        "when": "github:inReaMode"
```
rebornix: Typo, please revert.

package.json
```diff
@@ -397,7 +432,6 @@
         },
         {
           "command": "pr.refreshPullRequest",
-          "when": "view == pr && viewItem =~ /pullrequest|description/",
           "group": "pullrequest@2"
```
rebornix: It's a hack as context values seem not working when debugging.

src/github/commands.ts
```diff
@@ -0,0 +1,82 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { PullRequestModel } from './pullRequestModel';
+export function getCommentThreadCommands(commentControl: vscode.CommentControl, pullRequestModel: PullRequestModel, inDraftMode: boolean): vscode.Command[] {
+	let commands: vscode.Command[] = [];
+	if (inDraftMode) {
+		commands.push({
```
rebornix: new comment.

src/view/treeNodes/pullRequestNode.ts
```diff
@@ -504,265 +557,188 @@ export class PRNode extends TreeNode {
...
+	// 	await this._prManager.deleteCommentReaction(this.pullRequestModel, matchedRawComment.graphNodeId, reaction);
+	// 	const params = fromPRUri(document.uri);
+	// 	let comments = await this._prManager.getPullRequestComments(this.pullRequestModel);
+	// 	let changedCommentThreads = commentsToCommentThreads(fileChange, comments.filter(cmt => cmt.path === fileChange.fileName && cmt.position !== null), params!.isBase);
+
+		// this._onDidChangeCommentThreads.fire({
+		// 	added: [],
+		// 	changed: changedCommentThreads,
+		// 	removed: []
+		// });
```
RMacfarlane: When the comment controller is disposed, are all of the comment threads associated with it also removed?

src/view/treeNodes/pullRequestNode.ts
```diff
@@ -6,45 +6,39 @@
...
-	fileChanges: (RemoteFileChangeNode | InMemFileChangeNode | GitFileChangeNode)[],
+import { convertToVSCodeComment, createVSCodeCommentThread } from '../../github/utils';
+import { getCommentThreadCommands, getEditCommand, getDeleteCommand } from '../../github/commands';
+
+export function provideDocumentComments(
+	uri: vscode.Uri,
+	isBase: boolean,
+	fileChange: (RemoteFileChangeNode | InMemFileChangeNode | GitFileChangeNode),
+	matchingComments: Comment[],
 	inDraftMode: boolean) {
```
RMacfarlane: `inDraftMode` is unneeded

src/view/treeNodes/pullRequestNode.ts
rebornix: yes, they will all be disposed first before the controller is disposed.

package.json
```diff
@@ -291,6 +291,41 @@
         "title": "Sign in to GitHub",
         "category": "GitHub Pull Requests"
       },
+      {
+          "command": "pr.replyComment",
```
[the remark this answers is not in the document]
rebornix: Removed.

src/view/reviewDocumentCommentProvider.ts
```diff
@@ -55,56 +54,289 @@ function workspaceLocalCommentsToCommentThreads(repository: Repository, fileChan
...
+			let threads: vscode.CommentThread[] = [];
+			allFileChangeWorkspaceCommentThreads.forEach(thread => {
+				threads.push(createVSCodeCommentThread(thread, this._commentController!, this._prManager.activePullRequest!, inDraftMode, this));
+			});
+
+			this._obsoleteFileChangeCommentThreads[fileChange.fileName] = threads;
+		});
+	}
+
+	async initializeDocumentCommentThreadsAndListeners() {
```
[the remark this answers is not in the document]
rebornix: Fixed.

src/typings/vscode.proposed.d.ts
```diff
@@ -766,136 +835,129 @@ declare module 'vscode' {
...
-	}
-
-	interface DocumentCommentProvider {
 		/**
-		 * Provide the commenting ranges and comment threads for the given document. The comments are displayed within the editor.
+		 * The active (focused) [comment input box](#CommentInputBox).
 		 */
-		provideDocumentComments(document: TextDocument, token: CancellationToken): Promise<CommentInfo | undefined>;
+		readonly inputBox?: CommentInputBox;
+		createCommentThread(id: string, resource: Uri, range: Range): CommentThread;
```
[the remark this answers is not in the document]
rebornix: Good suggestion. Moving `comments` into ctor also ensures that we don't see comment editor being expaned by mistake.

src/view/treeNodes/pullRequestNode.ts
rebornix: yes, the way we refresh tree nodes should be improved. Previously it's pull mode so it's okay to dipose all tree nodes and recreate them.

src/github/commands.ts
```diff
@@ -0,0 +1,147 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { PRNode } from '../view/treeNodes/pullRequestNode';
+import { ReviewDocumentCommentProvider } from '../view/reviewDocumentCommentProvider';
+import { CommentHandler } from './utils';
+
+export function getEmptyCommentThreadCommands(thread: vscode.CommentThread, inDraftMode: boolean, handler: CommentHandler, supportGraphQL: boolean): { acceptInputCommand: vscode.Command, additionalCommands: vscode.Command[] } {
+	let commands: vscode.Command[] = [];
+	let acceptInputCommand = {
```
RMacfarlane: I think the title for this command should also change based on whether we are in draft mode like `getCommentThreadCommands` below, either `Add Comment` or `Add Review Comment`

src/github/utils.ts
```diff
@@ -12,21 +12,89 @@ import { parseDiffHunk, DiffHunk } from '../common/diffHunk';
...
+
+export function createVSCodeCommentThread(thread: vscode.CommentThread, commentController: vscode.CommentController, pullRequestModel: PullRequestModel, inDraftMode: boolean, node: PRNode | ReviewDocumentCommentProvider) {
+	let vscodeThread = commentController.createCommentThread(
+		thread.threadId,
+		thread.resource,
+		thread.range!,
+		[]
+	);
+
+	thread.comments.forEach(comment => {
```
RMacfarlane: could move this above and then call createCommentThread with vscodeThread.comments so there's no update event

src/view/treeNodes/pullRequestNode.ts
```diff
@@ -504,265 +571,206 @@ export class PRNode extends TreeNode {
...
+	// #region comment
+	public async createOrReplyComment(thread: vscode.CommentThread) {
+		if (await this._prManager.authenticate() && this.commentController!.inputBox !== undefined) {
+			if (thread.comments.length) {
+				let comment = thread.comments[0] as (vscode.Comment & { _rawComment: Comment });
+				const rawComment = await this._prManager.createCommentReply(this.pullRequestModel, this.commentController!.inputBox!.value, comment._rawComment);
 
-		if (fileChange instanceof RemoteFileChangeNode) {
-			throw new Error('Comments not supported on remote file changes');
+				thread.comments = [...thread.comments, convertToVSCodeComment(rawComment!, undefined)];
```
RMacfarlane: The edit and delete commands are not being added here, we should have another method that will create comments when the thread already exists

src/github/utils.ts
```diff
@@ -12,21 +12,89 @@ import { parseDiffHunk, DiffHunk } from '../common/diffHunk';
...
 		body: new vscode.MarkdownString(comment.body),
-		command: command,
+		selectCommand: command,
 		userName: comment.user!.login,
-		gravatar: comment.user!.avatarUrl,
-		canEdit: comment.canEdit,
-		canDelete: comment.canDelete,
-		isDraft: !!comment.isDraft,
+		userIconPath: vscode.Uri.parse(comment.user!.avatarUrl),
+		label: !!comment.isDraft ? 'Pending' : undefined,
```
RMacfarlane: This seems to be an upstream issue, if I start a review, and then reload vscode, the comments are not initially shown with the 'Pending' label. I see that the label is correctly being set here. If I add another pending comment to the thread, the labels for the thread are rendered

src/view/treeNodes/pullRequestNode.ts
```diff
@@ -504,265 +573,206 @@ export class PRNode extends TreeNode {
...
 	}
 
-	private async deleteComment(document: vscode.TextDocument, comment: vscode.Comment): Promise<void> {
-		const fileChange = this.findMatchingFileNode(document.uri);
-
+	public async deleteComment(thread: vscode.CommentThread, comment: vscode.Comment): Promise<void> {
 		await this._prManager.deleteReviewComment(this.pullRequestModel, comment.commentId);
-		const index = fileChange.comments.findIndex(c => c.id.toString() === comment.commentId);
+		const index = thread.comments.findIndex(c => c.commentId === comment.commentId);
 		if (index > -1) {
```
RMacfarlane: If the comment is the last in the thread, we should also dispose the thread. Ditto in the reviewDocumentCommentProvider

src/github/utils.ts
rebornix: comment relies on `thread` as it's one argument for `edit` and `delete` command.

rebornix: Here I put `comments` in ctor and then update `comments` again with proper edit/delete commands, the reason is if we create comment thread with empty comments, there are chances that VS Code focuses in the comment editor.

rebornix:
@RMacfarlane thanks for your thorough review, I feel confident with our product quality as your review is always high quality ❤️ 

I addressed your comments and fixed a few upstream issues (not in Insiders yet so please wait till Monday). In addition to issues you ran into, I also fixed a few bugs with Review and comments update. Since we didn't merge this pr when I was still in office, I need your help to merge this pr. You may still run into bugs so it would be great if you have a better understanding how comment threads cache is managed in this PR

* PR List View
  * Raw comments are saved in `fileChanges`, which is the same as before
  * When an editor is being opened, we will check if we already create comment threads for the newly opened editor or not. If not, we create comment threads based on `fileChange.comments`.
  * When an editor is being closed, we will dispose all related comment threads.
  * When we refresh the PR, we will compare the cached comment threads for this PR, with the new comments, and update accordingly.
  * **Note**, we only update all comment threads for PR when users attempt to refresh the PR, which is the same behavior as before.
  * When we create/update/delete comments, we need to update `fileChanges.comments` and visible comment threads. 
* Review
  * Raw comments are saved in `fileChanges`, which is the same as before
  * We will create comment threads when we detect the workspace is in review mode.
  * When we open diff view from PR List View or Changes in PR, we will create comment threads when the editor is being opened, and dispose the comment threads when the editor is being closed. The logic is the same as what we have in *PR List View*
  * When we create/update/delete comments, we need to update all visible `review`/`pr` comment threads and all workspace `file` comment threads.

The challenge here is we now moved to `push` mode from previous `pull` mode. It means we need to manage the state/cache of comment threads. It can be simple if we only need to support `file` scheme but life is life, we need to support `pr`, `review` and `file`. So if you run into weird bug, look into how we cache comment threads, usually that's where the bug is from.

The code architecture can be improved by moving the visible editors change listener to `PRDocumentCommentsProvider` instead of registering in `PRNode` and remove any `pr` related code from `ReviewDocumentCommentsProvider` but we don't need to do that now.

RMacfarlane:
Thanks for the clear explanation! I've done some more testing and found some more issues that need to be fixed, seems like a lot of them are upstream.

- Without checking out a PR, open a file and leave a comment. Switch to another file and then back, comment is gone
- Leave some text in a comment without submitting it, switch to another file and do the same. Switch back, the pending comment text is overwritten
- If clicking the gutter icon quickly, mutiple empty comment threads can be created
- No way to remove empty comment threads
- Toggling reaction on something I haven't reacted to tries to remove the reaction and errors out
- Checkout a PR, then return to master. No commenting ranges are shown on any file opened from the tree after that

PR #1020, opened 2019-03-01
[WIP] Changing the Pull Request node icon to represent merge-ability

StanleyGoldman:
Fixes #681

Issue #681, opened 2018-11-12
Show PR review status


rebornix:
- Extension version: Insiders
- VSCode Version: 1.29
- OS: macOS

Now we are using Pull Requests for features and bug fixes always, so I still rely on emails and .com to know if my pr is approved and ready to be merged or not. If we can show indicators of the review status in PRs tree view, I can simply go to category *Created By Me*.

StanleyGoldman:
@queerviolet and I were plotting on possible node structures we are partial to the first, but I thought we could put the others up here for discussion.

``` 
- [Checks Icon] [Review Icon] PR Title #xx by @StanleyGoldman
  - Desc
  - File 1
  - File 2
  - File N

- [Avatar] PR Title #xx by @StanleyGoldman
  - Desc
  - Reviewers
    - Ashi: +1
    - Stanley: -1
  - Checks
    - Check 1: +1
    - Check 2: +1
  - File 1
  - File 2
  - File N
  
- [Avatar] PR Title #xx by @StanleyGoldman
  - Desc
  - Reviewers: +1/-1/M
  - Checks:  +1/-1/M
  - File 1
  - File 2
  - File N
```

StanleyGoldman:
Displaying two icons in the tree would not be ideal from a UX standpoint. We are trying to figure out how to merge both sets of states into one icon.

Checks states:
1. Passing
2. Failture
3. Pending

Review states:
1. Approve
2. Request Changes
3. Review Requested
4. Required

Both of those states are not all of what goes into "merge-ability".

StanleyGoldman:
The simplest answer could very well be to display an icon for merge-ability directly.

queerviolet:
I think there are really three basic states we want to represent:
  - ✅This PR is good to merge
  - ⛔️This PR can't be merged:
    - checks are failing, so fix them
    - a review is required, so assign reviewers
    - a reviewer requested changes, so make them
  - ⚠️Waiting on:
    - A review from a requested reviewer
    - Checks to complete

StanleyGoldman:
Okay, so I've been looking deeper into the initial desire and comparing with other Pull Request centric GitHub plugins do.

Both plugins:
1. Keep the users avatar in a primary position
1. Display Pull Request Checks to the right instead of a Ready to Merge indicator.
  a. The right side margin is also maintained as the panels shrink horizontally

GitHub for Visual Studio:
[...]

PR #1022, opened 2019-03-04
Loading Pull Request List via GraphQL

StanleyGoldman:
In order to implement (#1020), this pull request changes functionality to query for the pull request list via GraphQL. REST API calls to obtain Pull Request list items do not include the `mergeable` flag.


src/view/treeNodes/categoryNode.ts
```diff
@@ -73,17 +73,11 @@ export class PRCategoryActionNode extends TreeNode implements vscode.TreeItem {
 	}
 }
 
-interface PageInformation {
-	pullRequestPage: number;
-	hasMorePages: boolean;
-}
```
StanleyGoldman: This seems to be an artifact from an old implementation.

src/view/treeNodes/categoryNode.ts
```diff
@@ -73,17 +73,11 @@ export class PRCategoryActionNode extends TreeNode implements vscode.TreeItem {
...
-
 export class CategoryTreeNode extends TreeNode implements vscode.TreeItem {
 	public readonly label: string;
 	public collapsibleState: vscode.TreeItemCollapsibleState;
 	public prs: PullRequestModel[];
 	public fetchNextPage: boolean = false;
-	public repositoryPageInformation: Map<string, PageInformation> = new Map<string, PageInformation>();
```
StanleyGoldman: As does this.

src/github/pullRequestManager.ts
```diff
@@ -582,15 +582,81 @@ export class PullRequestManager {
...
+						pullRequestFilter = pr => pr.reviewRequests.nodes
+							.findIndex(reviewRequest => reviewRequest.requestedReviewer.login === currentUserLogin) !== -1;
+					} else if (type === PRType.AssignedToMe) {
+						pullRequestFilter = pr => pr.assignees.nodes
+							.findIndex(asignee => asignee.login === currentUserLogin) !== -1;
+					} else {
+						throw new Error('Unexpected pull request filter');
+					}
+
+					pullRequestItems = pullRequestItems.filter(pullRequestFilter);
```
StanleyGoldman: Client side filtering happening here.

src/github/pullRequestManager.ts
```diff
@@ -582,15 +582,81 @@ export class PullRequestManager {
...
+							body: '',
+							createdAt: pullRequestItem.createdAt,
+							updatedAt: pullRequestItem.updatedAt,
+							merged: pullRequestItem.merged,
+							nodeId: pullRequestItem.nodeId,
+							number: pullRequestItem.number,
+							state: pullRequestItem.state,
+							title: pullRequestItem.title,
+							user: pullRequestItem.author
+						};
```
StanleyGoldman: Mapping the graphql object to an object the `PullRequestModel` consumes. Not thrilled about doing this here.

src/github/graphql.ts
```diff
@@ -1,3 +1,5 @@
+import { IAccount } from "./interface";
+
```
rebornix: I guess it's from auto import and it misplaces the code ;)

src/github/graphql.ts
```diff
@@ -309,4 +311,66 @@ export interface RateLimit {
...
+	};
+	reviewRequests: {
+		nodes: [{
+			requestedReviewer: IAccount;
+		}];
+	};
+	createdAt: string;
+	updatedAt: string;
+	merged: boolean;
+	headRef: {
```
rebornix: Can we somehow reuse `interface Ref`

rebornix: Or probably a new `interface RefParts`

src/github/graphql.ts
StanleyGoldman: Thanks, fixed.

src/github/graphql.ts
StanleyGoldman: Yea, it's good enough to re-use `Ref` thanks.

src/github/githubRepository.ts
```diff
@@ -203,103 +203,44 @@ export class GitHubRepository implements IGitHubRepository, vscode.Disposable {
 		};
 	}
 
-	async getPullRequests(prType: PRType, page?: number): Promise<PullRequestData | undefined> {
-		return prType === PRType.All ? this.getAllPullRequests(page) : this.getPullRequestsForCategory(prType, page);
-	}
+	async getPullRequestsGraphQL(type: PRType, nextCursor?: string|null):Promise<PullRequestListResponse|undefined> {
+		const { remote, query, octokit } = await this.ensure();
+		const currentUser = octokit && (octokit as any).currentUser;
```
annthurium: just out of curiosity, why `octokit as any`?  Do we not have ts bindings for octokit or something?

src/github/githubRepository.ts
```diff
@@ -203,103 +203,44 @@ export class GitHubRepository implements IGitHubRepository, vscode.Disposable {
...
-							);
-							return null;
-						}
-
-						const item = convertRESTPullRequestToRawPullRequest(pullRequest);
-						return new PullRequestModel(this, this.remote, item);
-					}
-				)
-				.filter(item => item !== null) as PullRequestModel[];
+		let filter = `type:pr is:open repo:${remote.owner}/${remote.repositoryName}`;
```
annthurium: Might be a little edge casey, but do we need to guard against the remote being `null` here?

src/github/githubRepository.ts
```diff
@@ -203,103 +203,44 @@ export class GitHubRepository implements IGitHubRepository, vscode.Disposable {
...
+		if (type !== PRType.All) {
+			if (type === PRType.Mine) {
+				filter += ` author:${currentUserLogin}`;
+			} else if (type === PRType.RequestReview) {
+				filter += ` review-requested:${currentUserLogin}`;
+			} else if (type === PRType.AssignedToMe) {
+				filter += ` assignee:${currentUserLogin}`;
 			} else {
-				throw e;
+				throw new Error('Unexpected pull request filter');
```
annthurium: would it make sense to log the filter type here?

src/github/githubRepository.ts
StanleyGoldman: I have a similar question as I'm pretty sure the bindings exist.

RMacfarlane: Yeah, we do have bindings for octokit, but `currentUser` is something we've added ourselves

https://github.com/Microsoft/vscode-pull-request-github/pull/600#pullrequestreview-167227402

We also added an `octokit.ts` file a while ago that modifies the typings slightly, there was something that should have been nullable that wasn't. `currentUser` can be added there to get rid of our `any` casts. But I think we should ultimately try to get rid of that file by correcting the typings upstream

src/github/githubRepository.ts
```diff
@@ -203,103 +203,44 @@ export class GitHubRepository implements IGitHubRepository, vscode.Disposable {
...
 
-	async getPullRequests(prType: PRType, page?: number): Promise<PullRequestData | undefined> {
-		return prType === PRType.All ? this.getAllPullRequests(page) : this.getPullRequestsForCategory(prType, page);
-	}
+	async getPullRequestsGraphQL(type: PRType, nextCursor?: string|null):Promise<PullRequestListResponse|undefined> {
+		const { remote, query, octokit } = await this.ensure();
+		const currentUser = octokit && (octokit as any).currentUser;
+		const currentUserLogin: string = currentUser.login;
 
-	private async getAllPullRequests(page?: number): Promise<PullRequestData | undefined> {
```
RMacfarlane: I don't think we should remove this - we still need to support GHE that doesn't have GraphQL

src/github/pullRequestManager.ts
RMacfarlane: the file `src/github/utils.ts` has functions for mapping response types to normalized types

src/github/githubRepository.ts
StanleyGoldman: I would say yes, but looking at all the other calls to `ensure` to get `remote` seem to make the assumption that remote will not be null.

src/github/githubRepository.ts
StanleyGoldman: :+1:

src/github/pullRequestManager.ts
StanleyGoldman: Thanks, i'll move the code.

src/github/githubRepository.ts
```diff
@@ -203,103 +203,44 @@ export class GitHubRepository implements IGitHubRepository, vscode.Disposable {
...
-			Logger.appendLine(`Fetching all pull requests failed: ${e}`, GitHubRepository.ID);
-			if (e.code === 404) {
-				// not found
-				vscode.window.showWarningMessage(`Fetching pull requests for remote '${this.remote.remoteName}' failed, please check if the url ${this.remote.url} is valid.`);
+		if (type !== PRType.All) {
+			if (type === PRType.Mine) {
+				filter += ` author:${currentUserLogin}`;
+			} else if (type === PRType.RequestReview) {
+				filter += ` review-requested:${currentUserLogin}`;
+			} else if (type === PRType.AssignedToMe) {
```
StanleyGoldman: asdf

src/github/githubRepository.ts
```diff
@@ -203,103 +203,44 @@ export class GitHubRepository implements IGitHubRepository, vscode.Disposable {
...
-			};
-		} catch (e) {
-			Logger.appendLine(`GitHubRepository> Fetching all pull requests failed: ${e}`);
-			if (e.code === 404) {
-				// not found
-				vscode.window.showWarningMessage(`Fetching pull requests for remote ${this.remote.remoteName}, please check if the url ${this.remote.url} is valid.`);
-			} else {
-				throw e;
-			}
+		if(!!nextCursor) {
```
smbkr: Just curious why you use `if(!!condition)` here, instead of `if(condition)`

alexr00:
Thanks for the PR @StanleyGoldman. Since the goal of this PR was to enable https://github.com/microsoft/vscode-pull-request-github/issues/681, I'm going to close it. We already have https://github.com/microsoft/vscode-pull-request-github/issues/681 as of the last stable release.

Issue #1030, opened 2019-03-07
Support dynamic git provider registration


rebornix:
Right now in GH PR, we depend on the builtin Git extension to provide git repository information while when users join a Live Share session, we create a mock Git provider which communitates between Live Share host and guest sides and feed all the git repository information we need. It works as expected but doesn't scale as we can't make sure GH PR works in all possible non-local (workspace is not local) cases.

To support workspaces whose file system is virtual, for example Live Share, FTP, etc, we can expose an API for other extensions who already register the file system provider, to provide git repository information.

The API GH PR exposes looks like below

```typescript
export function registerGitProvider(provider: IGit);
export function resolveGitProvider(uri: vscode.Uri): IGit;

export interface IGit {
	onDidOpenRepository: vscode.Event<Repository>;
	onDidCloseRepository: vscode.Event<Repository>;
	
	openRepositories: Repository[];
}
```

GH PR then listens to git providers change and broadcast all events received from registered Git provider (IGit) to Pull Request Manager, which is reponsible for the business logic of GH PR.

![image](https://user-images.githubusercontent.com/876920/53924865-95f64d00-4032-11e9-952a-9140e1e16a46.png)

PR #1050, opened 2019-03-14
Allow extensions to register git providers

rebornix:
Re https://github.com/Microsoft/vscode-pull-request-github/issues/1030

Since we expose API for other extensions to register their own git provider, we don't have to embed LiveShare logic inside GH PR extension. For example, we can comment out from `extension.ts`.

```
registerLiveShareGitProvider()
```

and then install another extension https://github.com/rebornix/vscode-gh-liveshare , which glues together GH PR and LiveShare. The glue extension declares GH PR and LiveShare as dependencies.


RMacfarlane:
Hmm, I just tried to submit a review with a comment and got some kind of GraphQL error. I accidentally closed the notification when trying to expand it, so I'm not sure what it was.

Anyway, my summary comment was something like:
The code looks good overall, I'm a little worried about the discoverability about the "glue" extensions. As a user I wouldn't think to search for such a thing, it seems a bit odd.

There is this vscode issue about optional dependencies: https://github.com/Microsoft/vscode/issues/6384

If the GHPR extension was instead able to list LiveShare as an "optional dependency" so that it would get activated later if LiveShare was installed, then it would be possible to just check if LiveShare was enabled and register a git provider for it

src/common/utils.ts
```diff
@@ -196,3 +196,321 @@ export function dateFromNow(date: Date | string): string {
 export interface Predicate<T> {
 	(input: T): boolean;
 }
+
+export class PathIterator implements IKeyIterator {
```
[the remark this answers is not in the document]
rebornix: I should better document this, it's from vscode.

rebornix:
@RMacfarlane thanks for the review and feedback. I happen to have the same discussion with IntelliNav folks about the same topic.

IntelliNav now sets GH PR as an extension dependency so it always attempt to activate GH PR when loaded. It then conflicts with our Nightly build. The workaround we have right now is removing the dependency from IntelliNav and it detects if GH PR or GH PR Nightly is installed or activated or not.

`git provider` is the same story. If an extension wishes GH PR to work in their scenario, they can detect if GH PR exists/activates or not, if so then register proper git providers.

> If the GHPR extension was instead able to list LiveShare as an "optional dependency" so that it would get activated later if LiveShare was installed

what we need here is more like an OR operator for `extensionDependencies`. We want GH PR to be activated when there is Git or IntellINav or LiveShare.

src/gitProviders/builtinGit.ts
```diff
@@ -0,0 +1,34 @@
...
+
+	private _gitAPI: GitAPI;
+	private _disposables: vscode.Disposable[];
+
+	constructor() {
+		const gitExtension = vscode.extensions.getExtension<GitExtension>('vscode.git')!.exports;
+		this._gitAPI = gitExtension.getAPI(1);
+		this._disposables = [];
+		this._disposables.push(this._gitAPI.onDidCloseRepository(e => this._onDidCloseRepository.fire(e as any)));
+		this._disposables.push(this._gitAPI.onDidOpenRepository(e => this._onDidCloseRepository.fire(e as any)));
```
homitc: Is this meant to be this._onDidOpenRepository?

RMacfarlane: Yes, definitely. Thanks for catching this!

PR #1092, opened 2019-03-29
Create comment threads for files with review query.

### LINKED ISSUE #1089: Comments not updated on review scheme files when added on other file type
Issue Type: <b>Bug</b>

1. Check out a PR
2. Add a comment to one of its files with a "pr" or "file" scheme (i.e open a local file and add a comment, or open a file under the "GitHub Pull Requests" tree and add one)
3. Open the same file under the "Changes in Pull Request" tree

Comment is not shown

VS Code version: Code - Insiders 1.33.0-insider (f5044f0910e4aa7e7e06cb509781f3d56e729959, 2019-03-28T16:20:44.233Z)
OS version: Darwin x64 18.2.0

rebornix:
Fix #1089.

This bug was caused by a feature we added to the "Changes In Pull Request" view. When users browse this viewlet, they expect the right side editor to be editable. However, we can't simply use local file for the right side editor, because

* Workspace files display comments on both original content and modified content. Say there is a comment on the left diff editor and another comment on the right diff editor on github.com, we will display both of them in workspace file.
* For "Changes In Pull Request" file diff view, if the right side editor shows the normal workspace file, users will see duplicated comments

To solve this problem, we add `ReviewParams`  query to right side `file` scheme document. https://github.com/Microsoft/vscode-pull-request-github/blob/2ceb7f61089d06a2ae117175efeef493cb8c9d1d/src/view/reviewManager.ts#L388

Now with the new API, we create comment threads for workspace files (`file://Users/abc/vscode/packaeg.json`) when we detect the repository is in review mode. However no one takes care of `file` scheme documents with `ReviewParams` query.

### DIFF
--- a/src/view/reviewDocumentCommentProvider.ts
+++ b/src/view/reviewDocumentCommentProvider.ts
@@ -179,6 +179,12 @@ export class ReviewDocumentCommentProvider implements vscode.Disposable, Comment
 			for (let fileName in this._reviewDocumentCommentThreads) {
 				let threads = this._reviewDocumentCommentThreads[fileName];
 				let visible = vscode.window.visibleTextEditors.find(editor => {
+					if (editor.document.uri.scheme !== 'review' && editor.document.uri.scheme === this._repository.rootUri.scheme && editor.document.uri.query) {
+						if (fileName === editor.document.uri.toString()) {
+							return true;
+						}
+					}
+
 					if (editor.document.uri.scheme !== 'review') {
 						return false;
 					}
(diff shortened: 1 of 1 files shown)


rebornix:
@RMacfarlane it sounds right to me. To make this consistent, we should also do the same filtering for the diff editor for PR views. GH.com shows the comment on the left side (or more precisely across both editors) when it's on non-edited lines, in our case, we always show it on the right side (or the local file).

PR #1098, opened 2019-04-02
Use vscode.env.uriScheme for callback.

rebornix:
Instead of guessing the URI scheme, we now use `vscode.env.uriScheme` to construct the callback url.

We merge this only after we release the next version of VS Code, as this one requires 1.34


queerviolet:
Yay!

PR #1108, opened 2019-04-05
Update commenting api to latest

rebornix:
Update to the latest commenting api.

### DIFF
--- a/src/typings/vscode.proposed.d.ts
+++ b/src/typings/vscode.proposed.d.ts
@@ -16,6 +16,60 @@
 
 declare module 'vscode' {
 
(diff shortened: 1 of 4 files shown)


src/typings/vscode.proposed.d.ts
```diff
@@ -16,6 +16,60 @@
```
RMacfarlane: can you copy this file in one more time? looks like `env.uriScheme` has been promoted to stable so there are errors about it being redeclared here

RMacfarlane:
Closing in favor of https://github.com/microsoft/vscode-pull-request-github/pull/1168

PR #1110, opened 2019-04-08
Simplify loading keytar

rebornix:


src/authentication/keychain.ts
```diff
@@ -8,14 +8,8 @@ import { IHostConfiguration } from './configuration';
 function getNodeModule<T>(moduleName: string): T | undefined {
 	const vscodeRequire = eval('require');
 	try {
-		return vscodeRequire(`${vscode.env.appRoot}/node_modules.asar/${moduleName}`);
+		return vscodeRequire('keytar');
```
chrmarti: @alexandrudima Are you going to update the shimming, so there is no adoption needed?

alexandrudima: This change should not be necessary anymore... The shim should work in all cases now...

PR #1117, opened 2019-04-13
Make PR viewlet always visible

RMacfarlane:
With this change, the GitHub viewlet will always be visible. It will either show the tree as usual, or display a message for these situations
- no folder is open yet
- the open folder has no git repositories
- the open folder has no remotes associated with GitHub

I think we still need to do a better job of displaying information when `isGitHub` fails, and there may still be an underlying timing issue with something. But this should at least make the install experience less confusing if the user doesn't haven't an open folder at that time.

### DIFF
--- a/src/view/prsTreeDataProvider.ts
+++ b/src/view/prsTreeDataProvider.ts
@@ -9,7 +9,7 @@ import { PRCategoryActionNode, CategoryTreeNode, PRCategoryActionType } from './
 import { PRType, ITelemetry } from '../github/interface';
 import { fromFileChangeNodeUri } from '../common/uri';
 import { getInMemPRContentProvider } from './inMemPRContentProvider';
-import { PullRequestManager } from '../github/pullRequestManager';
+import { PullRequestManager, SETTINGS_NAMESPACE, REMOTES_SETTING } from '../github/pullRequestManager';
 
 export class PullRequestsTreeDataProvider implements vscode.TreeDataProvider<TreeNode>, vscode.DecorationProvider, vscode.Disposable {
 	private _onDidChangeTreeData = new vscode.EventEmitter<TreeNode>();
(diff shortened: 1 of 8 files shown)


src/view/prsTreeDataProvider.ts
```diff
@@ -63,6 +70,18 @@ export class PullRequestsTreeDataProvider implements vscode.TreeDataProvider<Tre
 	}
 
 	async getChildren(element?: TreeNode): Promise<TreeNode[]> {
+		if (!this._prManager) {
```
rebornix: Probably not related to this change but when I set the `github.remotes` to `nonsense` and then revert it back, the prManager is still there. But it's a very low pri issue.

src/view/treeNodes/categoryNode.ts
```diff
@@ -16,7 +16,11 @@ export enum PRCategoryActionType {
 	Empty,
 	More,
 	TryOtherRemotes,
-	Login
+	Login,
+	NoRemotes,
+	NoGitRepositories,
+	NoOpenFolder,
+	ConfigureRemotes
```
rebornix: Seems it's not used anywhere?

rebornix:
It looks good in overral. The only catch is how the git state is updated and triggering events. Git extension attempts to batch/throttle events so when I open a  github repository and have GH PR viewlet open, I saw below information in sequence

* No data provider registered
* No GitHub repository found
* Categories

When I change the `githubPullRequests.remotes` to `upstream` and open a repository which doesn't have upstream remote, I still see `No GitHub repository` found message, which doesn't reflect what's happening.

RMacfarlane:
I updated the message for the case where the `remotes` setting doesn't match anything, there's now also a tree item to open settings for it.

I looked into the Git extension's eventing, it looks like the issue is that it first creates a model with some repositories, and then updates the data on the repositories to actually have remotes with urls. I think it could be changed to just await the first `status` call it does so that the repository information we get initially is complete

Issue #1029, opened 2019-03-06
Add more metrics for user type and pull request actions


meaghanlewis:
I'd like to capture more data about our types of users and what they are doing in the Pull Requests extension.

**User data**
- How many users are GitHub users compared to GitHub Enterprise users?
  - Is it possible to know what version of Enterprise customers are on?

**Pull Request data**
- How many times are pull requests are merged? closed?
- How many times is the local pull request branch deleted?
- How many times an inline/review comment is added?

**Interaction with activity bar**
One other thing that would be interesting to know is how many times users click the GitHub icon on the activity bar.

Any others that you'd like to see?

meaghanlewis:
See the metrics previously added here: https://github.com/Microsoft/vscode-pull-request-github/issues/228#issuecomment-418356099

meaghanlewis:
@annthurium I think another good metric to add would be for how many times a PR is created.

RMacfarlane:
Has been completed with https://github.com/Microsoft/vscode-pull-request-github/pull/1118

PR #1118, opened 2019-04-15
Add more telemetry

annthurium:
addresses https://github.com/Microsoft/vscode-pull-request-github/issues/1029

### Metrics added

- [x] How many users are GitHub users compared to GitHub Enterprise users?  Is it possible to know what version of Enterprise customers are on?

I did not see an easy way to see which users are enterprise users from the extension (although perhaps it might be easier from the auth server).  Super open to suggestions about how to approach this, though!

Do we really care about enterprise versions?  Or do we care about graphQL support?  I figured it was the latter, and added some metrics at the time we check for graphQL support.  One concern that I have is that it looks like this code is called every time we make a request.  It seems like some caching would be useful here.  Anyhow, on the dashboard side we should make sure we are de duping this query by UUID or something, so we get the number of users with supported graphQL versions and not just the number of requests.

- [x] How many times are pull requests are merged? closed?
Closed pull requests are already being logged and on the dashboard, but I added metrics for merging success and failure.

- [x] How many times is the local pull request branch deleted?
Added success and failure metrics.

- [x] How many times an inline/review comment is added?
Added metrics for replying to a comment, starting a review, finishing a review, deleting a review, editing a comment, deleting a comment, and deleting a thread.  

- [ ] how many times users click the GitHub icon on the activity bar.
It looks like this icon is defined by configuration in `package.json` but it looks like the click handling is handled outside of the code for the extension.  Is that correct?  Or am I missing something? 

### Verification plan
- [x]  used `Logger.appendLine` and ran through each of these activities.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -367,6 +367,7 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: Pu
 	}));
 
 	context.subscriptions.push(vscode.commands.registerCommand('pr.deleteThread', async (thread: vscode.CommentThread) => {
+		telemetry.on('pr.deleteThread');
 		thread.dispose!();
 	}));
 }
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -177,8 +177,10 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: Pu
 		}
 
 		if (error) {
+			telemetry.on('pr.deleteLocalPullRequest.failure');
 			await vscode.window.showErrorMessage(`Deleting local pull request branch failed: ${error}`);
 		} else {
+			telemetry.on('pr.deleteLocalPullRequest.success');
 			// fire and forget
 			vscode.commands.executeCommand('pr.refreshList');
 		}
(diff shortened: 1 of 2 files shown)


RMacfarlane:
There are a couple of places we check if the host authority is `github.com` or something else, which I assume distinguishes non-enterprise and enterprise users:

https://github.com/Microsoft/vscode-pull-request-github/blob/4e9288ed6a607b907d42f006a2e3ceb3f385cd04/src/authentication/configuration.ts#L18-L25

I'm not sure how important enterprise versions are, so far the extension has been operating under the assumption that if there's any graphQL support, all of the queries we make are supported. The place where we check for graphQL support should only be hit once on startup, so I think that's good.

For the number of times users click the GitHub icon, you're right, the click handling is done by the core of VSCode. Logging this from the extension seems troublesome. The tree's `getChildren` function is called when the tree is first rendered when switching to the viewlet, but also for any refresh the extension does. I'll check to see what we log in the core

annthurium:
I just heard from @meaghanlewis who said "I don’t think we really care about the specific version" so I'm going to merge this as is.  Thanks @RMacfarlane and @queerviolet !

Issue #574, opened 2018-10-12
GitHub Enterprise: TypeError: Cannot read property 'map' of undefined


Justin-Credible:
- Extension version: 0.2.0
- VSCode Version: 1.28.1 (3368db6750222d319c851f6d90eb619d886e08f5)
- OS: macOS 10.12.6

Steps to Reproduce:

1. Add `github.hosts` entry to workspace config
2. Open Source Control panel in VSCode
3. Expand GitHub Pull Requests pane
4. Expand any items in tree

Result: `Fetching all pull requests failed: TypeError: Cannot read property 'map' of undefined`

Output from extension shows:
```
Looking for git repository
Git repository found, initializing review manager and pr tree view.
No GitHub remotes found
Found GitHub remote
Review> no matching pull request metadata found for current branch feature/xxxxxxxxxxxxxxxx
GitHubRepository> Fetching all pull requests failed: TypeError: Cannot read property 'map' of undefined
GitHubRepository> Fetching all pull requests failed: TypeError: Cannot read property 'map' of undefined
```

Workspace settings:
```
    "github.hosts": [
        {
            "host": "https://git.company-name.com",
            "username": "justin-unterreiner",
            "token": "xxxxxxxxxxxxxxxxxxxx"
        }
    ]
```

Additional information:
* Our GitHub instance _does not_ use a self-signed certificate, it is a valid certificate signed by GoDaddy.
* I am able to use the GitLense VSCode extension with our enterprise GitHub instance by setting `gitlens.remotes` in the workspace settings

rebornix:
@queerviolet FYI the `data` property on return result is undefined 

https://github.com/Microsoft/vscode-pull-request-github/blob/197c1d6ad82d581725d041c8a3b005f6362f5e14/src/github/githubRepository.ts#L98-L106

EslamElHusseiny:
+1

rconnamacher:
I also have this issue

thetaylormcrae:
Glad I'm not the only one. GitHub Enterprise with VSCode seems to be a general pain. :-/

grork:
Any update or progress on this issue?

abathingchris:
This is still an issue, any update or workaround?

StanleyGoldman:
So I'm catching up with old issues and inspecting this one. I wanted to call out something in the original issue report that threw me for a bit.

[...]

PR #1127, opened 2019-04-19
Guard PR data

queerviolet:
This addresses the error seen by users in #574. I don't know that it'll address the underlying problem, which is that `result` should always return data, but perhaps it'll stop masking it.

### DIFF
--- a/src/github/githubRepository.ts
+++ b/src/github/githubRepository.ts
@@ -221,6 +221,13 @@ export class GitHubRepository implements IGitHubRepository, vscode.Disposable {
 			});
 
 			const hasMorePages = !!result.headers.link && result.headers.link.indexOf('rel="next"') > -1;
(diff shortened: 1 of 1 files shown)


src/github/githubRepository.ts
```diff
@@ -221,6 +221,13 @@ export class GitHubRepository implements IGitHubRepository, vscode.Disposable {
 			});
 
 			const hasMorePages = !!result.headers.link && result.headers.link.indexOf('rel="next"') > -1;
+			if (!result.data) {
```
StanleyGoldman: It feels like a nit pick, but since this is a scenario we can't identify, i think we should log a warning and leave a comment in the code stating why, link to the issue or something.

queerviolet: Done, PTAL.

Issue #1091, opened 2019-03-29
Extension seems needlessly slow on slow internet connections


queerviolet:
- Extension version: Insiders
- VSCode Version: Insiders
- OS: all

Steps to Reproduce:

1. Go somewhere with terrible internet
    *or*
    Throttle your internet
2. Enter review mode for a branch you already have locally.
3. Wait.
4. Wait.
5. Wait.
6. The checkout is complete.

I believe 3, 4, and 5 are unnecessary.

I think we're hitting the API for some reason, we should cache whatever we need so we don't have to.

rebornix:
Set the log level to Debug and you can see timestamps from the output channel. @queerviolet can you share them here? The output can tell us which operations are time consuming.

alexr00:
Closing since we don't have the debug log.

PR #1131, opened 2019-04-25
Decrease time to enter review mode.

queerviolet:
Low-hanging fruit for #1091.

Previously, we awaited the fetch of additional PR data, then checked out
the branch. Now, we attempt to checkout the branch, and then we fetch
additional data.

Also:
  - adds more specific status bar messages
  - immediately update the status bar message once the branch is checked out, indicating to
the user that they can go about their day while we fetch additional
information
  - unconditionally resets the status bar, so it doesn't appear to hang in a fetching state

Tested manually, on fast / slow / nonexistent connections, with branches
that both do and do not exist locally (the latter fail, as we would
expect).

### DIFF
--- a/src/view/reviewManager.ts
+++ b/src/view/reviewManager.ts
@@ -492,12 +492,10 @@ export class ReviewManager implements vscode.DecorationProvider {
 
 	public async switch(pr: PullRequestModel): Promise<void> {
 		Logger.appendLine(`Review> switch to Pull Request #${pr.prNumber} - start`);
-		this.switchingToReviewMode = true;
-		await this._prManager.fullfillPullRequestMissingInfo(pr);
-
 		this.statusBarItem.text = '$(sync~spin) Switching to Review Mode';
 		this.statusBarItem.command = undefined;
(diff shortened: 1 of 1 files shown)


src/view/reviewManager.ts
```diff
@@ -522,10 +520,22 @@ export class ReviewManager implements vscode.DecorationProvider {
 			return;
 		}
 
-		this._telemetry.on('pr.checkout');
-		Logger.appendLine(`Review> switch to Pull Request #${pr.prNumber} - done`, ReviewManager.ID);
-		this.switchingToReviewMode = false;
-		await this._repository.status();
+		try {
+			this.statusBarItem.text = `$(sync~spin) Fetching additional data: pr/${pr.prNumber}`;
```
rebornix: This one can be skipped if the pull request has all information needed?

queerviolet: It didn't seem obvious and quick to see if it has all the information needed. Maybe we can do another easy-win PR to early-out in `fullfillPullRequestMissingInfo` if it's all there?

Issue #1135, opened 2019-04-26
Checkout an existing branch


borekb:
I already have a branch of a pull request checked out locally, by previously running a [`hub`](https://hub.github.com/) command:

```
hub pr checkout 123
```

This created `feature-branch` in my local repository.

Now when I go to VSCode, it doesn't realize that this branch is related to a PR – fair enough, it doesn't have enough information about it.

However, I'm missing a way to tell the extension that my `feature-branch` represents PR 123. If I check out the branch via the GitHub Pull Requests sidebar, it creates a new branch. So I end up with this:

```
  master             4475b80e Merge pull request #122 from repo/branch
  feature-branch     e1b1886c Remove unnecessary strtolower
* pr/someuser/123    e1b1886c Remove unnecessary strtolower
```

I would like to end up with this instead:

```
  master             4475b80e Merge pull request #122 from repo/branch
* feature-branch     e1b1886c Remove unnecessary strtolower
```

borekb:
Is `pr/someuser/<number>` a pattern that the extension needs to see to be able to recognize branches representing pull requests? That would make the implementation of this feature request tricky...

RMacfarlane:
Thanks for reporting this, this sounds like a bug. The extension should try to do what you describe already.

The `pr/someuser/<number>` pattern should just be the default when it doesn't detect an existing branch, we do however write a different key value pair (`github-pr-owner-number = "<repo>#<pr number>"`) to the .git/config file

borekb:
Does it make a difference that the branch was coming from a fork, i.e., non-`origin` remote?

RMacfarlane:
Yep, you're exactly right. We have a bug with how we're handling those remotes. Thanks!

PR #1141, opened 2019-05-01
On checkout, make sure existing remote is found so that existing branch can be checked out

RMacfarlane:
Fixes https://github.com/Microsoft/vscode-pull-request-github/issues/1135

Our checkout process forks a couple of times - first we check to see if we have already added a key-value pair to the .git/config indicating that a branch is associated with a PR.

If that fails, we then check if there is already a remote for the branch we are trying to switch to. If there's no remote, we create one and create the branch with the pr/user/number pattern. The bug here was with matching the remote - we were looking at remotes that are "active", those we are actually using to display PRs from, and not all of the remotes in the repo. So even if a user already had both a remote from a fork and the local branch, we would fail to match the remote and then create a pr/user/number branch.

### DIFF
--- a/src/github/pullRequestGitHelper.ts
+++ b/src/github/pullRequestGitHelper.ts
@@ -10,7 +10,6 @@
 import Logger from '../common/logger';
 import { Protocol } from '../common/protocol';
 import { Remote, parseRepositoryRemotes } from '../common/remote';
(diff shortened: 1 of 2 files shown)


queerviolet:
LGTM!

PR #1159, opened 2019-05-15
Fix review comments

queerviolet:
@kuychaco and I discovered and fixed some issues with review comments on the overview page.

In particular, we now:
  * send the correct events in response to the end-of-overview-page buttons (we weren't)
  * show review the body of reviews (we weren't)
  * customize the "reviewed" string based on the review type (we weren't; I think we weren't even before the React PR, but anyway, we are now)

We unearthed all this while making an ergonomic tweak. Namely, we now accept ⌘Enter and C-Enter to submit a comment.

### DIFF
--- a/preview-src/timeline.tsx
+++ b/preview-src/timeline.tsx
@@ -8,9 +8,10 @@ import { Avatar, AuthorLink } from './user';
 import { groupBy } from '../src/common/utils';
 import { Spaced, nbsp } from './space';
 import Timestamp from './timestamp';
-import { CommentView } from './comment';
+import { CommentView, CommentBody } from './comment';
 import Diff from './diff';
 import PullRequestContext from './context';
+import { ReviewState } from '../src/github/interface';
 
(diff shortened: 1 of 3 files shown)


preview-src/timeline.tsx
```diff
@@ -76,11 +89,16 @@ const ReviewEventView = (event: ReviewEvent) => {
...
 						? <em>review pending</em>
 						: <>
-								reviewed{nbsp}
+								{reviewDescriptor(event.state)}{nbsp}
 								<Timestamp href={event.htmlUrl} date={event.submittedAt} />
 						</> }
 				</Spaced>
 			</div>
+			{
+				event.state !== 'PENDING'
```
RMacfarlane: can we also check that body exists? on .com nothing is shown if there's no content

queerviolet:
Fixed, PTAL

RMacfarlane:
LGTM!

PR #1162, opened 2019-05-16
Update the title when the PR changes.

queerviolet:
It wasn't updating because it was being held in component state. It was initialized to the value of a prop, but didn't track it.

This happened elsewhere earlier, so we went ahead and factored out the behavior as a new hook: `useStateProp`.

Co-Authored-By: Katrina Uychaco <kuychaco@github.com>


RMacfarlane:
Nice catch

PR #1163, opened 2019-05-16
Fix a test failure on master.

queerviolet:
It looks like the behavior of vscode.Uri has changed at some point to recognize file paths like '/opt/git/project.git' as `file:` scheme URLs.

Our tests were very surprised by this.

Co-Authored-By: kuychaco@github.com


kuychaco:
Did some code archaeology and found this for reference -- https://github.com/microsoft/vscode/commit/d2239c2473fda8ae7898f75a453f8a9eb0fdad1d#diff-05910b80b38cefe548a9d4f9c9d6c78eR60

Issue #1106, opened 2019-04-05
Allow custom pull request filters


szszoke:
On GitHub, we have the ability to filter pull requests, and reuse the same filter by storing the URL in a bookmark.
![image](https://user-images.githubusercontent.com/2766129/55618695-6fdddd00-5797-11e9-9877-3c14dfe99775.png)

Allow adding our own categories here :point_down: with custom filters using the pre-existing GitHub search qualifiers.
![image](https://user-images.githubusercontent.com/2766129/55618784-a0be1200-5797-11e9-99fd-db360a3aa65b.png)

[More info on the search qualifiers](https://help.github.com/en/articles/searching-issues-and-pull-requests)

RMacfarlane:
I love this idea! I made a prototype of how this could work:
![custom_categories](https://user-images.githubusercontent.com/3672607/58119981-17349900-7bb9-11e9-9e94-d6019e6c3c7c.gif)

On GitHub.com, it's nice to be able to use the dropdowns to help construct the query. I'm not sure what we could do here to help with query writing, validation seems challenging since my understanding of the query language is that anything not interpreted as a search qualifier would just be used as a keyword.

Other things we could have:
- "Edit", which allows you to pick a category, first modify the title, then modify the query
- Be able to view the query in the tree?
- Have an inline action on a category to either view or modify the query

PR #1167, opened 2019-05-21
Support specifying custom category in PR tree view

RMacfarlane:
https://github.com/microsoft/vscode-pull-request-github/issues/1106

Experimenting with being able to set custom tree categories using normal GitHub search syntax
queries_setting

This would also open up viewing closed pull requests, need to test that there are no problems with displaying these


kuychaco:
Oh ya know, I just realized that there might be some UX friction to smooth over before merging this. If the user has the "Use split JSON" option enabled (as shown in the gif above) then it's clear what information users will want to add to their settings.json file. The default is for this setting to be off, in which case it's not quite clear that the user will need to add an array of objects with label and query fields. 

Perhaps we can address this by adding more information in the setting description.

RMacfarlane:
@kuychaco I updated the description of the setting to try to clarify this, let me know if it makes sense to you.

It would be good to have telemetry on this, with that info we could even tweak the default categories. But I'm not sure if this would be OK to log, since it's direct user input.

package.json
```diff
@@ -87,6 +87,38 @@
...
+                "description": "The label to display for the query in the Pull Requests tree"
+              },
+              "query": {
+                "type": "string",
+                "description": "The query used for searching pull requests."
+              }
+            }
+          },
+          "scope": "resource",
+          "markdownDescription": "Specifies what queries should be used in the GitHub Pull Requests tree. Each query has a label that will be shown in the tree and a search query using [GitHub search syntax](https://help.github.com/en/articles/understanding-the-search-syntax). The variable `${user}` can be used to specify the logged in user within a search. By default these queries define the categories \"Waiting For My Review\", \"Assigned To Me\" and \"Created By Me\". If you want to preserve these, make sure they are still in the array when you modify the setting.",
```
kuychaco:
```suggestion
          "markdownDescription": "Specifies what queries should be used in the GitHub Pull Requests tree. Each query object has a `label` that will be shown in the tree and a search `query` using [GitHub search syntax](https://help.github.com/en/articles/understanding-the-search-syntax). The variable `${user}` can be used to specify the logged in user within a search. By default these queries define the categories \"Waiting For My Review\", \"Assigned To Me\" and \"Created By Me\". If you want to preserve these, make sure they are still in the array when you modify the setting.",
```

RMacfarlane:
I'll create a new issue and defer the telemetry for now, so we can double check GDPR. Also, looking at the typings for the telemetry module, right now just logs an action, so it might also need more work to send additional data.

PR #1169, opened 2019-05-22
Expand test coverage

smashwilson:
Let's expand the scope of our test suite to cover more of the extension's functionality. I'm starting here to provide us with some immediate benefits:

* It'll enable myself, @kuychaco, and @vanessayuenn to make changes with a lower risk of causing regressions. We're new to the codebase and don't yet have a mature mental model of how the source is structured and interconnected, so we're likely to break unrelated functionality otherwise.
* I'll have a chance to learn how things work and tour the code myself in a way that's more active than just reading it.
* We'll be better equipped to evaluate and accept community contributions with confidence.
* It'll be easier to update dependencies and enable [dependabot](https://dependabot.com/).
* We'll become less reliant on manual [testing checklist issues](https://github.com/microsoft/vscode-pull-request-github/labels/testplan-item).

Backfitting tests to existing code is tedious and time-consuming. My goal here isn't to get us to full coverage - mostly, I'm concerned with introducing a skeleton structure that lowers the barriers to adding tests organically as we work in the future. I'll call this PR "complete" once I have a test suite in place for a set of basis cases that cover the mocking needs of different parts of the codebase.

I'm also doing my best to minimize structural changes to the source. I'll likely need to change _some_ things to make things more easily testable, but I'll endeavor to keep them small and scopes to keep this effort tractable and minimize risk.

#### Remaining

Modules that use the VSCode API. The primary concern here is being able to avoid collisions with globally-scoped resources like command IDs.

- [x] src/commands.ts
- [x] src/github/credentials.ts
- [x] src/github/pullRequestManager.ts

Modules that expect to use the network. Testing these effectively requires a way to succinctly mock GraphQL and REST responses.

- [x] src/github/githubRepository.ts
- [x] src/github/pullRequestOverview.ts

Modules that use git operations. These rely on a way to mock git commands performed through the git extension's exported API.

- [x] src/github/pullRequestGitHelper.ts

Source that defines React components loaded within a Webview.

- [x] preview-src/app.tsx

Integration tests that exercise the UI components contributed to VSCode, like the tree view and the status bar tiles.

- [x] src/view/prsTreeDataProvider.ts

Finally - general support infrastructure work:

- [x] Support `*.gql` imports in Mocha tests.
- [x] Document the Builder setup.
- [x] Deal with the extension under test attempting to register commands in the `MockCommandRegistry` during initialization


rebornix:
@smashwilson this is a really great effort! 

When we introduced graphql into the project, we separate models and views by putting everything view related into `src/view` and others into their own components `common`, `github`.

For testing `src/view`, the only thing we need to mock (if we didn't leak other layers into it recently) is `PullRequestModel` and `PullRequestManager`. IMO we have unit tests for `PullRequestModel` and `PullRequestManager` and for `src/view`, we do integration tests directly by VS Code test runner.

smashwilson:
I've updated the description and checklist here with a rough outline of (somewhat more tractable?) testing work that I'd like to get in place before merge. If if drags on I'd also be open to merging with only a few of the test suite categories done and addressing the remaining ones in other pull requests.

src/github/pullRequestManager.ts
```diff
@@ -109,16 +110,19 @@ export class PullRequestManager {
 	constructor(
 		private _repository: Repository,
 		private readonly _telemetry: ITelemetry,
+		private _credentialStore: CredentialStore = new CredentialStore(_telemetry),
```
smashwilson: There are some tests where I was able to allow a `PullRequestManager` to create its own `CredentialStore` and access it later, and others where I had to create it ahead of time and pass it in. It's likely possible to do some later refactoring to unify them.

By the way, the complication here is that `CredentialStore` registers a command in its constructor, so you can't instantiate more than one instance at a time.

src/github/pullRequestManager.ts
```diff
@@ -1534,6 +1542,14 @@ export class PullRequestManager {
 
 		return events;
 	}
+
+	createGitHubRepository(remote: Remote, credentialStore: CredentialStore): GitHubRepository {
+		return new GitHubRepository(remote, credentialStore);
+	}
```
smashwilson: I believe sinon has the ability to mock a constructor, but it always feels too invasive for my tastes. (What if there's another instantiation somewhere on the stack that I _don't_ want to mock?)

src/github/pullRequestOverview.ts
```diff
@@ -259,6 +259,7 @@ export class PullRequestOverviewPanel {
 				}
 			});
 		}).catch(e => {
+			console.error(e.stack);
```
smashwilson: Oops. I'll delete this one.

src/test/builders.test.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+const ParentBuilder = createBuilderClass<IParent>()({
+	aStringProp: {default: 'abc'},
+	aNumberProp: {default: 123},
+	aBooleanProp: {default: true},
+	aChildProp: {linked: ChildBuilder},
+});
+
+describe('Builders', function () {
```
smashwilson: Fun fact: almost all of the actual complexity of the Builder tests is in getting them to typecheck. :stars:

src/test/index.ts
```diff
@@ -1,11 +1,62 @@
...
+
+	mockWebviewEnvironment.install(global);
+
+	const mocha = new Mocha({
+		ui: 'bdd',
+		useColors: true,
+	});
+
+	await addTests(mocha, testsRoot);
+	await addTests(mocha, path.resolve(testsRoot, '../../preview-src/dist/preview-src/test'));
```
smashwilson: This needs to be a custom runner primarily so that we can add tests from `preview-src/`. It's also a convenient place to install the `MockWebviewEnvironment`.

test.sh
```diff
@@ -1,4 +1,17 @@
 #!/bin/bash
+
+if [ ! -f media/extension.js ]; then
+	printf "Running webpack\n"
+	webpack --env.development
+fi
```
smashwilson: :point_up: This bit will be useful when we run in CI.

Fun fact: the tests are loaded from the `tsc`-compiled JavaScript emitted by this script, but the development extension that's loaded into the extension test host uses `media/extension.ts`, so they both need to exist.

smashwilson:
Okay, merged in the changes from #1173 and got everything compiling and green again. Ready for review for reals this time 😄

src/github/credentials.ts
```diff
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as Octokit from '@octokit/rest';
+import Octokit = require('@octokit/rest');
```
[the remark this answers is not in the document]
RMacfarlane: For octokit, the module ships with a .d.ts typings file, so I think the normal import syntax is fine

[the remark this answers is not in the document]
smashwilson: If I'm understanding correctly, the issue isn't the lack of a typings file, but rather the export syntax used within it:

```ts
declare namespace Octokit {
  // ...
}


declare class Octokit {
  // ...
}

export = Octokit
```

> The export = syntax specifies a single object that is exported from the module. This can be a class, interface, namespace, function, or enum.
>
> When exporting a module using `export =`, TypeScript-specific `import module = require("module")` must be used to import the module.

From a little investigation, it seems like importing it with `import * as Octokit from '@octokit/rest'` will import only the namespace definitions but not the class. Granted, I doubt we're using the Octokit class, so I don't think it matters much. `assert` is a similar case; if you `import * as assert from 'assert'` then you don't get the `assert(true)` function form.

Is any of this inaccurate? I'm slowly learning my way around TypeScript imports, advice is appreciated 😄

PR #1168, opened 2019-05-21
Adopt new commenting api

RMacfarlane:
### History

A longer explanation is here: https://github.com/microsoft/vscode/issues/68020#issue-407501199

The first iteration of the comments API was provider-based, which is a common pattern used by many VS Code APIs. In a **provider** based model, the extension is responsible for providing the data, and the rendering of the data is controlled by VS Code. The comments provider was essentially a class the extension implemented that had methods for returning comments for a specific document, adding, editing, and deleting comments, an event to fire for updates to comments, and methods for handling reviews.

The pros of this API were that it created consistent commenting experiences across extensions since all of the rendering was done by VS Code, and it was very explicit about how it should be used. The drawback to this API was that it was difficult to extend in a generic way. When we added the ability to create an entire review instead of leaving individual comments, we had to add explicit `startDraft`, `deleteDraft`, and `finishDraft` methods to the comment provider. The major driver for changing the comment API was to allow extensions more control over what actions a comment provider could take.

### The new API

In full, the new API looks like:
```ts
//#region Comments

	/**
	 * Collapsible state of a [comment thread](#CommentThread)
	 */
	export enum CommentThreadCollapsibleState {
		/**
		 * Determines an item is collapsed
		 */
		Collapsed = 0,

		/**
		 * Determines an item is expanded
		 */
		Expanded = 1
	}

	/**
	 * Comment mode of a [comment](#Comment)
	 */
	export enum CommentMode {
		/**
		 * Displays the comment editor
		 */
		Editing = 0,

		/**
		 * Displays the preview of the comment
		 */
		Preview = 1
	}

	/**
	 * A collection of [comments](#Comment) representing a conversation at a particular range in a document.
	 */
	export interface CommentThread {
		/**
		 * The uri of the document the thread has been created on.
		 */
		readonly uri: Uri;

		/**
		 * The range the comment thread is located within the document. The thread icon will be shown
		 * at the first line of the range.
		 */
		range: Range;

		/**
		 * The ordered comments of the thread.
		 */
		comments: ReadonlyArray<Comment>;

		/**
		 * Whether the thread should be collapsed or expanded when opening the document.
		 * Defaults to Collapsed.
		 */
		collapsibleState: CommentThreadCollapsibleState;

		/**
		 * Context value of the comment thread. This can be used to contribute thread specific actions.
		 * For example, a comment thread is given a context value as `editable`. When contributing actions to `comments/commentThread/title`
[...]

PR #1201, opened 2019-06-04
do not rerender the html

rebornix:
with 1.35, we can have the webview persistent now without restoring it when visibility changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "version": "0.7.0",
   "publisher": "GitHub",
   "engines": {
-    "vscode": "^1.34.0"
+    "vscode": "^1.35.0"
   },
   "categories": [
     "Other"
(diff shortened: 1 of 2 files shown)


queerviolet:
Nice! Though this does break my trick of naving away from the panel and back to it to reload changes in the webview ;)

Still an improvement.

rebornix:
@queerviolet I'll double confirm that if Webviews serialization is still necessary here but right now I think it can be removed as we don't support workspace reload.

RMacfarlane:
@rebornix This broke the build again since it bumped the engine version to 1.35. Should we temporarily revert it until https://github.com/microsoft/vscode-pull-request-github/pull/1168 is merged?

rebornix:
Sorry for that, yes let's revert it.

Issue #1175, opened 2019-05-23
Description page is blank after 0.7.0 upgrade


erihanse:
Issue Type: <b>Bug</b>

When I click "Description" on one of my pull requests, it takes a while for it to appear. Another one doesn't appear at all.

![image](https://user-images.githubusercontent.com/9251274/58227614-dc9f3d80-7d2b-11e9-9124-a7b00a8358c8.png)

Extension version: 0.7.0
VS Code version: Code 1.34.0 (a622c65b2c713c890fcf4fbf07cf34049d5fe758, 2019-05-15T21:55:35.507Z)
OS version: Linux x64 4.15.0-50-generic

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz (8 x 2700)|
|GPU Status|2d_canvas: enabled<br>checker_imaging: disabled_off<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: disabled_software<br>surface_synchronization: enabled_on<br>video_decode: unavailable_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|1, 1, 1|
|Memory (System)|31.12GB (18.20GB free)|
|Process Argv|--unity-launch|
|Screen Reader|no|
|VM|0%|
</details>

RMacfarlane:
@erihanse For the one that doesn't appear, can you press `F1` and then run `Developer: Toggle Developer Tools` and see if there is anything in the console?

omawhite:
hey @RMacfarlane seeing this same issue but on a different os, I haven't been able to get any of the pull request descriptions to load.

Extension version: 0.7.0
OS Version: MacOS 10.14.3
VS Code version: Version 1.34.0 (1.34.0)

heres what i see in the console output:
![Screen Shot 2019-05-30 at 3 13 37 PM](https://user-images.githubusercontent.com/10691377/58661452-9c245e80-82ed-11e9-9702-13de4f301ea8.png)

hope this helps some.

RMacfarlane:
Thanks @omawhite, that does help!

kaufmann42:
I've been getting the same error.

PR #1207, opened 2019-06-04
Make sure author association is set on review events

RMacfarlane:
Fixes https://github.com/microsoft/vscode-pull-request-github/issues/1175

We do authorAssocation.toLowerCase() in the WebView, but were not correctly setting authorAssocation on review events when using the REST API. There's still an "any" that's destroying type safety in the function converting the REST response to normalized form, we need to update our octokit typings. I removed some casts in our GraphQL conversion function that were preventing type checking.

### DIFF
--- a/src/github/utils.ts
+++ b/src/github/utils.ts
@@ -367,14 +367,14 @@ export function parseGraphQLReviewEvent(review: GraphQL.SubmittedReview, githubR
 }
 
 export function parseGraphQLTimelineEvents(events: (GraphQL.MergedEvent | GraphQL.Review | GraphQL.IssueComment | GraphQL.Commit | GraphQL.AssignedEvent)[], githubRepository: GitHubRepository): Common.TimelineEvent[] {
-	let ret: Common.TimelineEvent[] = [];
+	const normalizedEvents: Common.TimelineEvent[] = [];
 	events.forEach(event => {
 		let type = convertGraphQLEventType(event.__typename);
 
 		switch (type) {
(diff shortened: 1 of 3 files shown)


src/github/utils.ts
```diff
@@ -367,14 +367,14 @@ export function parseGraphQLReviewEvent(review: GraphQL.SubmittedReview, githubR
 }
 
 export function parseGraphQLTimelineEvents(events: (GraphQL.MergedEvent | GraphQL.Review | GraphQL.IssueComment | GraphQL.Commit | GraphQL.AssignedEvent)[], githubRepository: GitHubRepository): Common.TimelineEvent[] {
-	let ret: Common.TimelineEvent[] = [];
+	const normalizedEvents: Common.TimelineEvent[] = [];
```
kuychaco: 💯 for descriptive variable names

src/github/utils.ts
```diff
@@ -399,22 +399,24 @@ export function parseGraphQLTimelineEvents(events: (GraphQL.MergedEvent | GraphQ
...
 					state: reviewEvent.state,
 					id: reviewEvent.databaseId,
-				} as Common.ReviewEvent);
+				});
 				return;
 			case Common.EventType.Committed:
 				let commitEv = event as GraphQL.Commit;
-				ret.push({
+				normalizedEvents.push({
+					id: commitEv.databaseId,
```
kuychaco: Great catches here!

PR #1209, opened 2019-06-04
Change "wants to merge" text when viewing merged PR

### LINKED ISSUE #1027: Merged pull request should use past tense
- Extension version:
- VSCode Version:
- OS: MacOS

Steps to Reproduce:

1. Merge an open pull request
2. View pull request header

screen shot 2019-03-04 at 9 35 03 pm

Compared with GitHub.com:
screen shot 2019-03-05 at 8 26 45 am

Expected result:
meaghanlewis **merged** changes (instead of meaghanlewis **wants to merge**)

smashwilson:
When viewing an unmerged pull request, the header looks like so:

before-merge

When viewing a merged pull request, it now reads as:

after-merged

Fixes #1027.

### DIFF
--- a/preview-src/header.tsx
+++ b/preview-src/header.tsx
@@ -18,7 +18,8 @@ export function Header({ canEdit, state, head, base, title, number, url, created
 			<Avatar for={author} />
 			<span className='author'>
 				<Spaced>
-					<AuthorLink for={author} /> wants to merge changes
+					<AuthorLink for={author} />
+					{getActionText(state)}
 					into <code>{base}</code>
 					from <code>{head}</code>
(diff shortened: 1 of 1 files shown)


preview-src/header.tsx
```diff
@@ -105,4 +106,12 @@ export function getStatus(state: PullRequestStateEnum) {
 	} else {
 		return 'Closed';
 	}
+}
+
+export function getActionText(state: PullRequestStateEnum) {
```
RMacfarlane: nit: no need to export

smashwilson: Un-exported 👍

Issue #1213, opened 2019-06-05
Sign in command not working


RMacfarlane:
Repro steps:
1. Use the 'Sign out' command if signed in
2. Run 'Sign in' from the command palette

Nothing happens.

Clicking on the status bar icon or the notification to sign in both work, only the command in the command palette fails.

PR #1214, opened 2019-06-05
Fix command palette sign in command

RMacfarlane:
Fixes https://github.com/microsoft/vscode-pull-request-github/issues/1213

The `pr.signin` command was previously reading a property that hadn't been initialized. Changed to read from remotes directly instead.


RMacfarlane:
Since `this._githubRepositories` is initialized to an empty array, it's hard right now to check if its already up to date. And since sign in should be run infrequently, I think it's ok to make the expensive call for now

PR #1229, opened 2019-06-10
Create pipelines for nightly releases and CI

### LINKED ISSUE #1225: Continuous integration
Let's set up an Azure DevOps pipeline in a dedicated project for this repo to help keep our tests green.

Note that @rebornix has a pipeline under his personal account that's currently performing nightly releases.

smashwilson:
YAML configuration files to run CI on each pull request and push. Move the logic for the nightly release pipeline in-repo while I'm at it.

Fixes #1225.

- [x] Port existing build into YAML from the Azure UI.
- [x] Add a new pipeline that runs the test suite for each push and pull request.
- [x] Incorporate the code indexing job from the "pull request validation" pipeline.
- [x] Report test result metadata to Azure DevOps.
- [x] Bonus points: Checks API integration.
- [x] Capture and report stdout and stderr output during tests.
- [x] Fix the actual test failures


src/test/common/utils.test.ts
```diff
@@ -10,21 +9,6 @@ describe('utils', () => {
 			assert.equal(utils.formatError(error), 'No!');
 		});
 
-		it('should format an HttpError from octorest', (done) => {
```
smashwilson: This test was unreliable in CI because it relied on an actual network response. I couldn't find a different way to force Octokit to respond with an error without going into things like [nock](https://www.npmjs.com/package/nock) which felt redundant to the higher-level QueryProvider infra we already have, and we already test the branch that Octokit errors fall into in the previous case.

src/test/globalHooks.ts
```diff
@@ -0,0 +1,30 @@
...
+	console.log = function captureLog() {
+		original.log.apply(console, arguments);
+		const formatted = util.format.apply(util, arguments);
+		currentTest.consoleOutputs = (currentTest.consoleOutputs || []).concat(formatted);
+	};
+	console.error = function captureError() {
+		original.error.apply(console, arguments);
+		const formatted = util.format.apply(util, arguments);
+		currentTest.consoleErrors = (currentTest.consoleErrors || []).concat(formatted);
+	};
```
smashwilson: This _does_ result in stdout and stderr sections being placed in the JUnit XML output, but it doesn't look like Azure DevOps is exposing it for anything. It shouldn't hurt though.

smashwilson:
I've validated the nightly build pipeline by queueing tests manually:

* [Azure Pipelines build](https://rebornix.visualstudio.com/Pull%20Request/_build/results?buildId=343)

image

PR #1242, opened 2019-06-20
Consolidate merge event telemetry

RMacfarlane:


RMacfarlane:
Closing, will do this as part of moving over the a new telemetry endpoint

Issue #221, opened 2018-08-16
Whole commit line should be clickable not only commit hash


isidorn:
In the description view of the PR if there is a commit attached, it is only possible to click on the commit hash to follow that commit.
instead the whole commit line should be clickable (just like on github)

![screen shot 2018-08-16 at 16 54 35](https://user-images.githubusercontent.com/1926584/44216451-42471380-a175-11e8-85ec-f154ba826be4.png)

tobudim:
Hi, I'd like to do my first contribution here, if someone could tell me where to look. Thanks!

RMacfarlane:
Thanks @tobudim! Here's a pointer to the code that renders the commits:
https://github.com/microsoft/vscode-pull-request-github/blob/master/preview-src/timeline.tsx#L39-L50

tobudim:
Superb! Thank you for your help, I'm working on this then.

tobudim:
[I did a pull request](https://github.com/microsoft/vscode-pull-request-github/pull/1260), I don't know if I did it well since it does not appear here.
Thanks for your guidance @RMacfarlane . Please tell me if I did right!

PR #1260, opened 2019-07-08
Description view of PR : whole commit line clickable

tobudim:
This PR comes from: [Whole commit line should be clickable not only commit hash #221](https://github.com/microsoft/vscode-pull-request-github/issues/221).

**Initially**
Only the hash were clickable on commit lines in the description view of a pull request.

**Now**
All the line is clickable and leads to the current commit.

**_Little note_**
This is my first contribution to a project. I'm willing to hear advices and best practices if needed.

### DIFF
--- a/preview-src/timeline.tsx
+++ b/preview-src/timeline.tsx
@@ -37,7 +37,7 @@ export const Timeline = ({ events }: { events: TimelineEvent[] }) =>
 export default Timeline;
 
 const CommitEventView = (event: CommitEvent) =>
-	<div className='comment-container commit'>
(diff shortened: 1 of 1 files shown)


RMacfarlane:
@tobudim Thanks for making a contribution! One suggestion I have is to make the commit message a link instead of the whole commit. I think this would match GitHub more closely:
Screen Shot 2019-07-08 at 4 29 53 PM

I think the easiest way to achieve that would be to keep the sha a link, and change message also into a link

tobudim:
Thank you @RMacfarlane , beginning to contribute with your help is a pleasure.
I made the changes as you evoked.

PR #1261, opened 2019-07-09
Update icons to match new style

misolori:
See https://github.com/Microsoft/vscode/issues/8017#issuecomment-509349133, this updates all of the icons to match the new style. Also tried fixing some alignment/spacing issues with the new sizes:

image


RMacfarlane:
Thanks @misolori! There’s only one icon that looks a bit weird to me, the `diff.svg` that’s shown when a reviewer has requested changes:
Screen Shot 2019-07-12 at 12 02 17 PM
Is there anything that looks more like the original? Now that I know the filename is `diff` it makes sense to me, but when I first saw it in this context I was confused about what it was. Maybe it would be better to remove that icon from the extension and just use the ‘x’ instead?

misolori:
@RMacfarlane thanks, let me see if I can find an alternative that matches that or create a similar one.

misolori:
I've update the requested changes icon to match https://octicons.github.com/icon/request-changes/

image

PR #1271, opened 2019-07-16
update nightly version

rebornix:
Considering that current version of nightly breaks the pipeline, we can now simply follow the way how Remote SSH does versioning `YEAR.MONTH.MINUTES`.

The rule for releasing nightly manually is 

* Updating `patch` version only


rebornix:
@RMacfarlane that's right, as long as we are not releasing twice in a minute, it should always updates successfully.

PR #1277, opened 2019-07-18
Limit gh pr commands to itself

rebornix:
To avoid it bleed into comment widgets created by other extensions.


rebornix:
right, they are all leaked.

PR #1281, opened 2019-07-22
Fix issue with file path on windows for custom file system providers

RMacfarlane:
Currently, the result of `filePath` for the Uri that we create on windows is something like `"e:/dev/pullrequest-demo/data/products.json"`. This should have a leading `/` - for extensions that are registering their own git API and using a custom file system provider, the file system provider API in vscode will throw when trying to open the URI if it doesn't have this.

### DIFF
--- a/src/view/reviewManager.ts
+++ b/src/view/reviewManager.ts
@@ -374,7 +374,7 @@ export class ReviewManager implements vscode.DecorationProvider {
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


src/view/reviewManager.ts
```diff
@@ -374,7 +374,7 @@ export class ReviewManager implements vscode.DecorationProvider {
 				}
 			}
 
-			const filePath = nodePath.resolve(this._repository.rootUri.fsPath, change.fileName).replace(/\\/g, '/');
```
eljog: test comment from extension

Issue #798, opened 2019-01-08
Pull request template not working


tcigrand:
Issue Type: <b>Bug</b>

I'm working in a repo with a [pull request template](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/) and the description doesn't get pre-populated with the template, like it does on GitHub.

Extension version: 0.3.2
VS Code version: Code 1.30.2 (61122f88f0bf01e2ac16bdb9e1bc4571755f5bd8, 2019-01-07T22:48:31.260Z)
OS version: Darwin x64 18.0.0

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz (4 x 3500)|
|GPU Status|2d_canvas: enabled<br>checker_imaging: disabled_off<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: enabled<br>rasterization: enabled<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|3, 3, 3|
|Memory (System)|16.00GB (2.73GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|

</details>

johndangerstorey:
I also run into this.  We have a template that would be nice to include when PR is created from VScode.  Thanks :)

ammorrise:
Same here. It would be amazing to be able to edit the PR description before creating but at very least including PULL_REQUEST_TEMPLATE.md as the default description so it can be edited later is a must for me to be able to use this to create PRs.

DocEmmetBrown:
I would enjoy having this feature too, it would be super helpful.
Thanks ! ;)

RMacfarlane:
Just checked in a change for this, when creating a pull request, the extension will now look for a PR template and use it as the default description (or show a quick pick if the repo has multiple templates).

There are still open issues on improving the create experience to allow editing the title and description before create, so we still have work to do there.

bmbermn:
Im not sure why but when I create PR's via VSCODE I am still not getting the template pulled in, is there anything I need to do to enable this functionality?

RMacfarlane:
[...]

PR #1283, opened 2019-07-25
Look for pull request templates during PR creation

RMacfarlane:
Fixes https://github.com/microsoft/vscode-pull-request-github/issues/798

During PR creation, look to see if there is a pull request template file. If there's only one, use it for the description. If there are multiple, show a quick pick with the file paths for them. If there are none, then just use the commit information as before.

There's no GitHub API that I can see for fetching the template, so instead I'm doing a file search against the repo. To be completely correct, this should actually be a find against the files as they are in the default branch of the repo, but that would require a large number of network calls since the templates can be located in a variety of places, and with arbitrary names in some cases. Or checking out the default branch, doing the search, and returning to the previous branch. The current approach should be much faster than that and correct a vast majority of the time.

### DIFF
--- a/src/view/reviewManager.ts
+++ b/src/view/reviewManager.ts
@@ -650,13 +650,8 @@ export class ReviewManager implements vscode.DecorationProvider {
 	}
 
 	private async getPullRequestTitleAndDescriptionDefaults(progress: vscode.Progress<{message?: string, increment?: number}>): Promise<{ title: string, description: string } | undefined> {
-		progress.report({ increment: 5, message: 'Looking for pull request template...' });
 		const pullRequestTemplates = await this._prManager.getPullRequestTemplates();
 		let template: vscode.Uri | undefined;
-
-		if (pullRequestTemplates.length === 0) {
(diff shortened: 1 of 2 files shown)


src/view/reviewManager.ts
```diff
@@ -649,6 +649,57 @@ export class ReviewManager implements vscode.DecorationProvider {
 		return selected;
 	}
 
+	private async getPullRequestTitleAndDescriptionDefaults(progress: vscode.Progress<{message?: string, increment?: number}>): Promise<{ title: string, description: string } | undefined> {
+		progress.report({ increment: 5, message: 'Looking for pull request template...' });
+		const pullRequestTemplates = await this._prManager.getPullRequestTemplates();
+		let template: vscode.Uri | undefined;
+
+		if (pullRequestTemplates.length === 0) {
+			progress.report({ increment: 5, message: 'No pull request template found. Creating pull request...' });
```
rebornix: I'm wondering if this should be transparent to users who are not aware of Pull Request Templates.

PR #1284, opened 2019-07-26
Consume the git extension state API

RMacfarlane:
The git extension API now has a 'state' property that is either 'initialized' or 'uninitialized', more info here: https://github.com/microsoft/vscode/issues/77787

Use this when reading repository information, wait for it to be initialized before updating repository information. On Windows, this removes the flicker between "No GitHub repositories found." and the full tree

Note: this was just added, so I've been testing out of the OSS build, not insiders


src/github/pullRequestManager.ts
```diff
@@ -143,6 +145,11 @@ export class PullRequestManager implements vscode.Disposable {
 			}
 		}));
 
+		this._subs.push(this._git.onDidChangeState(async _ => {
+			await this.updateRepositories();
+			vscode.commands.executeCommand('pr.refreshList');
```
rebornix: Is this `pr.refreshList` redundant? If the state changes, `this.updateRepositories` will update `PullRequestManager.state` and trigger an udpate of the prs tree.

PR #1319, opened 2019-09-03
Remove unneeded validateDraftMode calls

RMacfarlane:
Only make a network call to validate draft mode on initialization and during update, as the methods for starting and deleting reviews should already correctly set the state.

Also made the interval for polling for updates longer, since I think it's very aggressive now. Let me know if you think we should keep it shorter.


RMacfarlane:
Yeah, it does appear that draft mode isn't being validated enough now: https://github.com/microsoft/vscode-pull-request-github/issues/1343

I'm going to revert this change, I think this causes more trouble than it's worth for the speed gain

Issue #1320, opened 2019-09-03
Checkout button should be disabled while checking out


RMacfarlane:
Ditto the "Exit Review Mode" button

PR #1323, opened 2019-09-04
Disable 'checkout' and 'exit review mode' buttons while action in progress

RMacfarlane:
Fixes https://github.com/microsoft/vscode-pull-request-github/issues/1320

### DIFF
--- a/src/github/pullRequestOverview.ts
+++ b/src/github/pullRequestOverview.ts
@@ -305,7 +305,7 @@ export class PullRequestOverviewPanel {
 			case 'pr.submit':
 				return this.submitReview(message);
 			case 'pr.checkout-default-branch':
-				return this.checkoutDefaultBranch(message.args);
+				return this.checkoutDefaultBranch(message);
 			case 'pr.comment':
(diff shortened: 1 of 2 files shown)


src/github/pullRequestOverview.ts
```diff
@@ -672,37 +672,30 @@ export class PullRequestOverviewPanel {
...
 
-	private async checkoutDefaultBranch(branch: string): Promise<void> {
+	private async checkoutDefaultBranch(message: IRequestMessage<string>): Promise<void> {
 		try {
+			const branch = message.args;
 			// This should be updated for multi-root support and consume the git extension API if possible
 			const branchObj = await this._pullRequestManager.repository.getBranch('@{-1}');
 
-			if (branch === branchObj.name) {
+			if (branchObj.upstream && branch === branchObj.upstream.name) {
```
RMacfarlane: `branchObj.name` seems to always be `@{-1}`. I can't remember what our original intention was with this logic, but this statement was previously always false

Issue #906, opened 2019-02-01
Choose name for PR


Ghostbird:
I don't like one thing about #661: It defaults the PR title to the last commit message. However, whenever I make a PR, my last commit is generally not the main commit of the branch, but a clean-up or maintenance commit. Which means that my PRs are almost always created with nonsensical titles.

I suggest this workflow:

0. Open Command Palette
1. _Github Pull Requests: Create Pull Request_
2. Pop-up asks for branch
3. Enter branch
5. Press <kbd>Enter</kbd>
6. Pop-up asks for PR name
    * Pre-filled with last commit message
    * Pre-selected so starting to type immediately erases the old value.
7. [_Optional_] Type new PR name
8. Press <kbd>Enter</kbd>

To get the same results as the current workflow, the only change is that you have to double-tap <kbd>Enter</kbd> to immediately accept the suggested PR name. In the current workflow you would press <kbd>Enter</kbd> only once. I think this is an acceptable trade-off.

GeorgeWL:
I'd also suggest to make this an opt-in in the settings however, as some people do the workflow of commit message titles

JCQuintas:
I agree with @daedalus28, I would expect a github tool to behave just like github web interface does. The way it is implemented right now feels a bit clumsy, where you have to create the PR and then clean up the things it did that you didn't really want or need.

phifa:
Maybe a setting, where you can choose where the suggested title is coming from, either from the branch name or last commit, or custom.

GeorgeWL:
I don't see why it can't be a select box with:

```html
<select>
     <option>Use Latest Commit Message</option
     <option>Use Branch Name</option
     <option>Ask with Prompt Every Time</option> // default to if press esc, use last commit message
</select>
```

Ghostbird:
Thanks so much for this feature!

GeorgeWL:
Yeah this is honestly one of the best features added this year.

GeorgeWL:
> More generally, I think just being able to open a draft of a PR where we can edit any of the properties (assignee, labels, etc) before actually creating it is probably the way to go. That would give it feature parity with github's web interface for creating PRs (which this is attempting to displace).

Draft PR's I feel should be a feature, but also should have their own Issue for it.

RMacfarlane:
[...]

PR #1366, opened 2019-10-09
allow the naming of PRs

nminchow:
addresses: https://github.com/microsoft/vscode-pull-request-github/issues/906

I couldn't find any tests covering this logic, but I did extensive manual testing.

Settings view:
image

Prompt for name source when `ask` is set:
image

Naming prompt when `custom` selected:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
             }
           }
         },
-        "githubPullRequests.pullRequestName": {
+        "githubPullRequests.pullRequestTitle": {
           "type": "string",
           "enum": [
             "commit",
@@ -70,11 +70,11 @@
           "enumDescriptions": [
             "Use the latest commit message",
             "Use the branch name",
-            "Specify a custom message",
+            "Specify a custom title",
             "Ask which of the above methods to use"
           ],
           "default": "ask",
-          "description": "The name used when creating pull requests."
+          "description": "The title used when creating pull requests."
         },
         "githubPullRequests.logLevel": {
           "type": "string",
--- a/package.json
+++ b/package.json
@@ -68,10 +68,10 @@
             "ask"
           ],
           "enumDescriptions": [
-            "Use the latest commit message.",
-            "Use the branch name.",
-            "Specify a custom message.",
-            "Ask which of the above methods to use."
+            "Use the latest commit message",
+            "Use the branch name",
+            "Specify a custom message",
+            "Ask which of the above methods to use"
           ],
           "default": "ask",
           "description": "The name used when creating pull requests."
--- a/package.json
+++ b/package.json
@@ -59,6 +59,23 @@
             }
           }
         },
+        "githubPullRequests.pullRequestName": {
+          "type": "string",
+          "enum": [
+            "commit",
+            "branch",
+            "custom",
+            "ask"
+          ],
+          "enumDescriptions": [
+            "Use the latest commit message.",
+            "Use the branch name.",
+            "Specify a custom message.",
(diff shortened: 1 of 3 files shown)


Ghostbird:
Thanks a lot, I'd like to see this merged.
I keep running into #906 especially as software that integrates Github and triggers on new pull request creation, often does not trigger it's functionality correctly when the PR is created with the wrong name. JIRA for example needs the PR name to include an issue key, and doesn't link issues to PRs when the PR is created without the key and is later changed to include it.

The option to use the branch name by default is a nice touch too.

package.json
```diff
@@ -59,6 +59,23 @@
...
+            "ask"
+          ],
+          "enumDescriptions": [
+            "Use the latest commit message",
+            "Use the branch name",
+            "Specify a custom message",
+            "Ask which of the above methods to use"
+          ],
+          "default": "ask",
+          "description": "The name used when creating pull requests."
```
RMacfarlane: Let's use "title" instead of "name" here to be consistent with the GitHub UI. Ditto for setting name

src/view/quickpick.ts
```diff
@@ -22,3 +22,28 @@ export class RemoteQuickPickItem implements vscode.QuickPickItem {
 		public label = `${owner}:${name}`,
 	) {}
 }
+
+export type PullRequestNameSource = 'commit' | 'branch' | 'custom' | 'ask';
```
RMacfarlane: Please create an enum for this

src/view/reviewManager.ts
```diff
@@ -704,6 +704,25 @@ export class ReviewManager implements vscode.DecorationProvider {
 		};
 	}
 
+	private async getPullRequestNameSetting(): Promise<PullRequestNameSource | undefined> {
+		const method = vscode.workspace.getConfiguration('githubPullRequests').get<PullRequestNameSource>('pullRequestName', 'ask');
+
+		if(method == "ask") {
```
RMacfarlane: Looks like there are a couple of tslint warnings in this file. They're printed as part of the build, but at the top, so easy to miss. Using the tslint extension should show them in the editor

src/view/reviewManager.ts
```diff
@@ -704,6 +704,25 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+			const titleSource = await vscode.window.showQuickPick<PullRequestNameSourceQuickPick>(PullRequestNameSourceQuickPick.allOptions(), {
+				ignoreFocusOut: true,
+				placeHolder: "Pull Request Name Source"
```
RMacfarlane: 'Pull Request Title Source'

src/view/reviewManager.ts
```diff
@@ -764,8 +783,37 @@ export class ReviewManager implements vscode.DecorationProvider {
...
+			}
+
+			switch(pullRequestNameMethod) {
+				case "branch":
+					if(branchName) {
+						title = branchName;
+					}
+					break;
+				case "custom":
+					const nameResult = await vscode.window.showInputBox({
```
RMacfarlane: I think it would be good to add a `validateInput` option for this input box to prevent the user from entering with an empty title. Would be handled fine below and just halt the whole process, but I think it would be better UX to display a warning in that case, and only treat escape as cancel

nminchow:
@RMacfarlane - all requested changes should be implemented 😄

moritzWa:
Was this feature removed? I'm not able to find it.
image

It's annoying that the default name is the last commit. A sensible default would be the branch name.

Issue #1369, opened 2019-10-09
Display pending label for all pending comments on the description page


RMacfarlane:
![Screen Shot 2019-10-09 at 2 37 22 PM](https://user-images.githubusercontent.com/3672607/66522533-80eb0c00-eaa2-11e9-91b6-6bfa9905239a.png)

Comments that are pending should be marked with a label like they are on GitHub. In this example, I replied twice to an existing comment. Both "test" comments have not yet been posted, but look like they are already appended to the existing review.

![Screen Shot 2019-10-09 at 2 41 32 PM](https://user-images.githubusercontent.com/3672607/66522703-ed660b00-eaa2-11e9-8364-78a797bd1808.png)

akosasante:
👋  I'd like to take a look at resolving this

RMacfarlane:
@akosasante Great! The code that renders the comments is here: https://github.com/microsoft/vscode-pull-request-github/blob/1432e8ab94c2b046e810b08ea14c21d71f15b543/preview-src/comment.tsx#L20

PR #1380, opened 2019-10-14
Issue 1369: Add pending label to comments of pending reviews on timeline

akosasante:
Resolves issue #1369 

Screen Shot 2019-10-14 at 10 26 14 AM

Added label and manually tested to confirm that it shows up only on comments for pending reviews. Added some minor styling to make the label more obvious like on GH, but am happy to leave as just an `em` tag, to match with the pending label already being used in the Diff view.

### DIFF
--- a/preview-src/comment.tsx
+++ b/preview-src/comment.tsx
@@ -15,6 +15,7 @@ import { useStateProp } from './hooks';
 export type Props = Partial<IComment & PullRequest> & {
 	headerInEditMode?: boolean
 	isPRDescription?: boolean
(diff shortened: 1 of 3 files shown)


akosasante:
Ah thanks, let me take a second look; I think I did see that property but somehow was getting the label showing up on non-draft comments. I may have been applying the conditional incorrectly or something though.

akosasante:
Ah I see, I must've previously passed in `isDraft` property of the PR object itself instead of just using the one belonging to the comment class. All fixed up now 😄

Issue #1381, opened 2019-10-14
Incorrect GraphQL URL for github enterprise


yulrizka:
- Extension version: 0.12.0
- VSCode Version:  1.40.0-insider
- OS: Osx
- Github Enterprise version: GitHub Enterprise Server 2.16.12

Steps to Reproduce:

1. Login with github enterprise
2. See the log message
```
[Info] https://github.[company].org/api/v3: GraphQL not supported (Network error: Response not successful: Received status code 404)
```
I think this is related to https://github.com/microsoft/vscode-pull-request-github/issues/947

This is caused by invalid graphQL URL for the enterprise version. The reason is that this function will fallback to `/api/v3` for URL other than github.com

https://github.com/microsoft/vscode-pull-request-github/blob/aae86fdfd2045e1250cef44782b08c047e9a6a85/src/authentication/configuration.ts#L18-L25

But when constructing the ApolloClient, it append the base url with `/graphql`. which the URL now becomes `/api/v3/graphql`. That's why it returns 404.

https://github.com/microsoft/vscode-pull-request-github/blob/aae86fdfd2045e1250cef44782b08c047e9a6a85/src/github/credentials.ts#L316-L320

Before fixing this, I couldn't create a pending PR review and comments. New comments was immediately posted as single comments (not as part of the pending review). I believe this is because it's using the REST API instead of the graphQL underneath.

How ever, once i use the correct graphQL URL. Loading Description page of PR no longer works. Got error with this 
```
[Debug 130796s 117ms] PullRequestOverviewPanel> initialized without PR
[Debug 130796s 320ms] GitHubRepository> Fetch available merge methods - done
[Debug 130796s 328ms] GithubRepository> Unable to fetch PR: Error: GraphQL error: Field 'isDraft' doesn't exist on type 
```
Not sure why that is. But at least now I can use Draft PR functionalities

IllusionMH:
Looks like #1382 resolved this issue but didn't use `fixes #1381` to trigger auto-close.

yulrizka:
It should, i might missed the '#' character in the PR title

PR #1382, opened 2019-10-14
Fix invalid graphql for github enterprise, Fixes 1381

yulrizka:
see https://github.com/microsoft/vscode-pull-request-github/issues/1381

### DIFF
--- a/preview-src/cache.ts
+++ b/preview-src/cache.ts
@@ -38,7 +38,7 @@ export interface PullRequest {
 	mergeMethodsAvailability: MergeMethodsAvailability;
 	supportsGraphQl: boolean;
 	reviewers: ReviewState[];
-	isDraft: boolean;
+	isDraft?: boolean;
 }
 
 export function getState(): PullRequest {
--- a/src/github/credentials.ts
+++ b/src/github/credentials.ts
@@ -18,6 +18,8 @@ import { handler as uriHandler } from '../common/uri';
 import { createHttpLink } from 'apollo-link-http';
 import fetch from 'node-fetch';
 import { ITelemetry } from '../common/telemetry';
+const defaultSchema = require('./queries.gql');
+const enterpriseSchema = require('./enterprise.gql');
 
 const TRY_AGAIN = 'Try again?';
 const SIGNIN_COMMAND = 'Sign in';
@@ -29,6 +31,7 @@ const AUTH_INPUT_TOKEN_CMD = 'auth.inputTokenCallback';
 export interface GitHub {
 	octokit: Octokit;
 	graphql: ApolloClient<NormalizedCacheObject> | null;
+	schema: any | null;
 }
 
 export class CredentialStore implements vscode.Disposable {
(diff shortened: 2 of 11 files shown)


yulrizka:
This functionality is not fully tested for github enterprise, This is because it now start using graphQL instead of the REST API. 

For example, at the time of writing, there is a bug when loading the Description page
```
[Debug 130796s 117ms] PullRequestOverviewPanel> initialized without PR
[Debug 130796s 320ms] GitHubRepository> Fetch available merge methods - done
[Debug 130796s 328ms] GithubRepository> Unable to fetch PR: Error: GraphQL error: Field 'isDraft' doesn't exist on type 
```

I think best to do some regresion test before releasing this commit.

yulrizka:
The error about isDraft I think because enterprise doesn't support this yet

src/github/queries.gql
```diff
@@ -248,7 +248,6 @@ query PullRequest($owner: String!, $name: String!, $number: Int!) {
 			mergeable
 			id
 			databaseId
-			isDraft
```
RMacfarlane: I'm wondering if we could still allow fetching this property in cases where it is supported. One option could be to have two different queries for PullRequest - if the first one fails with error about isDraft not existing, fall back to the second that doesn't have it. It's also possible to have a query to see what fields are available (https://developer.github.com/v4/guides/intro-to-graphql/#discovering-the-graphql-api)

yulrizka: interesting, so there is now way to basically request with is draft field, and server can response with something like null if they couldn't provide it?
I mean we basically need to use 2 different query?

RMacfarlane: yeah, the server will just throw instead of sending back a null or undefined value. It looks like there is a way to conditionally include things in the query: https://graphql.org/learn/queries/#directives

So we could use that, after either querying to see if the 'isDraft' field shows up for the endpoint, or just based on whether this is enterprise or github.com

yulrizka: Hmm this won't be great for future maintainability I'm afraid. I Could imagine if there are 5 version of the API. Each one is adding a new field it's going to get hairy. But I couldn't think of a better solution at the moment

RMacfarlane: Yeah, graphql is supposed to be "versionless", but since GitHub enterprise will ship with certain versions of each schema, it makes it really hard to work with. (At least this is my understanding, maybe older enterprise versions eventually get the new fields added to them?) I think the most correct fix would be to first run the query to see what the current schema looks like

```
query {
  __type(name: "PullRequest") {
    fields {
      name
    }
  }
}
```

and then run the actual pull request query with the subset of fields that we want to fetch that are part of that result. This would be very annoying to have to do before every query though. I think `isDraft` is the only thing we are likely to have a problem with right now, it's currently marked as a preview feature: https://developer.github.com/v4/object/pullrequest/

RMacfarlane: @yulrizka Do you want to take a look at making that fix? I'm also happy to make the change

yulrizka: Hi @RMacfarlane sorry I didn't not follow up. I was trying something but didn't finish it. Feel free to pick this up. Unfortunately I'm a bit swamp these and the next couple of weeks. Not sure if I can pick it up.

yulrizka: What I did try is to have different schema for github (existing query.gql) and another one (enterprise.gql). And in the `credentials.js` detect which version this is running (the enterprise return a header with a version). Based on that it will figure the correct schema and store this on the `GithHub` interface 

The benefit of this is that the controller or the view logic is simpler because it doesn't have to do extra or introspection query. and easily support different version of the API in the future. But unfortunately didn't manage to get it to work yet.

yulrizka: Hmm I think i managed to get it to work:  https://github.com/microsoft/vscode-pull-request-github/compare/master...yulrizka:wip-github-enterprise?expand=1

yulrizka: what do think @RMacfarlane? or is it maybe too much?

RMacfarlane: I think it looks pretty good! I'm a little bit worried about potentially having to maintain separate `.gql` files for different enterprise versions, but I actually like that this makes it very explicit what properties we expect for both cases. Right now I think there is very little API difference between enterprise versions and we could always use the schema of the oldest supported version. I think this change would be reasonable to merge.

yulrizka:
@RMacfarlane, I think this PR is ready, but i couldn't test it in GH enterprise. Unfortunately I don't have any instance to test it wih.

PR #1387, opened 2019-10-17
Support images in 'review' and `pr` diff viewers

### LINKED ISSUE #1356: Adding gif in PR shows "An error occurred while loading the image"
- Extension version: 2019.9.38342
- VSCode Version: 1.39.0-insider
- OS: macOS 10.14

Steps to Reproduce:

1. Have a PR that adds a gif (like https://github.com/microsoft/vscode-extension-samples/pull/108)
2. Open the Files -> demo.gif from "Changes in pull request"

image

JacksonKearl:
Fixes #1356.

Uses a trick from the builtin git extension to represent images as data URL's, as the `provideTextDocument` API does not handle images. There is a `fileSystemProvider` API that we could use instead, but it ends up being more complicated.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -8,7 +8,7 @@ import * as vscode from 'vscode';
 import * as pathLib from 'path';
 import { ReviewManager } from './view/reviewManager';
 import { PullRequestOverviewPanel } from './github/pullRequestOverview';
-import { fromReviewUri, ReviewUriParams, asImageDataURI } from './common/uri';
+import { fromReviewUri, ReviewUriParams, asImageDataURI, EMPTY_IMAGE_URI } from './common/uri';
 import { GitFileChangeNode, InMemFileChangeNode } from './view/treeNodes/fileChangeNode';
 import { CommitNode } from './view/treeNodes/commitNode';
 import { PRNode } from './view/treeNodes/pullRequestNode';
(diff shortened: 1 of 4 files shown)


src/common/uri.ts
```diff
@@ -43,9 +45,34 @@ export interface GitUriOptions {
...
+		const { size, object } = await repository.getObjectDetails(commit, uri.fsPath);
+		const { mimetype } = await repository.detectObjectType(object);
+
+		if (mimetype === 'text/plain') {
+			return;
+		}
+
+		if (ImageMimetypes.indexOf(mimetype) > -1) {
+			const contents = await repository.buffer(commit, uri.fsPath);
+			return Uri.parse(`data:${mimetype};label:${path.basename(uri.fsPath)};description:${commit};size:${size};base64,${contents.toString('base64')}`);
```
RMacfarlane: nit: tslint warning about `path` variable being shadowed, perhaps rename the import

src/common/uri.ts
```diff
@@ -43,9 +45,34 @@ export interface GitUriOptions {
...
+	'image/gif',
+	'image/jpeg',
+	'image/webp',
+	'image/tiff',
+	'image/bmp'
+];
+
+export async function asImageDataURI(uri: Uri, repository: Repository): Promise<Uri | undefined> {
+	try {
+		const { commit } = JSON.parse(uri.query);
```
RMacfarlane: for `pr` scheme uris, the query has `headCommit`, `baseCommit`, `isBase`. We could try to use those or just short-circuit for non-review scheme uris

src/commands.ts
```diff
@@ -160,7 +162,18 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: Pu
 		if (isPartial) {
 			vscode.window.showInformationMessage('Your local repository is not up to date so only partial content is being displayed');
 		}
-		vscode.commands.executeCommand('vscode.diff', parentFilePath, filePath, fileName, opts);
+
+		// if this is an image, encode it as a base64 data URI
+		const imageDataURI = await asImageDataURI(parentFilePath, prManager.repository);
+		const originalURI = imageDataURI || parentFilePath;
+
+		if (fileChangeNode.status === GitChangeType.ADD) {
```
RMacfarlane: I'm not sure if we should do this or not - on the one hand it adds a bunch of screen real estate when viewing these files, but it also makes it less obvious at first glance if they are an add or delete. My preference is for the old way

src/common/uri.ts
JacksonKearl: Done!

src/commands.ts
JacksonKearl: Moved to using a transparent pixel URI, which is what the builtin diff view does.

src/common/uri.ts
JacksonKearl: Changed this and the call up a bit, now we can view added, modified, and deleted images in both `pr` and `review` schemes.

Issue #198, opened 2018-08-15
Tree looses focus on mouse click


bpasero:
Focus should not leave the tree on mouse down, otherwise you break keyboard navigation.

![image](https://user-images.githubusercontent.com/900690/44139789-682f25d6-a078-11e8-84ee-e6a8bbb7897b.png)

bpasero:
@rebornix also happens when I click into a comment of the comment panel

PR #1388, opened 2019-10-17
Both Focus and Select tree item on triggering secondary action

JacksonKearl:
Fixes #1217 

Previously, on triggering a secondary action, the tree view item would gain selected status, but not focused status. Additionally, previously, the 'openFile' command would move focus to the editor, hiding this issue when `git.openDiffOnClick` was true, and reducing keyboard accessibility (#198).

This PR forces the selected item to also gain focus, and prohibits focus from transferring to editor on click.


JacksonKearl:
@RMacfarlane done!

PR #1399, opened 2019-10-22
add quote reply shortcut

### LINKED ISSUE #1396: Support quote reply
~1. support `@`  autocomplete~
2. support `Quote reply` shortcut

shaodahong:
Closed #1396 

@RMacfarlane Hi, this my first pull request, and `quote reply` need a new icon, so what I to do?

### DIFF
--- a/preview-src/comment.tsx
+++ b/preview-src/comment.tsx
@@ -8,9 +8,10 @@ import Timestamp from './timestamp';
 import { IComment } from '../src/common/comment';
 import { PullRequest } from './cache';
 import PullRequestContext from './context';
-import { editIcon, deleteIcon } from './icon';
+import { editIcon, deleteIcon, commentIcon } from './icon';
 import { PullRequestStateEnum } from '../src/github/interface';
 import { useStateProp } from './hooks';
+import emitter from './events';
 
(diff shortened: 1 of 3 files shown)


shaodahong:
now I use `commentIcon`, but I don't know this icon suitable for it

preview-src/comment.tsx
```diff
@@ -64,6 +65,7 @@ export function CommentView(comment: Props) {
 		onMouseLeave={() => setShowActionBar(false)}
 	>{ ((canEdit || canDelete) && showActionBar)
```
RMacfarlane: Let's update this to just `showActionBar` so you can quote reply even if you don't have permissions to edit or delete the comment

shaodahong:
and have a question, I don't  find `.editorconfig` or '.prettierrc', so document changed when I format document, but I don't modify the document

RMacfarlane:
Yeah, there is no `.editorconfig` or `.prettierrc` for this project. There are also no linting rules for this formatting. Prettier and the built in VSCode JSON formatter both format arrays in the same way, so I think the change is OK. We had just never run format document on this before, and have the `Format on Save` setting turned off.

PR #1402, opened 2019-10-23
Preserve tree loaded pages on reload. Fixes #1036.

### LINKED ISSUE #1036: Checking out a PR in the viewlet can hide the checked out PR from view
ext version: 1.67.0

Repro:

1. Open a repo with many PRs and multple remotes, click continue fetching from other remotes
2. Checkout One of the PRs that was loaded by fetching more

This will cause the pull requests view to reload and reintroduce the continue fetching from other remotes, hiding the actual checked out PR from that view.

image

After clicking again and scrolling down:

image

JacksonKearl:
Now, when `getPullRequests` is called with `fetchNextPage: false`, it will refetch all previously fetched pages, rather than just the first.


JacksonKearl:
Hm, doesn't seem to work for "Continue fetching from other remotes" pages, looking into it.

PR #1403, opened 2019-10-23
Preserve tree state on reload

### LINKED ISSUE #1036: Checking out a PR in the viewlet can hide the checked out PR from view

JacksonKearl:
Fixes #1036.

### DIFF
--- a/src/github/pullRequestManager.ts
+++ b/src/github/pullRequestManager.ts
@@ -628,8 +628,20 @@ export class PullRequestManager implements vscode.Disposable {
 		this._telemetry.sendTelemetryEvent('branch.delete');
 	}
 
-	// Keep track of how many repos we've pulled from, so when we reload we pull from the same ones.
-	private _numFetchedRepos = 0;
+	// Keep track of how many pages we've fetched for each query, so when we reload we pull the same ones.
+	private totalFetchedPages = new Map<string, number>();
+
+	/**
+	 * This method works in three different ways:
+	 * 1) Initialize: fetch the first page of the first remote that has pages
+	 * 2) Fetch Next: fetch the next page from this remote, or if it has no more pages, the first page from the next remote that does have pages
+	 * 3) Restore: fetch all the pages you previously have fetched
+	 *
+	 * When `options.fetchNextPage === false`, we are in case 2.
+	 * Otherwise:
+	 *   If `this.totalFetchQueries[queryId] === 0`, we are in case 1.
+	 *   Otherwise, we're in case 3.
+	 */
 	async getPullRequests(type: PRType, options: IPullRequestsPagingOptions = { fetchNextPage: false }, query?: string): Promise<PullRequestsResponseResult> {
 		if (!this._githubRepositories || !this._githubRepositories.length) {
 			return {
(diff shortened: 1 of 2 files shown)


src/github/pullRequestManager.ts
```diff
@@ -647,31 +639,66 @@ export class PullRequestManager implements vscode.Disposable {
...
+			}
+
+			pageInformation.hasMorePages = pullRequestData.hasMorePages;
+
+			const repoNumber = this._githubRepositories.findIndex(
+				repo => repo.remote.url.toString() === githubRepository.remote.url.toString()) + 1;
 
-			pageInformation.hasMorePages = !!pullRequestData && pullRequestData.hasMorePages;
-			pageInformation.pullRequestPage++;
+			this._numFetchedRepos = Math.max(this._numFetchedRepos, repoNumber);
```
RMacfarlane: Can this be simplified to Math.max(this._numFetchedRepos, i)? I'm hitting an off-by-one problem in a repo that has `origin` with no prs and `upstream` with actual results, `repoNumber` gets set to 2

src/github/pullRequestManager.ts
```diff
@@ -647,31 +639,66 @@ export class PullRequestManager implements vscode.Disposable {
...
+			const fetchPage = async (pageNumber: number) =>
+				type === PRType.All
+					? await githubRepository.getAllPullRequests(pageNumber)
+					: await githubRepository.getPullRequestsForCategory(query || '', pageNumber);
+
+			if (options.fetchNextPage) {
+				pageInformation.pullRequestPage++;
+				addPage(await fetchPage(pageInformation.pullRequestPage));
+			} else {
+				if (pageInformation.pullRequestPage === 0) { pageInformation.pullRequestPage = 1; }
```
RMacfarlane: should `pullRequestPage` be initialized to 1 above instead?

JacksonKearl: That would cause the "continue searching in other remotes" action, which triggers a "fetchNextPage" command, to pull from page 2. 

This whole thing is a bit tricky because there are three behaviours:
1) Initialize: fetch the first page of the first remote that has pages
2) Fetch Next: fetch the next page from this remote, or if it has no more pages, the first page from the next remote that does have pages
3) Restore (new): fetch all the pages you previously have fetched

And only 2 options: `fetchNextPage: true | false`.

So when `fetchNextPage: true` we do 2, and when it's `false`, we do either 1 or 3. The initial state of 0 `pullRequestPage`s fetched so far is what distinguishes 1 and 3.

Working on a better solution now...

src/github/pullRequestManager.ts
JacksonKearl: Moved to a different strategy for keeping track of past fetches.

src/github/pullRequestManager.ts
JacksonKearl: Just pushed a reworking of the logic.

src/view/treeNodes/categoryNode.ts
```diff
@@ -184,7 +184,7 @@ export class CategoryTreeNode extends TreeNode implements vscode.TreeItem {
 				}
 			} else {
 				try {
-					const response = await this._prManager.getPullRequests(this._type, { fetchNextPage: true });
+					const response = await this._prManager.getPullRequests(this._type, { fetchNextPage: true }, this._categoryQuery);
```
JacksonKearl: This was missing previously, causing "fetch more" on "created by me" (for instance), to just load the next page of "all"

src/github/pullRequestManager.ts
```diff
@@ -687,13 +740,6 @@ export class PullRequestManager implements vscode.Disposable {
 		};
 	}
 
-	public mayHaveMorePages(): boolean {
```
JacksonKearl: Unused and not really fundamentally compatible with the fact that there are many different queries that may or may not have more pages.

Issue #1065, opened 2019-03-23
edit title and description of others' PRs when allowed


adrinjalali:
The maintainers are allowed to edit the description and the title of other people's PRs, but this extension only allows that for one's own PRs at the moment.

PR #1404, opened 2019-10-24
Show PR actions only to users with push access (fixes #967)

### LINKED ISSUE #967: Pull request overview shows merge functionality when user does not have rights to merge
Steps to Reproduce:

1. Using a user that does not have rights to merge on a given repository
2. Open the pull request overview
3. Observe that "Merge pull request"/"Close pull request" buttons are enabled

*I did not click them out of fear, uncertainty and doubt*

IllusionMH:
This PR changes visibility rules for PR actions (Merge, Mark as ready for review) which were previously available to all users.

- "Merge Pull Request" button, merge method dropdown, and "Close Pull Request" button only available to users with push permission

- "Ready for review" button only available for PR author and users with push permission

With `canMerge` in place it is now possible to change visibility of more controls - edit title/description #1065 and add/remove reviewers and labels (not sure if there is tracking issue).
Should I change visibility rules for these places too or in separate PR?

### DIFF
--- a/preview-src/cache.ts
+++ b/preview-src/cache.ts
@@ -25,6 +25,11 @@ export interface PullRequest {
 	commitsCount: number;
 	repositoryDefaultBranch: any;
 	canEdit: boolean;
+	/**
+	 * Users with push access to repo have rights to merge/close PRs,
+	 * edit title/description, assign reviewers/labels etc.
+	 */
+	hasWritePermission: boolean;
 	pendingCommentText?: string;
 	pendingCommentDrafts?: { [key: string]: string; };
 	status: ReposGetCombinedStatusForRefResponse;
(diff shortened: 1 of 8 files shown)


RMacfarlane:
Really nice!

Yeah, let's handle the other visibility changes in a different PR. (Currently, there is no issue for hiding the add/remove buttons for reviewers/labels UI when there aren't sufficient permissions.)

Would it make more sense to name `canMerge` as `canPush`? I suppose in either case there needs to be some explanations in comments about what that permission grants

IllusionMH:
I had several options for name, however this PR was limited to scope of Merge controls so I proceed with `canMerge` which is similar to existed `canEdit`.
However may be `hasWriteAccess` be more suitable that `canPush`?
And about comments - it will be enough to add JSDoc comment in `cache.ts` or what places do you have in mind?
image

RMacfarlane:
That makes sense. My preference is for`hasWriteAccess`, I think that is the clearest descriptor of what it is. Adding the JSDoc comment in `cache.ts` should be enough

PR #1410, opened 2019-10-25
Configure extension on statusbar item click.

### LINKED ISSUE #360: Remove "Signed in" status bar item
Follow up from https://github.com/Microsoft/vscode-pull-request-github/issues/274

The status bar item that shows that I am signed in to GH does not seem to have any purpose other than taking away space from the already crowded status bar. It should be removed in its current form.

Reasons to keep it (maybe, under discussion):
* clicking brings me to PR views
* clicking offers actions to quickly checkout to another PR or back to the branch where I left from

JacksonKearl:
And add "Sign out of GitHub.." to extension configuration quick pick.
Closes #360.

### DIFF
--- a/src/view/prsTreeDataProvider.ts
+++ b/src/view/prsTreeDataProvider.ts
@@ -60,7 +60,8 @@ export class PullRequestsTreeDataProvider implements vscode.TreeDataProvider<Tre
 		this._childrenDisposables = [];
 
 		this._disposables.push(vscode.commands.registerCommand('pr.configurePRViewlet', async () => {
-			const configuration = await vscode.window.showQuickPick(['Configure Remotes...', 'Configure Queries...', 'Sign out of GitHub...']);
+			const isLoggedIn = this._prManager.state === PRManagerState.RepositoriesLoaded;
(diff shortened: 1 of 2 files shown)


src/view/prsTreeDataProvider.ts
```diff
@@ -70,6 +70,8 @@ export class PullRequestsTreeDataProvider implements vscode.TreeDataProvider<Tre
 					return vscode.commands.executeCommand('workbench.action.openSettings', `@ext:${extensionId} queries`);
 				case 'Configure Remotes...':
 					return vscode.commands.executeCommand('workbench.action.openSettings', `@ext:${extensionId} remotes`);
+				case 'Sign out of GitHub...':
```
RMacfarlane: I think it would be better to either create a new command, or conditionally include this. Otherwise, even if you're not signed in and you click the gear on the PR tree, you see an option to sign out, which seems unexpected

JacksonKearl: Oh, good point.

JacksonKearl: Fixed

PR #1417, opened 2019-10-26
Add back "Outdated" label

### LINKED ISSUE #1407: Display of outdated comments on description page
- Extension version: 0.12.0

Steps to Reproduce:

1. Open the description page on a PR that has outdated comments

I believe there was previously a box rendered near the top of the comment saying "Outdated". At a minimum, we should bring that back. Even better would be having a collapsed view of these by default as they are on github.com and a way of toggling their visibility

JacksonKearl:
Fixes #1407.

Removed some CSS that to me made it look worse, might want to find some new CSS for this?


JacksonKearl:
image
The new look

PR #1425, opened 2019-10-29
api - adopt vscode.env.asExternalUri

bpasero:


RMacfarlane:
@bpasero With either these changes or with the old `createAppUri` method, it seems that the url handler isn't being triggered on the callback. has that logic changed at all?

bpasero:
No, but there could be other bugs in URI creation/parsing that cause this.

Issue #1408, opened 2019-10-25
Display add/remove controls for reviewers and labels only to users with write access


IllusionMH:
Currently any user can see controls to add reviewers and labels which might be confusing (PR author might try to request review which will lead to an error notification).

With PR #1404 it is now possible to make Overview closed to web version and display these controls only to users who can make changes.

PR #1428, opened 2019-10-30
Show suggested reviewers and limit list to assignable users

### LINKED ISSUE #1424: Reviewers should be listed in a "relevance" order
Currently, the "Add Reviewers" popup lists potential reviewers in an order that I don't understand. It's not alphabetical, it doesn't appear to have anything to do with people who've recently interacted with the files in the PR, or even the repo itself. I simply see everyone in my organization.

image

It would be helpful if "relevant" people were at the top of the list, however you might define "relevant". Perhaps that's a simple local cache of previously-selected users; perhaps you would be able to use a GitHub API similar to the one that the GH site itself uses to suggest specific reviewers based on people who've edited the relevant files.

image

If nothing else, I'd prefer to have the list of possible reviewers sorted in alphabetical order (as seen in the lower section of the GH screenshot above) instead of the seemingly-random order that I get now.

IllusionMH:
Fixes #1424 

This PR provides next improvements to add reviewers (quick)pick list:
1. Suggested reviewers are displayed at the top of the pick list with suggestion reason below login and name. 
It is possible to add ⭐ before their logins to make suggested reviewers look more distinctive, but skipped this until review.

1. Only assignable users are displayed in pick list instead of all mentionable users. Big improvement for list in some repos.

1. List is now sorted by `login` (sort is case insensitive, same as quick pick built-in filter)

image

The only difference from GitHub that I've noticed is that I'm in list of suggested reviewers even if I already posted review - [example from same PR](https://user-images.githubusercontent.com/1312662/67826407-0f1c4600-fad5-11e9-9d8b-82d5fabc35b0.png).

It may be related to fact that GitHub has "Re-request review" for existing reviewers, but AFAIK this extension doesn't support this action so I left filtering out existing reviewers from suggested at this point.

### DIFF
--- a/src/github/graphql.ts
+++ b/src/github/graphql.ts
@@ -205,14 +205,30 @@ export interface PullRequestCommentsResponse {
 export interface MentionableUsersResponse {
 	repository: {
 		mentionableUsers: {
-			nodes: [
-				{
-					login: string;
-					avatarUrl: string;
-					name: string;
-					url: string;
-				}
-			];
+			nodes: {
+				login: string;
+				avatarUrl: string;
+				name: string;
+				url: string;
+			}[];
+			pageInfo: {
+				hasNextPage: boolean;
+				endCursor: string;
+			};
+		}
+	};
+	rateLimit: RateLimit;
+}
+
+export interface AssignableUsersResponse {
+	repository: {
+		assignableUsers: {
+			nodes: {
+				login: string;
+				avatarUrl: string;
+				name: string;
+				url: string;
+			}[];
 			pageInfo: {
 				hasNextPage: boolean;
 				endCursor: string;
@@ -300,6 +316,17 @@ export interface Ref {
 	};
 }
 
+export interface SuggestedReviewerResponse {
+	isAuthor: boolean;
+	isCommenter: boolean;
+	reviewer: {
+		login: string;
+		avatarUrl: string;
+		name: string;
+		url: string;
+	};
+}
+
 export interface PullRequestResponse {
 	repository: {
 		pullRequest: {
@@ -323,11 +350,12 @@ export interface PullRequestResponse {
 			labels: {
 				nodes: {
 					name: string;
-				}[],
+				}[];
 			}
 			merged: boolean;
 			mergeable: 'MERGEABLE' | 'CONFLICTING' | 'UNKNOWN';
 			isDraft?: boolean;
+			suggestedReviewers?: SuggestedReviewerResponse[];
 		}
 	};
 	rateLimit: RateLimit;
--- a/src/github/graphql.ts
+++ b/src/github/graphql.ts
@@ -355,7 +355,7 @@ export interface PullRequestResponse {
 			merged: boolean;
 			mergeable: 'MERGEABLE' | 'CONFLICTING' | 'UNKNOWN';
 			isDraft?: boolean;
-			suggestedReviewers?: SuggestedReviewerResponse[];
+			suggestedReviewers: SuggestedReviewerResponse[];
 		}
 	};
 	rateLimit: RateLimit;
--- a/src/github/pullRequestManager.ts
+++ b/src/github/pullRequestManager.ts
@@ -20,7 +20,7 @@ import { Repository, RefType, UpstreamRef } from '../api/api';
 import Logger from '../common/logger';
 import { EXTENSION_ID } from '../constants';
 import { fromPRUri } from '../common/uri';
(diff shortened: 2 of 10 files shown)


src/github/graphql.ts
```diff
@@ -205,14 +205,30 @@ export interface PullRequestCommentsResponse {
...
-					name: string;
-					url: string;
-				}
-			];
+			nodes: {
+				login: string;
+				avatarUrl: string;
+				name: string;
+				url: string;
+			}[];
```
IllusionMH: Now it is properly typed as an array instead of a tuple with one element.
Not directly related to PR, but changed to match `AssignableUsersResponse`

src/github/pullRequestOverview.ts
```diff
@@ -340,30 +340,74 @@ export class PullRequestOverviewPanel {
...
+				}
+			);
 
 			if (reviewersToAdd) {
 				await this._pullRequestManager.requestReview(this._pullRequest, reviewersToAdd.map(r => r.label));
 				const addedReviewers: ReviewState[] = reviewersToAdd.map(reviewer => {
 					return {
-						reviewer: newReviewers.find(r => r.login === reviewer.label)!,
+						// assumes that suggested reviewers will be a subset of assignable users
+						reviewer: assignableUsers.find(r => r.login === reviewer.label)!,
```
IllusionMH: From what I've seen in a few repositories - suggested reviewers always were present in assignable users list.
Are there cases when it won't be true?

src/github/pullRequestManager.ts
```diff
@@ -998,6 +1030,46 @@ export class PullRequestManager implements vscode.Disposable {
 		return;
 	}
 
+	async getSuggestedReviewers(pullRequest: PullRequestModel): Promise<SuggestedReviewer[]> {
```
IllusionMH: Initially implemented as separate call which will be triggered on each click on Add Reviewers.

On the other hand it is possible to get this data in [`getPullRequest`](https://github.com/microsoft/vscode-pull-request-github/blob/f39d8159f98da9cc36d8caa576ff30347f0c688b/src/github/githubRepository.ts#L318) and make this data available on `PullRequestModel`. In this way it will be fetched/updated only when user opens Description node or clicks Refresh which looks good enough.

Should I move this data to `PullRequestModel` instead of separate call?

IllusionMH:
I also found [`collaborators` connection](https://developer.github.com/v4/object/repository/#collaborators) on `repository`. It is accessible only to existing collaborators (should be OK when #1408 resolved, however will require additional checks) and in only repo where I was able to retrieve this data it was same as `assignableUsers`. I don't know what's exact difference between those lists.
Which one I should use for this PR?

src/github/pullRequestManager.ts
IllusionMH: After implementations comparison `suggestedReviewers` on `PullRequestModel` looks better that with separate request - pushed updated `PullRequestModel`

RMacfarlane:
I found this doc on different permissions levels: https://help.github.com/en/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization#repository-access-for-each-permission-level

The row "Have an issue assigned to them" is checked for all permissions levels, so it seems to me like `assignableUsers` would always be identical to `collaborators`. From the `collaborators` connection you could also get information about what their permission level is, but seems a bit odd to have the same users returned on both, so maybe I'm still missing something. Continuing to use `assignableUsers` makes sense to me.

src/github/pullRequestManager.ts
RMacfarlane: Nice, looks good

src/github/pullRequestOverview.ts
RMacfarlane: I think this is a safe assumption. I just tried removing a collaborator who was previously a suggested reviewer and they were removed from both lists

src/github/queries.gql
```diff
@@ -369,4 +379,27 @@ query GetMentionableUsers($owner: String!, $name: String!, $first: Int!, $after:
 		remaining
 		resetAt
 	}
-}
\ No newline at end of file
+}
+
```
RMacfarlane: can you also update the `enterprise.gql` file? this query and the above should work with the oldest supported github enterprise version https://developer.github.com/enterprise/2.16/v4/object/repository/

IllusionMH: Checked it yesterday to see if it has support and missed/failed somehow

IllusionMH:
@RMacfarlane I saw this doc however I still have some problems with understanding levels across different repos: looks like doc is only for orgs and doesn't mention collaborators role, however personal repos only has collaborators list.

Looks like Collaborator = Write from table, and only Write+ can "Submit reviews that affect a pull request's mergeability". Still cant find info if it still possible to ask for review from person even if their review can't affect mergability.

In any case - `assignableUsers` can be changed to `collaborators` later if needed.

IllusionMH:
Thanks for review and suggestions!

curtisgibby:
This looks fantastic! I can't wait to try it out! Thanks folks!

PR #1438, opened 2019-11-10
Context menu entries for opening all files

### LINKED ISSUE #1248: "Open all files" button
Issue Type: <b>Feature Request</b>

It'd be nice to have a quick way to open all changed files in a PR in their own tabs. Idk, just a thought. 

Extension version: 0.8.0
VS Code version: Code 1.35.1 (c7d83e57cd18f18026a8162d042843bda1bcf21f, 2019-06-12T14:29:22.216Z)
OS version: Darwin x64 18.6.0

JeffreyCA:
Closes #1248.

I added context menu entries mirroring the single-selection context menu entries:

- Open All Diff Views
- Open All Modified Files
- Open All Original Files

### Demo
output2

### Notes
- Because of some of the asynchronous nature of the code, I had to add `await` to several calls like `vscode.commands.executeCommand();`, otherwise files would open in new tabs in an arbitrary order every time "Open all Files" is activated. However, the main downside is that opening a large number of files may take noticeable amount of time because it waits until each "open file" command is finished executing, as seen in the video. Not sure if this is the best approach.

- `FilesCategoryNode` has a private field `_fileChanges` which stores a list of all changed files, but the ordering is different than just traversing straight down the file tree, which is why I did not opt for that method.

- A confirmation dialog is displayed when the user tries to open 10 or more files all at once.

- Currently only supports opening files under the current Pull Request (Changes in Pull Request viewlet).

Open to feedback and suggestions on improving how this all should work.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -152,44 +152,68 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: Pu
 		vscode.commands.executeCommand('vscode.open', e.filePath);
 	}));
 
-	context.subscriptions.push(vscode.commands.registerCommand('pr.openAllFiles', (e: DirectoryTreeNode | FilesCategoryNode) => {
-		const options: vscode.TextDocumentShowOptions = {
+	context.subscriptions.push(vscode.commands.registerCommand('pr.openAllFiles', async (e: DirectoryTreeNode | FilesCategoryNode) => {
+		// Minimum number of files for warning message to be shown
+		const MIN_FILES_FOR_WARNING = 10;
+		const openOptions: vscode.TextDocumentShowOptions = {
 			preserveFocus: true,
 			preview: false
 		};
+		const messageOptions: vscode.MessageOptions = {
+			modal: true
+		}
 
 		/**
-		 * Opens all files in tree by recursive traversal.
-		 * @param node Tree node
+		 * Populates Uri array with file paths of all files under given node.
+		 * @param node Files category or directory node
 		 */
-		function open(node: FilesCategoryNode | DirectoryTreeNode | GitFileChangeNode) {
-			if (node instanceof GitFileChangeNode) {
-				vscode.commands.executeCommand('vscode.open', node.filePath, options);
-				return;
+		async function populateFilePaths(node: FilesCategoryNode | DirectoryTreeNode): Promise<vscode.Uri[]> {
+			var filePaths: vscode.Uri[] = []
+
+			if (node instanceof FilesCategoryNode) {
+				const fileChanges = await node.getFileChanges();
+				filePaths = fileChanges.filter(fc => fc instanceof GitFileChangeNode).map(fc => fc.filePath)
+			} else if (node instanceof DirectoryTreeNode) {
+				populateFilePathsHelper(filePaths, node);
 			}
 
-			if (node instanceof DirectoryTreeNode) {
+			return filePaths;
+		}
+
+		/**
+		 * Populate array with Uris of all GitFileChangeNode file paths under a given node.
+		 * @param filePaths Array of Uris to be populated
+		 * @param node Current node
+		 */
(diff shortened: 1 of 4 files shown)


src/commands.ts
```diff
@@ -140,22 +142,95 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: Pu
...
+			vscode.window.showWarningMessage(`Are you sure you want to open ${nodes.length} files?`, messageOptions, 'Yes').then(async option => {
+				if (option === 'Yes') {
+					// Ensure files are processed in the same order
+					for (const node of nodes) {
+						await command(node);
+					}
+				}
+			});
+		} else {
+			// Ensure files are processed in the same order
```
RMacfarlane: I'm torn on this - I like having the files opened in order, but as you said, it does make the action take a noticeable amount of time. I think I prefer the snappiness of opening them in parallel to having them in exactly tree order. @JacksonKearl @rebornix thoughts?

rebornix: It's a good idea but opening tons of editors and switching between them automatically looks a bit scary to me.

Having them opening in order sounds right to me as I would love to see them being opened in the predictable order. However, only the first editor should be opened and then focused, all following editors should only be hidden tabs. You can select multiple files in the File Explorer and right click to open them, you can see that they are all opened but you are only focusing in the first one.

I'm not sure yet if our API allows that but IMHO that's the right behavior if we want to take this feature in.

JeffreyCA: I agree, it should function like selecting files in Explorer and clicking **Open to the Side**.

Found some relevent code [here](https://github.com/microsoft/vscode/blob/87064db1033a9d8d768cd7ed676083413c8a222c/src/vs/workbench/contrib/files/browser/fileCommands.ts#L286) where that command is registered.

I tried calling `vscode.commands.executeCommand('explorer.openToSide', ___);` but it expects a single resource URI as the argument, not a list of URIs. It gets the list of URIs to open from the File, List, and Editor Services. If I call `explorer.openToSide` multiple times, a new column is created in the editor window every time, which is not what we want.

Does some additional work need to be done in `vscode` then?

rebornix:
Taking both UX and performance into account, I thought about it again and probably we can take another approach to solve the underling problem described in #1248 .

Currently, following things happen when a diff is being opened

* GH PR generate file paths for original and modified content, send both file paths to VS Code 
* Diff Editor will try to load both files
* GH PR needs to resolve the content for both file paths, either reading from local disk or calculating based the diff hunks
* Diff Editor compare the content from original and modified files, and render them with decorations

As you can see, opening a diff file provided by GH PR is pretty costly. If what we want in  #1248 is reducing the time for opening multiple diff files, it can never be satisfied: as long as a file is not opened and kept visible, our model service will dispose it when necessary. It means even if we open 3 diff editors within one tab in a UX-friendly way, only one model stays alive most of the time. If we switch another tab, the diff editor still needs to ask GH PR to provide the right content before switching them upfront. 

If the problem we try to solve in #1248 is allowing users to switch to next modified file from the editor tabs, instead of going back to the tree view, I think what we can do is contributing an editor action for the diff editor, which opens the next modified file when clicked.

@JacksonKearl @RMacfarlane troughs?

JacksonKearl:
@rebornix The problem I had in #1248 isn't really a matter of time or easily switching from one file to the next, its more that when I'm reviewing changes I follow a process something where I basically open all the files and close them one by one as I finish requesting changes or I'm satisfied with the file's change. Once there are no more tabs left, the review is over. 

In this use case, the ordering of the tabs being consistent with the order in the tree isn't particularly important, as the order I want to review them in is likely not whatever order they happen to be in the tree. So I think making the open calls async is fine. By the same token, having a "show next modified" wouldn't be as helpful, as this forces me to review linearly, and it doesn't allow for marking "done" like closing a tab does.

alexr00:
@JeffreyCA thank you for the PR! We have since added a "Open All Files" command with https://github.com/microsoft/vscode-pull-request-github/issues/3047.

PR #1477, opened 2020-01-14
Add additional Github API usage

alexr00:


src/github/pullRequestModel.ts
```diff
@@ -31,7 +31,12 @@ export class PullRequestModel implements IPullRequestModel {
 	public localBranchName?: string;
 	public mergeBase?: string;
 	public isDraft?: boolean;
-	public suggestedReviewers: ISuggestedReviewer[];
+	public suggestedReviewers?: ISuggestedReviewer[];
+	public milestone?: IMilestone;
+
+	public get isIssue(): boolean {
```
RMacfarlane: Instead of having an issue be a type of pull request, I think the reverse would be more accurate. Could we have a separate `IssueModel` that `PullRequestModel` would extend?

src/github/queries.gql
```diff
@@ -270,6 +290,39 @@ query PullRequest($owner: String!, $name: String!, $number: Int!) {
 	}
 }
 
+query Issue($owner: String!, $name: String!, $number: Int!) {
+	repository(owner: $owner, name: $name) {
+		pullRequest: issue(number: $number) {
```
RMacfarlane: I'm not sure if this is any faster, but it looks like there's a direct connection from `repository` to `issues`, so `pullRequest:` could be removed here and above

src/github/pullRequestManager.ts
```diff
@@ -771,6 +771,101 @@ export class PullRequestManager implements vscode.Disposable {
...
+			};
+		}
+
+		const queryId = PRType.All.toString();
+		const getTotalFetchedPages = () => this.totalFetchedPages.get(queryId) || 0;
+		const setTotalFetchedPages = (numPages: number) => this.totalFetchedPages.set(queryId, numPages);
+
+		for (const repository of this._githubRepositories) {
+			const remoteId = repository.remote.url.toString() + queryId;
+			if (!this._repositoryPageInformation.get(remoteId)) {
```
RMacfarlane: This map stores the number of pages we've fetched so far for PRs so that we can restore to the correct state when the tree refreshes. Right now it looks like this reuses the key for all PRs, so loading more PRs would cause issues to also get paged. I think we should use an issue-specific key in the map

src/github/pullRequestManager.ts
```diff
@@ -771,6 +771,101 @@ export class PullRequestManager implements vscode.Disposable {
 		};
 	}
 
+	async getIssues(options: IPullRequestsPagingOptions = { fetchNextPage: false }): Promise<PullRequestsResponseResult> {
+		if (!this._githubRepositories || !this._githubRepositories.length) {
```
RMacfarlane: Is it possible to extract this into something like `fetchPagedData` which could take a flag to determine if pull requests or issues should be requested?

src/github/pullRequestModel.ts
alexr00: I started down that path initially, but it involves large amounts of refactoring. Would you like me to do it in this PR, or in an immediate follow up PR to keep things easier to review?

src/github/queries.gql
alexr00: I _thought_ I was just renaming the property by doing this so that I could re-use more code. Is that not the case?

src/github/pullRequestModel.ts
RMacfarlane: Splitting it up into a follow up PR sounds good to me 👍

src/github/queries.gql
RMacfarlane: Ah, I see, you're able to use the `PullRequestResponseType` this way! I think that's fine for now. When doing the follow up PR to split out an `IssueModel`, I think it would be good to change this to query issues instead of doing the rename, and have a separate type for it.

src/github/queries.gql
```diff
@@ -270,6 +290,63 @@ query PullRequest($owner: String!, $name: String!, $number: Int!) {
 	}
```
RMacfarlane: I forgot to mention this - there's also an `enterprise.gql` file that is nearly identical to this. This is meant to track the schema of the oldest supported version of GitHub enterprise, which in some cases is missing properties compared to this file. Even though I don't have a way of testing enterprise, I'd prefer to keep it (theoretically) working. Can you copy these changes there?

src/github/queries.gql
```diff
@@ -270,6 +290,63 @@ query PullRequest($owner: String!, $name: String!, $number: Int!) {
...
+			labels(first: 50) {
+				nodes {
+					name
+				}
+			}
+			id
+			databaseId
+		}
+	}
+	rateLimit {
```
RMacfarlane: we query for the rate limit in a lot of places but then ultimately don't use this information, this can be removed

src/github/pullRequestModel.ts
alexr00: That's what I'll do then!

src/github/queries.gql
alexr00: 👍

src/github/queries.gql
alexr00: Copied!

src/github/queries.gql
alexr00: Removed from all new queries.

PR #1481, opened 2020-01-27
Introduce issue concept

alexr00:
The only functional changes should be those required to have the concept of an issue

### DIFF
--- a/src/github/interface.ts
+++ b/src/github/interface.ts
@@ -15,7 +15,7 @@ export enum ReviewEvent {
 	Comment = 'COMMENT'
 }
 
-export enum PullRequestStateEnum {
+export enum GithubItemStateEnum {
 	Open,
 	Merged,
 	Closed,
@@ -71,7 +71,7 @@ export interface ILabel {
 	name: string;
 }
 
-export interface PullRequest {
+export interface Issue {
 	id: number;
 	graphNodeId: string;
 	url: string;
@@ -83,15 +83,18 @@ export interface PullRequest {
 	assignee?: IAccount;
 	createdAt: string;
 	updatedAt: string;
-	head?: IGitHubRef;
-	base?: IGitHubRef;
 	user: IAccount;
 	labels: ILabel[];
+	isDraft?: boolean;
+	milestone?: IMilestone;
+}
+
+export interface PullRequest extends Issue {
+	head?: IGitHubRef;
+	base?: IGitHubRef;
 	merged?: boolean;
 	mergeable?: PullRequestMergeability;
-	isDraft?: boolean;
 	suggestedReviewers?: ISuggestedReviewer[];
-	milestone?: IMilestone;
 }
 
 export interface IRawFileChange {
--- a/src/github/interface.ts
+++ b/src/github/interface.ts
@@ -85,11 +85,11 @@ export interface Issue {
 	updatedAt: string;
 	user: IAccount;
 	labels: ILabel[];
-	isDraft?: boolean;
 	milestone?: IMilestone;
 }
 
 export interface PullRequest extends Issue {
+	isDraft?: boolean;
 	head?: IGitHubRef;
 	base?: IGitHubRef;
 	merged?: boolean;
--- a/src/github/issueModel.ts
+++ b/src/github/issueModel.ts
@@ -24,9 +24,6 @@ export class IssueModel {
 	public readonly remote: Remote;
 	public item: Issue;
 	public bodyHTML?: string;
-	private _inDraftMode: boolean = false;
-	private _onDidChangeDraftMode: vscode.EventEmitter<boolean> = new vscode.EventEmitter<boolean>();
-	public onDidChangeDraftMode = this._onDidChangeDraftMode.event;
 
 	constructor(githubRepository: GitHubRepository, remote: Remote, item: Issue) {
 		this.githubRepository = githubRepository;
@@ -64,17 +61,6 @@ export class IssueModel {
 		}
 
 		return undefined;
-	}
-
-	public get inDraftMode(): boolean {
-		return this._inDraftMode;
-	}
-
-	public set inDraftMode(inDraftMode: boolean) {
-		if (this._inDraftMode !== inDraftMode) {
-			this._inDraftMode = inDraftMode;
-			this._onDidChangeDraftMode.fire(this._inDraftMode);
-		}
 	}
 
 	public get body(): string {
--- a/src/github/issueModel.ts
+++ b/src/github/issueModel.ts
@@ -11,7 +11,7 @@ import { IAccount, Issue, GithubItemStateEnum, IMilestone } from './interface';
 export class IssueModel {
 	public id: number;
 	public graphNodeId: string;
-	public githubNumber: number;
+	public number: number;
 	public title: string;
 	public html_url: string;
 	public state: GithubItemStateEnum = GithubItemStateEnum.Open;
@@ -96,7 +96,7 @@ export class IssueModel {
 	update(issue: Issue): void {
 		this.id = issue.id;
 		this.graphNodeId = issue.graphNodeId;
-		this.githubNumber = issue.number;
+		this.number = issue.number;
 		this.title = issue.title;
 		this.bodyHTML = issue.bodyHTML;
 		this.html_url = issue.url;
@@ -116,7 +116,7 @@ export class IssueModel {
 			return false;
 		}
 
-		if (this.githubNumber !== other.githubNumber) {
+		if (this.number !== other.number) {
 			return false;
 		}
 
--- a/src/github/issueModel.ts
+++ b/src/github/issueModel.ts
@@ -19,7 +19,6 @@ export class IssueModel {
 	public assignee: IAccount;
 	public createdAt: string;
 	public updatedAt: string;
-	public isDraft?: boolean;
 	public milestone?: IMilestone;
 	public readonly githubRepository: GitHubRepository;
 	public readonly remote: Remote;
@@ -101,8 +100,8 @@ export class IssueModel {
 		this.bodyHTML = issue.bodyHTML;
 		this.html_url = issue.url;
 		this.author = issue.user;
-		this.isDraft = issue.isDraft;
 		this.milestone = issue.milestone;
+		this.createdAt = issue.createdAt;
 
 		this.updateState(issue.state);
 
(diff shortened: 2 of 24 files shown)


src/github/pullRequestManager.ts
```diff
@@ -372,13 +373,20 @@ export class PullRequestManager implements vscode.Disposable {
...
+	}
+
+	get activeItem(): (IssueModel | undefined) {
+		return this._activeGithubItem;
 	}
 
-	set activePullRequest(pullRequest: (PullRequestModel | undefined)) {
-		this._activePullRequest = pullRequest;
-		this._onDidChangeActivePullRequest.fire();
+	set activeItem(issue: (IssueModel | undefined)) {
```
RMacfarlane: For pull requests, the `activePullRequest` represents the one that is associated with the branch you're currently on/one you have checked out. Is there an analog for issues? When would this be set?

src/github/issueModel.ts
```diff
@@ -0,0 +1,129 @@
...
+
+import * as vscode from 'vscode';
+import { Remote } from '../common/remote';
+import { GitHubRepository } from './githubRepository';
+import { IAccount, Issue, GithubItemStateEnum, IMilestone } from './interface';
+
+export class IssueModel {
+	public id: number;
+	public graphNodeId: string;
+	public githubNumber: number;
```
RMacfarlane: nit: I think just `number` would be fine

src/github/issueOverview.ts
```diff
@@ -0,0 +1,416 @@
...
+		return Promise.all([
+			this._pullRequestManager.resolveIssue(
+				issueModel.remote.owner,
+				issueModel.remote.repositoryName,
+				issueModel.githubNumber
+			),
+			this._pullRequestManager.getIssueTimelineEvents(issueModel),
+			this._pullRequestManager.getPullRequestRepositoryDefaultBranch(issueModel),
+		]).then(result => {
+			const [pullRequest, timelineEvents, defaultBranch] = result;
```
RMacfarlane: local variables and error messages should use "issue" here

src/github/issueOverview.ts
```diff
@@ -0,0 +1,416 @@
...
+			if (this._pullRequestManager && this._item) {
+				const isCurrentlyCheckedOut = this._item.equals(this._pullRequestManager.activeItem);
+				this._postMessage({
+					command: 'pr.update-checkout-status',
+					isCurrentlyCheckedOut: isCurrentlyCheckedOut
+				});
+			}
+		}, null, this._disposables);
+
+		onDidUpdatePR(pr => {
```
RMacfarlane: could be moved to `pullRequestOverview`

src/github/pullRequestManager.ts
RMacfarlane: Oh, I guess the idea is that you could have a branch/PR specifically targeting an issue, so there would be a similar notion of it being checked out? In that case, it seems like you could have both an active pull request and issue at the same time, so perhaps exposing `activeIssue` instead would be better?

src/github/pullRequestManager.ts
```diff
@@ -1033,7 +1041,7 @@ export class PullRequestManager implements vscode.Disposable {
 		return;
 	}
 
-	async validateDraftMode(pullRequest: PullRequestModel): Promise<boolean> {
+	async validateDraftMode(pullRequest: IssueModel): Promise<boolean> {
```
RMacfarlane: I don't think it's possible to have draft issue, so let's keep this as `PullRequestModel` here on on `updateDraftModeContext`

src/github/pullRequestManager.ts
```diff
@@ -1044,13 +1052,13 @@ export class PullRequestManager implements vscode.Disposable {
...
-	async updateDraftModeContext(pullRequest: PullRequestModel) {
-		if (this._activePullRequest && this._activePullRequest.prNumber === pullRequest.prNumber) {
+	async updateDraftModeContext(pullRequest: IssueModel) {
+		if (this._activeGithubItem && this._activeGithubItem.githubNumber === pullRequest.githubNumber) {
 			await vscode.commands.executeCommand('setContext', 'reviewInDraftMode', pullRequest.inDraftMode);
 		}
 	}
 
-	async getPendingReviewId(pullRequest = this._activePullRequest): Promise<string | undefined> {
+	async getPendingReviewId(pullRequest = this._activeGithubItem): Promise<string | undefined> {
```
RMacfarlane: same here, since you can't add reviews to issues, I think this should remain `this._activePullRequest`

src/github/pullRequestOverview.ts
```diff
@@ -464,20 +294,23 @@ export class PullRequestOverviewPanel {
 	}
 
 	private async addReviewers(message: IRequestMessage<void>): Promise<void> {
+		if (!issueAsPullRequest(this._item)) {
```
RMacfarlane: can this check be removed? shouldn't `this._item` always be a PR?

src/github/pullRequestManager.ts
```diff
@@ -939,21 +947,21 @@ export class PullRequestManager implements vscode.Disposable {
...
 			per_page: 100
 		});
-		Logger.debug(`Fetch issue comments of PR #${pullRequest.prNumber} - done`, PullRequestManager.ID);
+		Logger.debug(`Fetch issue comments of PR #${pullRequest.githubNumber} - done`, PullRequestManager.ID);
 
 		return promise.data;
 	}
 
-	async createIssueComment(pullRequest: PullRequestModel, text: string): Promise<IComment> {
+	async createIssueComment(pullRequest: IssueModel, text: string): Promise<IComment> {
```
RMacfarlane: nit: `issue: IssueModel`

RMacfarlane: or perhaps `item: IssueModel | PullRequestModel`? might be clearer

src/github/pullRequestManager.ts
alexr00: I was thinking of it more as the branch you have checked out would only have one github item associated with it. You're right though that it could have multiple. For now, since it doesn't change existing behavior, can we stick with only having one item associated with it?

src/github/issueModel.ts
alexr00: Changed!

src/github/issueOverview.ts
alexr00: Changed

src/github/issueOverview.ts
alexr00: 👍

src/github/pullRequestManager.ts
alexr00: There is no calling code that would pass in an IssueModel instead of a PullRequestModel here, but there's nothing stopping us from doing that some day. Since it dosn't need to be a PullRequestModel, can we leave it as an IssueModel?

src/github/pullRequestManager.ts
alexr00: Changed

src/github/pullRequestManager.ts
alexr00: Changed to issue

src/github/pullRequestOverview.ts
alexr00: Leftover check from while refactoring was in progress. Removed!

src/github/pullRequestManager.ts
RMacfarlane: If this is meant to only track the active issue, let's name it `activeIssue`. If it's meant to be for both PRs and issues, then I think it makes `activePullRequest` unclear and redundant. This would be especially confusing if we did support having an active issue and an active pull request at the same time. IMO it should be the responsibility of the calling code to use either the active issue or active pull request as needed

src/view/reviewCommentController.ts
```diff
@@ -124,7 +124,7 @@ export class ReviewCommentController implements vscode.Disposable, CommentHandle
 	}
 
 	async initializeWorkspaceCommentThreads(): Promise<void[]> {
-		await this._prManager.validateDraftMode(this._prManager.activePullRequest!);
+		await this._prManager.validateDraftMode(this._prManager.activeItem!);
```
RMacfarlane: I think this should still be `activePullRequest`. The comment controller here is specifically for review comments, those on files, which issues wouldn't have.

src/view/treeNodes/pullRequestNode.ts
```diff
@@ -138,11 +138,11 @@ export class PRNode extends TreeNode implements CommentHandler, vscode.Commentin
 			this._fileChanges = await this.resolveFileChanges();
 
 			if (!this._inMemPRContentProvider) {
-				this._inMemPRContentProvider = getInMemPRContentProvider().registerTextDocumentContentProvider(this.pullRequestModel.prNumber, this.provideDocumentContent.bind(this));
+				this._inMemPRContentProvider = getInMemPRContentProvider().registerTextDocumentContentProvider(this.pullRequestModel.number, this.provideDocumentContent.bind(this));
 			}
 
 			// The review manager will register a document comment's controller, so the node does not need to
-			if (!this.pullRequestModel.equals(this._prManager.activePullRequest)) {
+			if (!this.pullRequestModel.equals(this._prManager.activeItem)) {
```
RMacfarlane: Since this file is also pull request specific, this should use `activePullRequest`

src/github/pullRequestManager.ts
alexr00: Changed to have `activeIssue` and `activePullRequest`

src/github/interface.ts
```diff
@@ -83,15 +83,18 @@ export interface PullRequest {
 	assignee?: IAccount;
 	createdAt: string;
 	updatedAt: string;
-	head?: IGitHubRef;
-	base?: IGitHubRef;
 	user: IAccount;
 	labels: ILabel[];
+	isDraft?: boolean;
```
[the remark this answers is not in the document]
alexr00: Moved!

src/github/issueModel.ts
```diff
@@ -0,0 +1,128 @@
...
+				};
+
+				return uri;
+			}
+		}
+
+		return undefined;
+	}
+
+	public get inDraftMode(): boolean {
```
RMacfarlane: Sorry, I missed leaving a comment on this earlier 😱 

I think this should be moved to the PullRequestModel since only pull requests have a notion of being in draft mode. (Draft mode is overloaded, in this case it means in the middle or writing a PR review.)

alexr00: Moved all draft mode members into PullRequestModel.

alexr00:
Thank you for reviewing!

PR #1485, opened 2020-02-04
Add the initial test issue features

alexr00:
Adds some issue related features to try out.

### DIFF
--- a/src/issues/issueFeatureRegistrar.ts
+++ b/src/issues/issueFeatureRegistrar.ts
@@ -1,0 +1,70 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { PullRequestManager } from '../github/pullRequestManager';
+import * as vscode from 'vscode';
+import { NewIssue } from './issueTodoProvider';
+import { IssueHoverProvider } from './issueHoverProvider';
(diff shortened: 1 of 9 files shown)


alexr00:
@RMacfarlane this PR doesn't touch much existing code but I definitely wanted to get your input on the package.json changes. I think I should wait until the VS Code insiders version is 1.43, then update the version here before merging so that it ends up in the nightly build only for insiders.

src/issues/issueFeatureRegistrar.ts
```diff
@@ -0,0 +1,70 @@
...
+			titlePlaceholder = newIssue.line.substring(insertIndex + 4, newIssue.line.length).trim();
+			range = newIssue.range;
+		} else {
+			return undefined;
+		}
+
+		const title = await vscode.window.showInputBox({ value: titlePlaceholder, prompt: 'Issue title' });
+		if (title) {
+			let codePermalink: string | undefined;
+			if (this.manager.repository.state.HEAD && this.manager.repository.state.HEAD.commit) {
```
RMacfarlane: Not sure what the best way to handle this case is, but if I have a local HEAD commit that hasn't been pushed to the remote and create a TODO issue, the link will 404. There's an `ahead` property on `HEAD` that could be used to check this

alexr00: Good catch! I'm adding a fallback to just use a link, not a permalink in this case. You could still hit an issue where the file doesn't exist on the remote, but it should be less common.

alexr00:
Will merge once we have a new insiders version 1.43 build.

PR #1490, opened 2020-02-11
Add a completion provider for issues

alexr00:
Works in git SCM input and editor. In markdown, it inserts a markdown link

### DIFF
--- a/src/github/interface.ts
+++ b/src/github/interface.ts
@@ -47,6 +47,8 @@ export interface ISuggestedReviewer extends IAccount {
 export interface IMilestone {
 	title: string;
 	dueOn?: string | null;
+	createdAt?: string;
+	id?: string;
 }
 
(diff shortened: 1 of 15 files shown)


alexr00:
@RMacfarlane feel free to review as much or as little as you want. Any feedback is always welcome!

src/github/interface.ts
```diff
@@ -47,6 +47,8 @@ export interface ISuggestedReviewer extends IAccount {
 export interface IMilestone {
 	title: string;
 	dueOn?: string | null;
+	createdAt?: string;
```
RMacfarlane: shouldn't a milestone always have a `createdAt` and `id` field?

alexr00: I just wasn't getting them on all issues. Updated to include them.

PR #1542, opened 2020-03-08
Add a button to fetch when viewing partial content

### LINKED ISSUE #1462: Add a fetch button to the "local repository out of date" message box
Related to #1288

It feels like the message box that shows "Your local repository is not up to date so only partial content is being displayed" could be improved to add a button which does a fetch on the base branch and reloads the document.

I would be willing to try and contribute a fix for this if the team is open to this.

RikkiGibson:
Resolves #1462

This currently works by running a 'git fetch' and then reloading the whole window. I would appreciate some guidance on how to refresh things in a more granular way.

questions I have:
- do comment caches need to be cleared after the fetch? I thought maybe they do if the comments are associated with the line numbers of the partial content.
- do we need to re-resolve all file changes for the current PR? or maybe for all loaded PRs with the same base branch? It looks like the file change information is written right into the tree nodes, so it's difficult to refresh it without recreating the tree and maybe the associated diff views. but I wasn't 100% sure of what I was looking at.
- haven't looked into automated testing yet, should I add some tests for this change?

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -153,6 +153,7 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: Pu
 
 	context.subscriptions.push(vscode.commands.registerCommand('pr.openDiffView', async (fileChangeNode: GitFileChangeNode | InMemFileChangeNode) => {
 		const GIT_FETCH_COMMAND = 'Run \'git fetch\'';
+		const TITLE = "GitHub Pull Requests";
 
 		const parentFilePath = fileChangeNode.parentFilePath;
 		const filePath = fileChangeNode.filePath;
@@ -177,13 +178,12 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: Pu
 		if (isPartial) {
 			const selection = await vscode.window.showInformationMessage('Your local repository is not up to date. Fetch the PR base branch to show full content.', GIT_FETCH_COMMAND);
 			if (selection === GIT_FETCH_COMMAND) {
-				const { remote: { remoteName }, base: { ref } } = fileChangeNode.pullRequest;
 				const prNode = getPRNode();
 				await vscode.window.withProgress({
 					location: vscode.ProgressLocation.Notification,
-					title: `Running 'git fetch ${remoteName} ${ref}'`,
+					title: TITLE,
 					cancellable: false
-				}, progress => prNode.fetchBaseBranchAndReload(prManager.repository, remoteName, ref, progress));
+				}, progress => prNode.fetchBaseBranchAndReload(progress));
 
 				function getPRNode(): PRNode {
 					let parent: TreeNode | undefined;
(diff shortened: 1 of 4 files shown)


RMacfarlane:
Hey, thanks for creating this PR! Sorry for the delay in reviewing it!

Yes, both comments and the file content would need to be refreshed when this is done. For both of these, the data is stored at the tree node level, in `pullRequestNode.ts`. The comment cache is through the comment controller there. The file content that actually gets displayed is from the `provideDocumentContent` method.

From the command here, I think it would be possible to traverse up from the `fileChangeNode` to the PR node by following the parent, so one possibility is to do that and then add a method for specifically clearing the comment cache and rehydrating it with `refreshExistingPREditors`. In this scenario, an editor is already open with the old file content, so I think we also need to fire the `onDidChange` event for the TextDocumentContentProvider to refresh the editor content. (This event is currently not exposed, but is private on the `_inMemContentProvider` on the PR node.) Another potential solution is just calling refresh on the tree, which should handle resetting the comments and file contents, but would still need some additional event to refresh the editors that are already open.

There are some existing integration tests in `prsTree.test.ts`, but I don't think we have anything that checks the comment cache/file content on file open. They're pretty cumbersome to write because there's a high degree of mocking, both local git and responses from github, so I wouldn't worry about them

RikkiGibson:
Thanks for the detailed guidance! I will take a shot at gracefully refreshing the content when I get the chance, sometime in the next couple of weeks.

RikkiGibson:
Hi @RMacfarlane, I think I have figured out how to refresh the PR content pretty gracefully.

Some documents take a while to finish updating after the onChange event is kicked off, so I wanted there to be a progress notification while that was going on. I found it was necessary to do some more complex stuff to listen for the documents to finish updating. See `allEditorsUpdatedPromise`. I would like to know if there's a better/simpler way of doing what I'm trying to do.

Here's a demo video of the feature: https://i.imgur.com/nL6BTeC.mp4

RikkiGibson:
Here, I recorded a better demo :smile:

demo-fetch

RikkiGibson:
Hi @RMacfarlane, please let me know if there's anything I can do to help drive this PR :)

src/commands.ts
```diff
@@ -175,6 +175,28 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: Pu
 		}
 
 		vscode.commands.executeCommand('vscode.diff', parentURI, headURI, fileName, opts);
+		if (isPartial) {
```
RMacfarlane, line 178: I think this should still be above the `vscode.diff` command

RikkiGibson: Sounds good. I changed 'await' to '.then()' to make sure the diff still comes up when the user hasn't selected an option in the information message yet.

RikkiGibson:
@RMacfarlane Thanks for the review! I have addressed your feedback. Sorry about the lint warnings, they must have snuck in last minute. :smile:

RMacfarlane:
Thank you!

PR #1570, opened 2020-03-12
README - sign into privately hosted Github repo's

j-potts:
Just thought I'd add a note about the ability to sign in using a personal authentication token, as I'd imagine a lot of people using this extension are using it in the workplace!


alexr00:
@j-potts thanks for the PR! This year we expanded support for GitHub Enterprise and removed the need for a PAT.

PR #1626, opened 2020-04-14
Adopt VSCode Authentication Provider API

RMacfarlane:
Lots to delete here :)

Replaces old authentication code with `vscode.authentication` API. As part of this, I have also removed the GitHub Enterprise code. We weren't actively supporting it since we could not test it, but now we explicitly do not support it since we cannot auth against it.


RMacfarlane:
Should also remove the statusbar item as part of this.

src/authentication/githubServer.ts
```diff
@@ -2,36 +2,11 @@ import * as https from 'https';
...
-const SCOPES: string = 'read:user user:email repo write:discussion';
-const GHE_OPTIONAL_SCOPES: { [key: string]: boolean } = { 'write:discussion': true };
-
-const AUTH_RELAY_SERVER = 'vscode-auth.github.com';
+import { HostHelper } from './configuration';
 
 export class GitHubManager {
 	private _servers: Map<string, boolean> = new Map().set('github.com', true);
 
-	private static GitHubScopesTable: { [key: string]: string[] } = {
```
alexr00, line 20: There are a lot more scopes removed here than are added in github/credentials.ts. Are they just not needed?

src/github/credentials.ts
```diff
@@ -76,24 +52,16 @@ export class CredentialStore implements vscode.Disposable {
...
+		const existingSessions = await vscode.authentication.getSessions(AUTH_PROVIDER_ID, SCOPES);
 
-		if (token) {
-			if (await server.validate(token)) {
-				octokit = await this.createHub({ host, token });
-			} else {
-				Logger.debug(`Token is no longer valid for host ${host}.`, 'Authentication');
-			}
+		if (existingSessions.length) {
+			const token = await existingSessions[0].getAccessToken();
```
alexr00, line 58: Why is it always the session at 0?

src/github/enterprise.gql
```diff
@@ -1,609 +0,0 @@
-# /*---------------------------------------------------------------------------------------------
```
alexr00, line 1: Not supporting this anymore?

src/authentication/githubServer.ts
RMacfarlane: Yeah, these were equivalences between different scopes, used to validate that a token had the correct permissions. Since the auth provider should always return a token with the exact scopes we requested, this isn't necessary. (Before, a user could provide their own token, and the GHE token also had a slightly different list of scopes)

src/github/credentials.ts
RMacfarlane: This was under the assumption that all sessions from the github auth provider would be for the same account. Signing into multiple accounts with the provider is challenging, because github only allows you to be signed into one account at at time in the browser. so I guess theoretically you could sign in, then logout in the browser and trigger sign in again from vscode

perhaps I should show a quickpick if there are different accounts available

src/github/enterprise.gql
RMacfarlane: yup!

PR #1632, opened 2020-04-15
Issues: Custom query not scoped to current repository

### LINKED ISSUE #1587: Issues: Custom query not scoped to current repository
I've added a custom query to the Issues View
```
		{
			"label": "Recent Issues",
			"query": "is:issue is:open "
		}
```
I would have expected that the query is scoped to the current repository. However, the result of the query is not scoped to the current repository but it appears to be all repositories:

image

I can scope the query to a particular repo using the `repo:` attribute, but then I cannot write a query for 'the recent issues in this repository'.

alexr00:
Fixes #1587

### DIFF
--- a/src/issues/stateManager.ts
+++ b/src/issues/stateManager.ts
@@ -10,7 +10,7 @@ import { IAccount } from '../github/interface';
 import { PullRequestManager, PRManagerState, NO_MILESTONE, PullRequestDefaults } from '../github/pullRequestManager';
 import { MilestoneModel } from '../github/milestoneModel';
 import { API as GitAPI, GitExtension } from '../typings/git';
-import { ISSUES_CONFIGURATION, QUERIES_CONFIGURATION, DEFAULT_QUERY_CONFIGURATION, BRANCH_CONFIGURATION } from './util';
+import { ISSUES_CONFIGURATION, BRANCH_CONFIGURATION, QUERIES_CONFIGURATION, DEFAULT_QUERY_CONFIGURATION, variableSubstitution } from './util';
 import { CurrentIssue } from './currentIssue';
 
 // TODO: make exclude from date words configurable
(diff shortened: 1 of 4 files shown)


src/issues/util.ts
```diff
@@ -246,12 +246,14 @@ export async function createGithubPermalink(manager: PullRequestManager, positio
...
 
 const VARIABLE_PATTERN = /\$\{(.*?)\}/g;
-export async function variableSubstitution(value: string, issueModel: IssueModel, user?: string): Promise<string> {
+export async function variableSubstitution(value: string, issueModel?: IssueModel, defaults?: PullRequestDefaults, user?: string): Promise<string> {
 	return value.replace(VARIABLE_PATTERN, (match: string, variable: string) => {
 		switch (variable) {
-			case 'user': return user ? user : '';
-			case 'issueNumber': return `${issueModel.number}`;
-			case 'issueTitle': return issueModel.title;
+			case 'user': return user ? user : match;
```
RMacfarlane, line 252: should we include this list of variables in the `githubIssues.queries` setting description?

src/issues/stateManager.ts
```diff
@@ -80,7 +80,7 @@ export class StateManager {
 			this.context.subscriptions.push(repository.state.onDidChange(async () => {
 				if ((repository.state.HEAD ? repository.state.HEAD.commit : undefined) !== this._lastHead) {
 					this._lastHead = (repository.state.HEAD ? repository.state.HEAD.commit : undefined);
-					this.setIssueData();
+					await this.setIssueData();
 				}
 
 				const newBranch = repository.state.HEAD?.name;
```
RMacfarlane, line 86: sorry, not directly related to this PR - but should this also be inside the check above? Seems like it would only be necessary to update this when HEAD changes

src/issues/util.ts
alexr00: Added!

src/issues/stateManager.ts
alexr00: You can have a branch change that doesn't result in the head commit being changed, so I would leave this out of the check above since it needs to happen when the branch changes, not head.

PR #1633, opened 2020-04-15
Inline issue action to go to query

### LINKED ISSUE #1622: Support to navigate from an issue query to its definition in the settings.
Provide an action in the Issues view that opens the corresponding query definition in the settings.

alexr00:
Fixes #1622

### DIFF
--- a/src/issues/issueFeatureRegistrar.ts
+++ b/src/issues/issueFeatureRegistrar.ts
@@ -9,7 +9,7 @@ import { IssueHoverProvider } from './issueHoverProvider';
 import { UserHoverProvider } from './userHoverProvider';
 import { IssueTodoProvider } from './issueTodoProvider';
 import { IssueCompletionProvider } from './issueCompletionProvider';
-import { NewIssue, createGithubPermalink, USER_EXPRESSION, ISSUES_CONFIGURATION } from './util';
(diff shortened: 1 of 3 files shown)


src/issues/issueFeatureRegistrar.ts
```diff
@@ -50,6 +51,28 @@ export class IssueFeatureRegistrar implements vscode.Disposable {
...
+		let command: string;
+		if (inspect?.workspaceValue) {
+			command = 'workbench.action.openWorkspaceSettingsFile';
+		} else {
+			command = 'workbench.action.openSettingsJson';
+		}
+		await vscode.commands.executeCommand(command);
+		const editor = vscode.window.activeTextEditor;
+		if (editor) {
+			const text = editor.document.getText();
```
RMacfarlane, line 66: This is really cool! The "Configure Queries" action for changing the pull request queries just opens the settings page with prefilled search text, this seems much better. I'll create an issue to match this behavior with that action

alexr00:
> This is really cool! The "Configure Queries" action for changing the pull request queries just opens the settings page with prefilled search text, this seems much better. I'll create an issue to match this behavior with that action  

In very large settings files it will likely be slow, but since it's not an action I expect people to take very often it seemed worth it!

PR #1657, opened 2020-04-20
Simplify credentials, create only one octokit

RMacfarlane:
Previously, we created octokits for each domain in a folder's github remotes. This allowed separate octokits for github.com vs. different github enterprise endpoints. Since we will now only support github.com with the new authentication API, we can just create one octokit for it.

### DIFF
--- a/src/github/pullRequestManager.ts
+++ b/src/github/pullRequestManager.ts
@@ -176,16 +176,13 @@ export class PullRequestManager implements vscode.Disposable {
 		}
 	}
 
-	// Check if the remotes are authenticated and show a prompt if not, but don't block on user's response
+	// Check if authenticated and show a prompt if not, but don't block on user's response
 	private async showLoginPrompt(): Promise<void> {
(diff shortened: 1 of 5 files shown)


src/github/pullRequestManager.ts
```diff
@@ -441,35 +438,13 @@ export class PullRequestManager implements vscode.Disposable {
...
-			Logger.appendLine(`serverAuthPromises failed: ${e}`);
-		});
+		const isAuthenticated = await this._credentialStore.hasOctokit();
+		vscode.commands.executeCommand('setContext', 'github:authenticated', isAuthenticated);
 
 		const repositories: GitHubRepository[] = [];
 		const resolveRemotePromises: Promise<void>[] = [];
 
-		authenticatedRemotes.forEach(remote => {
+		activeRemotes.forEach(remote => {
```
alexr00, line 447: No need to only do this on authenticated remotes now?

PR #1658, opened 2020-04-20
Remove sign out option from PR gear menu

RMacfarlane:
Missed removing this, it's already gone from the `package.json`. Sign out is now located unded the accounts menu for the account

### DIFF
--- a/src/view/prsTreeDataProvider.ts
+++ b/src/view/prsTreeDataProvider.ts
@@ -60,8 +60,7 @@ export class PullRequestsTreeDataProvider implements vscode.TreeDataProvider<Tre
 		this._childrenDisposables = [];
 
 		this._disposables.push(vscode.commands.registerCommand('pr.configurePRViewlet', async () => {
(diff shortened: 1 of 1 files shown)


IllusionMH:
Almost reported this as an issue.

Is it expected that user would know that now they are supposed to use "unified" Account menu for this?

Because it wasn't obvious at the first glace that command was removed instead of using VS Code to trigger similar operation for unified login (is it available) or at least show notification with instructions.

---

P.S. At least it wasn't clear that "unified" Sign in was used from notifications in process. Probably VS Code message needs to be updated to be more precise about creation of GitHub Account for whole app.

The extension 'GitHub Pull Request Nightly Build' wants to sign in using GitHub.

RMacfarlane:
@IllusionMH Agree that this experience is still confusing. I'd like to make it so that both sign in and sign out happen from the accounts menu, I think that would be much easier to understand. I'm still going to go ahead and merge this instead of adding back the "Sign out" with the intent of working on a more centralized sign in flow

PR #1659, opened 2020-04-21
Enable issue details editing

### LINKED ISSUE #1608: Allow editing of issue details when creating from code action
When I create an issue from a TODO, I don't know that the comment would be exactly what I would want in the issue. It would be nice to get a details screen to add or remove details from the issue before it was submitted. 

Perhaps this is something that could be turned off with a setting in case people don't want the friction.

alexr00:
Fixes #1608

### DIFF
--- a/src/issues/issueFeatureRegistrar.ts
+++ b/src/issues/issueFeatureRegistrar.ts
@@ -75,9 +75,19 @@ export class IssueFeatureRegistrar implements vscode.Disposable {
 			return;
 		}
 		text = vscode.window.activeTextEditor.document.getText();
-		const indexOfEmptyLine = text.indexOf('\n\n');
-		if (indexOfEmptyLine < 0) {
+		const indexOfEmptyLineWindows = text.indexOf('\r\n\r\n');
+		const indexOfEmptyLineOther = text.indexOf('\n\n');
+		let indexOfEmptyLine: number;
+		if (indexOfEmptyLineWindows < 0 && indexOfEmptyLineOther < 0) {
 			return;
+		} else {
+			if (indexOfEmptyLineWindows < 0) {
+				indexOfEmptyLine = indexOfEmptyLineOther;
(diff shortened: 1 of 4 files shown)


src/issues/issueFeatureRegistrar.ts
```diff
@@ -61,6 +66,29 @@ export class IssueFeatureRegistrar implements vscode.Disposable {
...
+		if (!vscode.window.activeTextEditor) {
+			return;
+		}
+		text = vscode.window.activeTextEditor.document.getText();
+		const indexOfEmptyLine = text.indexOf('\n\n');
+		if (indexOfEmptyLine < 0) {
+			return;
+		}
+		const title = text.substring(0, indexOfEmptyLine);
+		const body = text.substring(indexOfEmptyLine + 2);
```
RMacfarlane, line 83: should we trim off the explanation comment from the issue body?

src/issues/issueFeatureRegistrar.ts
```diff
@@ -61,6 +66,29 @@ export class IssueFeatureRegistrar implements vscode.Disposable {
...
+	async createIssueFromFile() {
+		if (this.createIssueInfo === undefined) {
+			return;
+		}
+		let text: string;
+		if (!vscode.window.activeTextEditor) {
+			return;
+		}
+		text = vscode.window.activeTextEditor.document.getText();
+		const indexOfEmptyLine = text.indexOf('\n\n');
```
RMacfarlane, line 78: is it at all possible that the line ending would be changed on save or anything like that?

src/issues/issueFeatureRegistrar.ts
alexr00: I would leave it, similar to how GitHub leaves the comment in the issue templates.

src/issues/issueFeatureRegistrar.ts
alexr00: Yes, added some handling for that.

Issue #1673, opened 2020-04-22
Creating comment failed: HttpError: Not Found


RikkiGibson:
Using the nightly build (2020.4.31142) or master (0aa30ea) gives this error when I hit the "Add Comment" button. The comment view then changes to the "edit" mode with cancel/save buttons.

I tried to get a stack for this but I didn't manage to hit a breakpoint for it. The error in the developer console has a stack with one frame, but it's core workbench code. I wonder if there's something that can be done in general to make it so more useful stacks are available in the developer console when errors happen.

0.15.0 stable creates comments just fine, so this may be a regression.

RMacfarlane:
Do you remember if you had the PR checked out when trying to comment?

RikkiGibson:
I had an unrelated feature branch checked out in my local repo. Is that what you mean?

RMacfarlane:
Yeah, that's what I meant. I'm able to repro this, looking into it now.

roblourens:
Should the fix be in the latest nightly build? I still repro this.

RMacfarlane:
Yeah, it should be. Can you try signing out and in again? I'm wondering if the first-party vs third-party app registration is also making a difference here

roblourens:
That worked

PR #1675, opened 2020-04-23
Create comment as part of creating a pending review

RMacfarlane:
Fixes https://github.com/microsoft/vscode-pull-request-github/issues/1673

I haven't figured out what exactly exposed this timing issue. Currently, when creating a comment and clicking "Start Review", we first create a pending review, and then add a comment to it. The add comment code always queries github to try to get the pending review id, and there seems to be a slight delay between when the pending review is created and when queries return it. So we query, find nothing, and then try to create the comment as not part of a review, which fails.

The start review graphql API supports passing comments to create with the review, so I'm now just making one call to github to start the review.

### DIFF
--- a/src/view/reviewCommentController.ts
+++ b/src/view/reviewCommentController.ts
@@ -860,7 +860,7 @@ export class ReviewCommentController implements vscode.Disposable, CommentHandle
 		});
 	}
 
-	private async  updateWithNewComment(comment: IComment, matchedFile: GitFileChangeNode): Promise<void> {
+	private async updateWithNewComment(comment: IComment, matchedFile: GitFileChangeNode): Promise<void> {
 		matchedFile.update(matchedFile.comments.concat(comment));
 		this._comments.push(comment);
 
--- a/src/view/reviewCommentController.ts
+++ b/src/view/reviewCommentController.ts
@@ -480,7 +480,7 @@ export class ReviewCommentController implements vscode.Disposable, CommentHandle
 
 	// #region Helper
 
-	private async createNewThread(thread: GHPRCommentThread, matchedFile: GitFileChangeNode, text: string): Promise<IComment | undefined> {
+	private async getNewCommentPosition(thread: GHPRCommentThread, matchedFile: GitFileChangeNode): Promise<number> {
 		const uri = thread.uri;
 		let isBase = false;
 		if (uri.query) {
@@ -503,7 +503,7 @@ export class ReviewCommentController implements vscode.Disposable, CommentHandle
 		}
 
 		if (!this._prManager.activePullRequest.validatePullRequestModel('Creating new comment failed')) {
-			return;
+			throw new Error('No upstream branch');
 		}
 
 		const headCommitSha = this._prManager.activePullRequest.head.sha;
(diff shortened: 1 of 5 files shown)


src/view/reviewCommentController.ts
```diff
@@ -826,6 +860,14 @@ export class ReviewCommentController implements vscode.Disposable, CommentHandle
 		});
 	}
 
+	private async  updateWithNewComment(comment: IComment, matchedFile: GitFileChangeNode): Promise<void> {
```
RikkiGibson, line 863:
```suggestion
	private async updateWithNewComment(comment: IComment, matchedFile: GitFileChangeNode): Promise<void> {
```

src/view/treeNodes/pullRequestNode.ts
```diff
@@ -903,9 +907,35 @@ export class PRNode extends TreeNode implements CommentHandler, vscode.Commentin
...
+			);
+
+			await this.updateCommentThreadCache(thread, fileChange, newComment);
+			this.replaceTemporaryComment(thread, newComment, temporaryCommentId);
+			fileChange.update(fileChange.comments.concat(newComment));
+			this.setContextKey(true);
+		} catch (e) {
+			vscode.window.showErrorMessage(`Starting a review failed: ${e}`);
+
+			thread.comments = thread.comments.map(c => {
```
RikkiGibson, line 931: Consider using a for..of loop on the existing array.

src/view/treeNodes/pullRequestNode.ts
```diff
@@ -903,9 +907,35 @@ export class PRNode extends TreeNode implements CommentHandler, vscode.Commentin
 
 	// #region Review
 	public async startReview(thread: GHPRCommentThread, input: string): Promise<void> {
-		await this._prManager.startReview(this.pullRequestModel);
-		await this.createOrReplyComment(thread, input);
-		this.setContextKey(true);
+		const temporaryCommentId = this.optimisticallyAddComment(thread, input, true);
+
+		try {
+
```
RikkiGibson, line 913:
```suggestion
```

src/view/reviewCommentController.ts
```diff
@@ -717,8 +721,38 @@ export class ReviewCommentController implements vscode.Disposable, CommentHandle
...
+			);
+
+			thread.threadId = comment.id.toString();
+			this.addToCommentThreadCache(thread);
+			this.replaceTemporaryComment(thread, comment, temporaryCommentId);
+			await this.updateWithNewComment(comment, matchedFile);
+		} catch (e) {
+			vscode.window.showErrorMessage(`Starting review failed: ${e}`);
+
+			thread.comments = thread.comments.map(c => {
```
RikkiGibson, line 748: Consider using a for..of loop on the existing array.

src/view/treeNodes/pullRequestNode.ts
RMacfarlane: Ah, the assignment is actually necessary here - this triggers a rerender on the VSCode side. So I am going to leave as is. But thanks for taking the time to look at this! 😄

RikkiGibson:
Thanks!

PR #1686, opened 2020-04-23
Publish to GitHub from no repo folder

### LINKED ISSUE #1664: Support to export code to GitHub
There are a few scenarios:

- [x] **Single folder open, no repo.** Show welcome view content and status bar item to export the currently open file to a GitHub repository
- [x] **Single folder open, repo, but no remotes.** Show status bar item to export the currently open file to a GitHub repository.
- [x] **Multi folder open, some repos with no remotes.** Show status bar item to export one of the currently opened folder repos to a GitHub repository.
- [x] **Multi folder open, no repos.** Show status bar item to export one of the currently opened folder repos to a GitHub repository.

joaomoreno:
Related to #1664

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -13,7 +13,7 @@ import { GitFileChangeNode, InMemFileChangeNode } from './view/treeNodes/fileCha
 import { CommitNode } from './view/treeNodes/commitNode';
 import { PRNode } from './view/treeNodes/pullRequestNode';
 import { PullRequest } from './github/interface';
-import { formatError } from './common/utils';
+import { formatError, onceEvent } from './common/utils';
 import { GitChangeType } from './common/file';
 import { getDiffLineByPosition, getZeroBased } from './common/diffPositionMapping';
 import { DiffChangeType } from './common/diffHunk';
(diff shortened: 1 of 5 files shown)


src/commands.ts
```diff
@@ -547,3 +550,162 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: Pu
...
+}
+
+export function registerGlobalCommands(context: vscode.ExtensionContext, gitAPI: GitAPI, credentialStore: CredentialStore) {
+	// TODO: join with other code from https://github.com/microsoft/vscode-pull-request-github/pull/1661
+	async function getHub(): Promise<GitHub | undefined> {
+		if (await credentialStore.hasOctokit()) {
+			return await credentialStore.getHub()!;
+		}
+
+		const hub = await credentialStore.loginWithConfirmation();
```
RMacfarlane, line 565: Ah, I missed this in your other PR, but I think we can simplify this to just calling `credentialStore.login`. Otherwise the behavior is that the user sees a notification asking them to sign in, and if they cancel it, they get a dialog asking  if the extension can sign in. I think we can skip the notification step, this is used just when the extension is starting up

src/commands.ts
```diff
@@ -547,3 +550,155 @@ export function registerCommands(context: vscode.ExtensionContext, prManager: Pu
...
+
+		if (!hub) {
+			return credentialStore.login();
+		}
```
joaomoreno, line 569: @RMacfarlane You mean something like this?

```suggestion
		return credentialStore.login();
```

RMacfarlane: Something like this:

```ts
		if (await credentialStore.hasOctokit()) {
			return await credentialStore.getHub()!;
		} else {
                        return await credentialStore.login();
                }
```

PR #1694, opened 2020-04-24
Support mentionable users into liveshare 'suggested' contacts

rodrigovaras:
Liveshare as a pane to display 'suggested' contacts to allow them an easy click to invite option. The model build around that allow external providers to extend the mechanism on how to suggest contacts and aggregate on multiple providers.
This PR will have the vscode-pr extension to regsiter a provider where it can monitor the calculated mentionable users and push them into liveshare which then will enable presence for them and allow them invite.
The changes involve using the liveshare API to register the new provider and also a way to hookup when mentionable users are finished, then it will convert back into liveshare 'contact' models.
Please note that the IAccount was modified to add the 'email' optional property in order to have liveshare to be able to match the correct liveshare 'live' user (with presence support)


src/extension.ts
```diff
@@ -116,7 +124,10 @@ export async function activate(context: vscode.ExtensionContext): Promise<ApiImp
 	}
 
 	context.subscriptions.push(registerBuiltinGitProvider(apiImpl));
-	context.subscriptions.push(registerLiveShareGitProvider(apiImpl));
+	const liveshareGitProvider = registerLiveShareGitProvider(apiImpl);
+	context.subscriptions.push(liveshareGitProvider);
+	const liveshareApi = liveshareGitProvider.initialize();
```
RMacfarlane, line 129: `await` here?

rodrigovaras: the await (actually a .then()) is done later on th einit method. I will rename the name to 'liveshareApiPromise'

src/gitProviders/PullRequestProvider.ts
```diff
@@ -0,0 +1,121 @@
+import * as vscode from 'vscode';
```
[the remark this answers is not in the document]
rodrigovaras: done

src/gitProviders/PullRequestProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+        parameters: Object,
+        cancellationToken?: vscode.CancellationToken)
+        : Promise<Object>;
+
+    readonly onNotified: vscode.Event<NotifyContactServiceEventArgs>;
+}
+
+interface NotifyContactServiceEventArgs {
+    type: string;
+    body?: any | undefined;
```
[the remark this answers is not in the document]
rodrigovaras: Yes, the interface is intented to notify multiple type of bodies controlled by the 'type' property. I did not want to define the whole list of notifications available. This definitions should have come from oru public npm package (that is auto generated). The problem is that the package was triggering all sort of errors and warnings so i decided to handcraft what we really need

src/gitProviders/PullRequestProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+interface Contact {
+    id: string;
+    displayName?: string | undefined;
+    email?: string | undefined;
+}
+
+/**
+ * A contact service provider for liveshare that would suggest contacts based on the pull request manager
+ */
+export class PullRequestProvider implements ContactServiceProvider {
```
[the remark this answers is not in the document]
rodrigovaras: Done

src/gitProviders/PullRequestProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+    private readonly onNotifiedEmitter = new vscode.EventEmitter<NotifyContactServiceEventArgs>();
+
+    public onNotified: vscode.Event<NotifyContactServiceEventArgs> = this.onNotifiedEmitter.event;
+
+	constructor(private readonly pullRequestManager: PullRequestManager) {
+		pullRequestManager.onDidChangeMentionableUsers(e => {
```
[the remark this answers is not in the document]
rodrigovaras: I check with Jonathan Carter and we both agree to follwo your seuggestion, thanks for the advice

src/gitProviders/PullRequestProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+            this.notifySuggestedAccounts(e);
+		})
```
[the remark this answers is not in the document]
rodrigovaras: done

src/gitProviders/PullRequestProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+    public async requestAsync(
+        type: string,
+        parameters: Object,
```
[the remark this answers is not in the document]
rodrigovaras: done

src/gitProviders/PullRequestProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+							supportsPresence: false,
+							supportsContactPresenceRequest: false,
+							supportsPublishPresence: false
+						}
+                    }
+
+                    // if we get initialized and
+                    const allMentionableUsers = this.pullRequestManager.getAllMentionableUsers();
+                    if (allMentionableUsers) {
+                        setTimeout(() => this.notifySuggestedAccounts(allMentionableUsers), 10);
```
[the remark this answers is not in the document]
rodrigovaras: I did have some concerns that we are sending a  notify without the 'initialize' request to be even returned. But i check again liveshare and it is prepared to receive notifications during the initialize so i remove the timeout

src/gitProviders/GitHubContactServiceProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+						supportsDispose: false,
+						supportsInviteLink: false,
+						supportsPresence: false,
+						supportsContactPresenceRequest: false,
+						supportsPublishPresence: false
+					}
+				};
+
+				// if we get initialized and users are available on the pr manager
+				const allMentionableUsers = this.pullRequestManager.getAllAssignableUsers();
```
RMacfarlane, line 67: nit: `allAssignableUsers`

src/github/pullRequestManager.ts
```diff
@@ -466,6 +472,28 @@ export class PullRequestManager implements vscode.Disposable {
 		});
 	}
 
+	getAllMentionableUsers(): IAccount[] | undefined {
```
RMacfarlane, line 475: would you mind removing this and the corresponding events since it is unused?

src/extension.ts
```diff
@@ -55,6 +57,12 @@ async function init(context: vscode.ExtensionContext, git: ApiImpl, gitAPI: GitA
 	const prManager = new PullRequestManager(repository, telemetry, git, credentialStore);
 	context.subscriptions.push(prManager);
 
+	liveshareApiPromise.then((api) => {
```
RMacfarlane, line 60: prefer `await` syntax over .then

src/gitProviders/GitHubContactServiceProvider.ts
rodrigovaras: done

src/github/pullRequestManager.ts
rodrigovaras: done

src/extension.ts
rodrigovaras: i intentionally changed the 'await' for the 'then', the reason behind is that awaiting the liveshare api means that the liveshare api return would have to wait for the liveshare extension to fully activate. Waiting for the activation would block (awaited fo course) the pr github extension without any good benefit. If you really think that 'await' would be enough i can change the code.

RMacfarlane: Got it, yeah let's not block

RMacfarlane:
Thank you, everything looks good! I've merged it in

PR #1715, opened 2020-04-29
Add back sign out command

RMacfarlane:
Settings sync is going to remain in insiders still, so the account UI will also not be visible. Since this was the only place you could sign out from, I'm now adding back the sign out command to the extension itself.

### DIFF
--- a/package.json
+++ b/package.json
@@ -393,6 +393,11 @@
         "category": "GitHub Pull Requests"
       },
       {
+        "command": "auth.signout",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -392,6 +392,11 @@
         },
         "category": "GitHub Pull Requests"
       },
+      {
+        "command": "auth.signout",
```
alexr00, line 396: This is just a test comment, please ignore it.

PR #1718, opened 2020-04-29
Add inline action for start working

### LINKED ISSUE #1717: Add a hover command for issues in the Issues View
‘Start to work on issue’ is currently somewhat hidden in the context menu of the Issues view. Adding an inline command will improve its discoverability.

image

alexr00:
And a few linting errors
Fixes #1717

### DIFF
--- a/package.json
+++ b/package.json
@@ -642,12 +642,14 @@
       {
         "command": "issue.startWorking",
         "title": "Start Working on Issue",
-        "category": "GitHub Issues"
+        "category": "GitHub Issues",
+        "icon": "$(check)"
       },
       {
         "command": "issue.continueWorking",
         "title": "Continue Working on Issue",
-        "category": "GitHub Issues"
+        "category": "GitHub Issues",
+        "icon": "$(check)"
       },
       {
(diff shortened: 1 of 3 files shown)


alexr00:
@misolori here are the icons I picked for starting and stopping working. What do you think?
image

misolori:
Hmm...the check is often used for "done" or "active" so it feels a bit odd here. You could use the pencil though that could imply editing the issue or title:

image

You could also try the right arrow:
image

But i think in general you are going to have a hard time finding a good representation of "Start working".

For stopping, would the square be better for an actual stop icon? We use that in notebooks for stopping an execution:
image

RMacfarlane:
For "Start working", would the git branch icon make sense?

misolori:
@RMacfarlane we could use that, though it is the same icon as the SCM in the activity bar:

image

alexr00:
New icons:
image

We have the pencil for editing the queries in the same tree, so I think that would be confusing. I also think the scm icon will be confusing since it's already in the activity bar.

PR #1742, opened 2020-05-01
Clean up credential store API

RMacfarlane:


src/github/pullRequestManager.ts
```diff
@@ -180,14 +180,11 @@ export class PullRequestManager implements vscode.Disposable {
...
-			if (!authd) {
-				this._credentialStore.loginWithConfirmation();
-			}
-		});
-
-		return Promise.resolve();
+	private showLoginPrompt(): void {
+		const authd = this._credentialStore.isAuthenticated();
+		if (!authd) {
+			this._credentialStore.showSignInNotification();
```
alexr00, line 186: No longer wating for this?

RMacfarlane: Right, `isAuthenticated` is now syncronous, and we previously didn't want to block on this notification

PR #1760, opened 2020-05-06
Make PRs and Issues views have same commands

### LINKED ISSUE #1744: Make PRs and Issues explorers have same commands
the issues explorer doesn't have the same commands as the PR explorer.

image

- collapse all
- settings gear - could bring up the `"githubIssues.queries"` setting

alexr00:
Fixes #1744

### DIFF
--- a/package.json
+++ b/package.json
@@ -951,7 +951,7 @@
         },
         {
           "command": "pr.configurePRViewlet",
-          "when": "gitOpenRepositoryCount != 0 && github:initialized && view =~ /pr:/",
+          "when": "gitOpenRepositoryCount != 0 && github:initialized && view =~ /(pr|issues):/",
           "group": "navigation"
         },
         {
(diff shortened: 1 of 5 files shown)


alexr00:
@RMacfarlane, while I was doing this I noticed that the Sign in isn't working right now. Steps:
Sign out from User icon near gear
Sign in through "Sign in" in Pull Requests vew
See that the view isn't updated.

package.json
```diff
@@ -955,6 +964,11 @@
           "when": "view =~ /prStatus/",
           "group": "navigation"
         },
+        {
+          "command": "pr.configurePRViewlet",
```
RMacfarlane, line 968: I think we could combine this with the entry above for configurePRViewlet by changing the when clause there

RMacfarlane:
@alexr00 ah, I see the problem - This only happens when logging out and back in again. In the credential store reset method, I always just set the API to undefined, `this.initialize()` should be called after that

RMacfarlane:
I'll push a commit here to fix it

PR #1796, opened 2020-05-12
Support creating issues within editor

### LINKED ISSUE #1713: Support creating issues within extension
Just like you can create a pull request from within the extension, we should also support creating issues from within the extension. Simple two-step process for title and body message.

alexr00:
Fixes #1713


RMacfarlane:
Should we show the flow as creating an issue from a comment, where there quickpick with just "Title" is shown before the markdown file?

Should we close the .md file once the issue is created?

alexr00:
The .md file is closed when created. 

For showing the title quick input: For the other "create issue" entry points there is some content already added to the issue body, so having a way to create the issue without adding more is reasonable. However, in the case we have zero context to add to the issue body. And we all know from experience that receiving an issue without an issue body is usually not what we want. Because of this, I'd rather not offer a way to create an issue without a body and instead skip the quick input step.

PR #1798, opened 2020-05-13
Use codicons in commands

alexr00:
Fixes microsoft/vscode-internalbacklog#1088


alexr00:
Steps I took to make this change:
- Went through every command with an icon.
- Looked at the icon, found the identical codicon, and replaced it.
- Built and verified that the correct icon was still used.
- Searched for icon-name.svg in the project. If not found, I deleted the svg.

Issue #1613, opened 2020-04-10
A list of files displayed in a PR is trimmed (very large PR)


phobos2077:
A PR with 1907 files is loaded and working, but there are many files missing in the extension UI. It seems like there is some kind of hard limit on the number of files loaded from github API and the rest is just discarded. Or some bug in the extension logic?

- Extension version: 0.15.0
- VSCode Version: 1.44.0
- OS: Windows 10 x64

Steps to Reproduce:

1. Open PR with 1907 files in "GitHub Pull Requests".
2. Observe as many files are missing.

MFAshby:
I have observed the same problem. I think the problem is here https://github.com/microsoft/vscode-pull-request-github/blob/b7c553b2676cb6594cbda0c815737314e71d1fe4/src/github/pullRequestManager.ts#L1915

This is only returning 300 changed files maximum.

MFAshby:
I think files should be listed with octokit.pulls.listFiles, I'll submit a PR for this.

PR #1812, opened 2020-05-15
Show all files affected by a PR, not just the top 300

MFAshby:
Uses octokit.pulls.listFiles to fetch the file changes, instead of using the `files` member of `compareCommits`. This API isn't limited to the top 300 files affected.

Note that I'd prefer to use octokit pagination wrapper as described [here](https://github.com/octokit/plugin-paginate-rest.js/#octokitpaginate) instead of a do..while loop, but this doesn't seem to compile in typescript unfortunately.

Addresses https://github.com/microsoft/vscode-pull-request-github/issues/1613

### DIFF
--- a/src/github/pullRequestManager.ts
+++ b/src/github/pullRequestManager.ts
@@ -86,6 +86,7 @@ export class BadUpstreamError extends Error {
 
 export const SETTINGS_NAMESPACE = 'githubPullRequests';
 export const REMOTES_SETTING = 'remotes';
+const MAX_FILE_CHANGES_IN_COMPARE_COMMITS = 300;
 
 interface NewCommentPosition {
 	path: string;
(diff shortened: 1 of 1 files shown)


MFAshby:
I'd appreciate comments regarding how to test this, I'm unsure what kind of test infrastructure is available, `pullRequestManager.test.ts` is somewhat anaemic right now.

src/github/pullRequestManager.ts
```diff
@@ -1904,11 +1904,30 @@ export class PullRequestManager implements vscode.Disposable {
...
+				repo: remote.repositoryName,
+				per_page: 100,
+				page: page
+			});
+			response.data.forEach(i => files.push(i));
+			hasNext = response.headers.link.includes('rel="next"');
+			page++;
+		} while (hasNext);
+
+		// const files = await octokit.paginate(, );
```
RMacfarlane, line 1927: nit: remove comment

src/github/pullRequestManager.ts
```diff
@@ -1904,11 +1904,30 @@ export class PullRequestManager implements vscode.Disposable {
...
+			Logger.debug(`Fetch file changes page ${page} of PR #${pullRequest.number}`, PullRequestManager.ID);
+			const response = await octokit.pulls.listFiles({
+				owner: pullRequest.base.repositoryCloneUrl.owner,
+				pull_number: pullRequest.number,
+				repo: remote.repositoryName,
+				per_page: 100,
+				page: page
+			});
+			response.data.forEach(i => files.push(i));
+			hasNext = response.headers.link.includes('rel="next"');
```
RMacfarlane, line 1923: `link` is sometimes undefined, let's defend against that

src/github/pullRequestManager.ts
```diff
@@ -1904,11 +1904,30 @@ export class PullRequestManager implements vscode.Disposable {
 			base: `${pullRequest.base.repositoryCloneUrl.owner}:${encodeURIComponent(pullRequest.base.ref)}`,
 			head: `${pullRequest.head.repositoryCloneUrl.owner}:${encodeURIComponent(pullRequest.head.ref)}`
 		});
-
 		pullRequest.mergeBase = data.merge_base_commit.sha;
 
+		Logger.debug(`Fetching file changes of PR #${pullRequest.number}`, PullRequestManager.ID);
```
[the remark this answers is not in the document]
MFAshby: Good idea, done

src/github/pullRequestManager.ts
MFAshby: Done

src/github/pullRequestManager.ts
MFAshby: Hopefully this is handled by octokit's paginate function 👍

MFAshby:
I've discovered `// @ts-ignore`, which works around the lack of type information available for octokit's paginate function. It'll be better to type that library correctly, but I wouldn't block this PR on it.

src/github/pullRequestManager.ts
```diff
@@ -1904,11 +1905,30 @@ export class PullRequestManager implements vscode.Disposable {
...
+			Logger.debug(`More than ${MAX_FILE_CHANGES_IN_COMPARE_COMMITS} files changed, fetching all file changes of PR #${pullRequest.number}`, PullRequestManager.ID);
+			// @ts-ignore
+			files = await octokit.paginate(`GET /repos/:owner/:repo/pulls/:pull_number/files`, {
+				owner: pullRequest.base.repositoryCloneUrl.owner,
+				pull_number: pullRequest.number,
+				repo: remote.repositoryName,
+				per_page: 100
+			}, (r) => {
+				Logger.debug(`Fetched another batch of ${r.data.length} file changes for PR #${pullRequest.number}`, PullRequestManager.ID);
+				return r;
```
RMacfarlane, line 1923: I think this should be `r.data`

MFAshby: I don't think it should. Pagination function handles selecting the data it seems, this last optional parameter _can_ be used for transforming the data, but I only wanted to log some debug info so I could be happy it was respecting the batch size I specified. 

Anyway, it works as expected returning r and not r.data.

MFAshby: On second thought, I just removed the logging if that's ok.
I don't think there's any great reason to keep it, I just put it in to prove the pagination was happening during testing.

MFAshby:
Thanks @RMacfarlane for review, I wonder if you could re-check the latest changes if you have a moment? 

I'm keen to get this bug fixed as I find myself reviewing PRs with many files a lot on my current project, and I really like VS-code's deep integration with GitHub PR review 😄

RMacfarlane:
Hey, thanks for the reminder! I've just tried running it locally again, and I'm still seeing that the return of paginate is a structured object with `data` instead of just the data. I did a clean build, and changed the `MAX_FILE_CHANGES_IN_COMPARE_COMMITS` to 1 so that it would always be hit. You're still seeing that returning only `r` works correctly?

MFAshby:
sorry for the delay, I'm still aware this is outstanding I just haven't had a break from my work projects to do it yet 😂 
I plan to debug to confirm what RMacfarlane has observed and make the suggested change if it's needed.

MFAshby:
Closing as @pakrym has opened a new one based on this with conflicts resolved. Thanks! 

https://github.com/microsoft/vscode-pull-request-github/pull/2145

PR #1816, opened 2020-05-18
Permalink should use the upstream repo if possible

### LINKED ISSUE #1781: Permalink should use the upstream repo if possible
I copied the permalink to paste into github but it ended up using my fork which doesn't show the preview, it should have been xtermjs/xterm.js instead:

image

I guess it should do this naively if the latest commit _for the file_ exists in upstream?

alexr00:
Fixes #1781
This is done by convention now. If there are requests to make the convention configurable, then I will make it configurable!


marcuslyons:
I apologize for this but I accidentally "approved" this while testing the plugin locally.

PR #1817, opened 2020-05-18
Remove core github functionality

joaomoreno:
We're moving this to VS Code core.


package.json
```diff
@@ -15,7 +15,7 @@
   "version": "0.16.0",
   "publisher": "GitHub",
   "engines": {
-    "vscode": "^1.45.0"
+    "vscode": "^1.46.0"
```
joaomoreno, line 18: I've bumped this so we don't get clashing commands/welcome views

Issue #1156, opened 2019-05-15
Suggested description page styling tweaks


RMacfarlane:
Found while testing https://github.com/microsoft/vscode-pull-request-github/issues/1152

- While the page is loading, the text "Loading..." is displayed at the very top of the page with no padding. It would be nice to have a top margin for this, or possibly make it more central
- The comment editor seems to have lost its padding between the text area and buttons:
<img width="849" alt="Screen Shot 2019-05-14 at 11 03 57 AM" src="https://user-images.githubusercontent.com/3672607/57740076-e0b3c700-766a-11e9-8095-962bb682da39.png">

Ashniu123:
Can I try this?

RMacfarlane:
@Adam13531 Yes! The second bullet point has already been fixed. The code that adds the "Loading..." text is here: https://github.com/microsoft/vscode-pull-request-github/blob/d8088a6473db83738e477ec116b3bf9041dcb13c/preview-src/app.tsx#L22 and the stylesheet is `preview-src/index.css`

Adam13531:
> @Adam13531 Yes!

I think you may have tagged the wrong person. @Ashniu123, see the last response.

Ashniu123:
Should I just put an `margin-top` like  
```
<div margin-top="10px">Loading...</div>
```
? Or Add use the "title" class ?

JeffreyCA:
It would be helpful to post some screenshots so we can visually compare the two

marcuslyons:
Would something like this work, or is this too far?

### Before - no spacing: 
![before-spacing-change](https://user-images.githubusercontent.com/10541922/81375737-61ab7000-90bf-11ea-89a0-008f7745d3af.gif)

### After - centered:
![after-spacing-changes](https://user-images.githubusercontent.com/10541922/81375788-7f78d500-90bf-11ea-8cfa-d47f18372412.gif)

rishab-pdx:
@queerviolet  
if this is still open and available? I would like to take it. I have centered the "Loading..." text and increased the font size so that it is more visible.

![#1156](https://user-images.githubusercontent.com/7886919/82716481-79046480-9c4c-11ea-83cd-06aa5b7e9d3a.gif)

Ashniu123:
@rishab-pdx if your PR is ready then by all means... Go ahead! :D

rishab-pdx:
@Ashniu123 that makes sense. Here is the final view.

[...]

PR #1829, opened 2020-05-24
Centered the  "Loading..." text.

rishab-pdx:
Tweaks for "Loading" text as per First bullet point of #1156 

wrapped "Loading..." in a div and added a class to make it central.


rishab-pdx:
let me know if further tweaks/changes are required.

PR #1830, opened 2020-05-24
Resolves ssh config issue resolving host

### LINKED ISSUE #1541: Support for ssh config hosts is broken
Issue Type: <b>Bug</b>

- I have an ssh config setup with two hosts, for two accounts:
```
 Host github.com
    User git  
    Hostname github.com
    PreferredAuthentications publickey  
    IdentityFile <A_FILE>

Host work    
    User git  
    Hostname github.com  
    PreferredAuthentications publickey  
    IdentitiesOnly yes  
    IdentityFile <ANOTHER_FILE>
```
- When I install the extension no login option appears, as the tree view says I need to configure my remotes.    
- `git remote -v` does list origin, which maps to `git@work:<repo_url>.git` (a private repo)    
- The output for the extension reveals it's trying to load `https://work/<repo_url>` which of course doesn't exist.
- On creating a new remote mapping directly to the github.com host works, as a sign in option shows up on the tree view, and all is good.
- Looks related to the change [here](https://github.com/microsoft/vscode-pull-request-github/commit/59ffe49f0cdaa259a1b9f01a24f2b500dd7469da), and could be very similar to #414 

Extension version: 0.14.0
VS Code version: Code 1.42.1 (c47d83b293181d9be64f27ff093689e8e7aed054, 2020-02-11T14:44:27.652Z)
OS version: Darwin x64 18.7.0

Rusty-1:
# SSH Config issues

## Background

When configuring a ssh host in the `.ssh/config` with a entry to resolve github.com e.g.)

```
Host MY-GITHUB-ALIAS
   Hostname github.com
   IdentityFile my_authentication_identity
``` 
The extension fails to resolve the host name and displays the following error in the log

```
[Info] GitHubServer> No response from host https://MY-GITHUB-ALIAS/xxxxx/repo: getaddrinfo ENOTFOUND MY-GITHUB-ALIAS
```

## Root cause

The ssh config file was ignoring the `Hostname` property in favour of `HostName` in `protocol.ts`. The man pages for ssh config define the entry a `Hostname` and the result was the wrong value was being assigned to host in `parseSshProtocol()`.

## Resolution

Align with the ssh config specification as defined by

- [SSH_CONFIG(5) ](http://man7.org/linux/man-pages/man5/ssh_config.5.html)

To avoid configuration brittleness `Hostname` or `HostName` will now resolve correctly.

## Related Issues

- Fixes #1541

### DIFF
--- a/src/test/common/protocol.test.ts
+++ b/src/test/common/protocol.test.ts
@@ -10,7 +10,7 @@ import { Protocol, ProtocolType } from '../../common/protocol';
 const SSH_CONFIG_WITH_HOST_ALIASES = `
 Host gh
   User git
-  HostName github.com
+  Hostname github.com
 `;
 
 const str = (x: any) => JSON.stringify(x);
--- a/src/test/common/protocol.test.ts
+++ b/src/test/common/protocol.test.ts
@@ -8,9 +8,13 @@ import * as ssh from '../../common/ssh';
 import { Protocol, ProtocolType } from '../../common/protocol';
 
 const SSH_CONFIG_WITH_HOST_ALIASES = `
-Host gh
+Host gh_nocap
   User git
   Hostname github.com
+
+Host gh_cap
+  User git
+  HostName github.com
 `;
 
 const str = (x: any) => JSON.stringify(x);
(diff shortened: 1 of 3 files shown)


src/test/common/protocol.test.ts
```diff
@@ -10,7 +10,7 @@ import { Protocol, ProtocolType } from '../../common/protocol';
 const SSH_CONFIG_WITH_HOST_ALIASES = `
 Host gh
   User git
-  HostName github.com
+  Hostname github.com
```
RMacfarlane, line 13: Instead of changing this, could you add an additional test duplicating the last test here but using a config with `Hostname` instead of `HostName` so that we test both?

Rusty-1: Thank you for the feedback Rachel I have implemented your suggestion of a second test.

Rusty-1:
Thank you for the review Rachel!

PR #1831, opened 2020-05-26
Refresh the PR tree when PR is approved

### LINKED ISSUE #1749: Consider refreshing the PR view when a pull request is approved
If I'm going through the "Waiting for my review" query and approving PRs, it would be nice if as I approve them they vanish from that query (which they will do when I click the button to refresh the view).

rishab-pdx:
Fixes #1749 
Made a call to ```pr.refreshList``` so that PR tree is updated.


src/github/pullRequestOverview.ts
```diff
@@ -585,6 +585,8 @@ export class PullRequestOverviewPanel extends IssueOverviewPanel {
 				review: review,
 				reviewers: this._existingReviewers
 			});
+			//just making sure that pr review status is approved on git servers
```
RMacfarlane, line 588: I think this timeout isn't necessary, I believe GitHub will return the correct information immediately. Let me know if you've seen different behavior

rishab-pdx: When I was testing I could see some inconsistency with refreshing the list/view.  After I added ```timeout``` there wasn't in any inconsistency.

rishab-pdx: Let me know if you want me to remove the ```timeout```

rishab-pdx:
> Thank you, this looks good! I think the timeout can be removed


When I was testing I could see some inconsistency with refreshing the list/view. After I added timeout there wasn't in any inconsistency. Let me know if you want me to remove timeout

src/github/pullRequestOverview.ts
rishab-pdx: I have removed the ```timeout```.

PR #1834, opened 2020-05-27
Use token API to check if token is comment

alexr00:
Adopt for hovers, completions, and links

Re-enable links


alexeyr:
Will this trigger completion only inside a comment? I was just about to post that as an issue.

alexr00:
@alexeyr that's exactly what this will do (other than in markdown and plaintext).

alexeyr:
Great, looking forward to it (and not posting the issue then)! Very annoying to have it show up in `#include` :)

PR #1849, opened 2020-06-03
Create issue shouldn't create new labels

alexr00:
Fixes microsoft/vscode-internalbacklog#1135

### DIFF
--- a/src/issues/issueFeatureRegistrar.ts
+++ b/src/issues/issueFeatureRegistrar.ts
@@ -21,6 +21,7 @@ import { Resource } from '../common/resources';
 import { IssueFileSystemProvider } from './issueFile';
 import { ITelemetry } from '../common/telemetry';
 import { IssueLinkProvider } from './issueLinkProvider';
+import Octokit = require('@octokit/rest');
 
 const ISSUE_COMPLETIONS_CONFIGURATION = 'issueCompletions.enabled';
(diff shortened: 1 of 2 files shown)


src/issues/issueFeatureRegistrar.ts
```diff
@@ -481,6 +482,15 @@ ${body ?? ''}\n
 		editor.setDecorations(labelsDecoration, [new vscode.Range(new vscode.Position(3, 0), new vscode.Position(3, labelLine.length))]);
 	}
 
+	private async verifyLabels(createParams: Octokit.IssuesCreateParams) {
+		if (!createParams.labels) {
+			return;
+		}
+		const allLabels = await this.manager.getLabels(undefined, createParams);
+		const filteredLabels = allLabels.filter(label => createParams.labels?.includes(label.name)).map(label => label.name);
```
RMacfarlane, line 490: I think it would be good to give some kind of feedback when a label isn't found. Maybe we could show a confirmation dialog showing the label with options "Proceed anyway", "Create label", and "Cancel"?

RMacfarlane: If we do decide to do that, I think it would also be good to verify assignees and similarly show a dialog if validation fails

alexr00:
> If we do decide to do that, I think it would also be good to verify assignees and similarly show a dialog if validation fails 

Good point, I've made this issue to track https://github.com/microsoft/vscode-pull-request-github/issues/1855

PR #1853, opened 2020-06-03
Duplicated placeholder text when pressing new issue many times

alexr00:
Fixes microsoft/vscode-internalbacklog#1138

### DIFF
--- a/src/issues/issueFeatureRegistrar.ts
+++ b/src/issues/issueFeatureRegistrar.ts
@@ -465,7 +465,7 @@ export class IssueFeatureRegistrar implements vscode.Disposable {
 
 	private async makeNewIssueFile(title?: string, body?: string, assignees?: string[] | undefined) {
 		const bodyPath = vscode.Uri.parse(`${NEW_ISSUE_SCHEME}:/NewIssue.md`);
-		if (vscode.window.visibleTextEditors.filter(editor => editor.document.uri.scheme === NEW_ISSUE_SCHEME).length > 0) {
+		if (vscode.window.visibleTextEditors.filter(visibleEditor => visibleEditor.document.uri.scheme === NEW_ISSUE_SCHEME).length > 0) {
 			return;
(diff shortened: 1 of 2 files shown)


src/issues/issueFeatureRegistrar.ts
```diff
@@ -465,6 +465,10 @@ export class IssueFeatureRegistrar implements vscode.Disposable {
 
 	private async makeNewIssueFile(title?: string, body?: string, assignees?: string[] | undefined) {
 		const bodyPath = vscode.Uri.parse(`${NEW_ISSUE_SCHEME}:/NewIssue.md`);
+		if (vscode.window.visibleTextEditors.filter(editor => editor.document.uri.scheme === NEW_ISSUE_SCHEME).length > 0) {
```
RMacfarlane, line 468: tslint warning about shadowed name 'editor' here

src/issues/issueFeatureRegistrar.ts
```diff
@@ -465,6 +465,10 @@ export class IssueFeatureRegistrar implements vscode.Disposable {
 
 	private async makeNewIssueFile(title?: string, body?: string, assignees?: string[] | undefined) {
 		const bodyPath = vscode.Uri.parse(`${NEW_ISSUE_SCHEME}:/NewIssue.md`);
+		if (vscode.window.visibleTextEditors.filter(editor => editor.document.uri.scheme === NEW_ISSUE_SCHEME).length > 0) {
+			return;
+		}
+		await vscode.workspace.fs.delete(bodyPath);
```
RMacfarlane, line 471: not sure I understand this line, why is this needed?

alexr00: It seemed like the easiest way to reset the file and get rid of the decorations. If the file is still open in a tab that isn't currently visible, the decorations will still be there and can end up in the wrong place.

PR #1858, opened 2020-06-04
`getTokenInformationAtPosition` API causes too much traffic and blocks the renderer process

alexr00:
Part of microsoft/vscode#99356


alexr00:
More context https://github.com/microsoft/vscode/issues/99356#issuecomment-638843051

alexr00:
We will remove the link provider again this iteration, and have a better solution next iteration.

PR #1914, opened 2020-06-16
Node-fetch prevents us from minifying

### LINKED ISSUE #1912: Node-fetch prevents us from webpacking
This code exists in the version of node-fetch we use:
```javascript
function isAbortSignal(signal) {
	const proto = signal && typeof signal === 'object' && Object.getPrototypeOf(signal);
	return !!(proto && proto.constructor.name === 'AbortSignal');
}
```
But after webpacking, `proto.constructor.name` is absolutely not `AbortSignal` (it was `i` when I checked).
This causes the issue fetching to error out in an unexpected way.

alexr00:
Fixes #1912

### DIFF
--- a/package.json
+++ b/package.json
@@ -1212,7 +1212,7 @@
     "marked": "^0.8.0",
     "moment": "^2.22.1",
     "node-emoji": "^1.8.1",
-    "node-fetch": "^2.1.1",
+    "node-fetch": "v3.0.0-beta.7",
     "query-string": "^6.2.0",
     "ssh-config": "^2.0.0-alpha.3",
     "tunnel": "0.0.6",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1212,7 +1212,7 @@
     "marked": "^0.8.0",
     "moment": "^2.22.1",
     "node-emoji": "^1.8.1",
-    "node-fetch": "^2.1.1",
+    "node-fetch": "v3.0.0-beta.7",
```
alexr00, line 1215: I had to go for the specifc version since doing ^ caused it to pull `3.0.0-beta.6-exportfix`, which doesn't work for other reasons.

PR #1921, opened 2020-06-18
Sign-in state should use the new Welcome Views

### LINKED ISSUE #1699: Signin state could be more prominent and use the new Welcome Views
Testing https://github.com/microsoft/vscode/issues/96313

The PR View is currently showing a text label to signin:

image

This could be made more prominent using a welcome view shown in this context, similar as the explorer is using in the empty state:

image

alexr00:
Fixes #1699

### DIFF
--- a/package.json
+++ b/package.json
@@ -673,6 +673,18 @@
         "category": "GitHub Issues"
       }
     ],
+    "viewsWelcome": [
+      {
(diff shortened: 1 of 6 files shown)


alexr00:
What it looks like with this change:
image


"Loading..." is unchanged

PR #1928, opened 2020-06-22
Adopt generalized rich hover support for issues in the Issues View

### LINKED ISSUE #1905: Adopt generalized rich hover support for issues in the Issues View
Once https://github.com/microsoft/vscode/issues/97496 becomes available, the rich hover support should be adopted for issues in the issues view.

alexr00:
Fixes #1905


src/typings/vscode.proposed.d.ts
```diff
@@ -0,0 +1,2104 @@
+/*---------------------------------------------------------------------------------------------
```
RMacfarlane, line 1: I don't think we need to add this file, running `yarn` should pull the latest typings from `master` automatically. Is that not working?

alexr00: You're right, removed!

PR #1993, opened 2020-07-13
Update to latest auth provider API

RMacfarlane:


src/extension.ts
```diff
@@ -40,7 +40,7 @@ async function init(context: vscode.ExtensionContext, git: ApiImpl, gitAPI: GitA
 	Logger.appendLine('Git repository found, initializing review manager and pr tree view.');
 
 	vscode.authentication.onDidChangeSessions(async e => {
-		if (e['github']) {
+		if (e.provider.id === 'github') {
```
alexr00, line 43: Do you also need to update the VS Code verion in the package.json?

RMacfarlane: Yes, thank you

PR #2009, opened 2020-07-17
Create a fork when user has no commit rights

### LINKED ISSUE #1903: Work on Issue: Support to create a fork when user has no commit rights
A common OSS scenario is that someone from the community with no commit rights to a repository would like to contribute a fix to an issue. To support this scenario the 'Work on Issue' support should offer to create a fork when the user has no commit rights.

alexr00:
Fixes #1903


alexr00:
@RMacfarlane, I tried to set things up so that this can be added into the create PR flow too. Let me know if you want me to add it in there too!

PR #2022, opened 2020-07-28
Changes needed in PR manager and tree

alexr00:


alexr00:
@RMacfarlane here is part one of the transition to multi-repo:

- I started with a rename, and did that in branch alexr00/multifolder, so that you wouldn't have to sort through as much rename (I hope).
- Have a PulllRequestManager for each repo (named FolderPullRequestManager)
- Have collection of FolderPullRequestManagers (named PullRequestManager)
- Updates to the tree
- Tests still pass

Coming soon:

- Changes to review manager. I expect that to be one review manager for each repo, with a collection of them similar to how the pull request manager go restructured.
- Changes to issue features

My plan is to merge all changes with review into alexr00/multifolder, then merge that branch into master without review.

src/github/pullRequestManager.ts
```diff
@@ -0,0 +1,183 @@
...
+
+export interface PullRequestDefaults {
+	owner: string;
+	repo: string;
+	base: string;
+}
+
+export const NO_MILESTONE: string = 'No Milestone';
+
+export class PullRequestManager implements vscode.Disposable {
```
RMacfarlane, line 75: Sorry, this is only tangentially related to these changes, but while we're reorganizing - do you think this would make more sense named something like `RepositoriesManager`? And the other one `FolderRepositoriesManager`? That seems a bit more accurate and focused to me, I'd eventually like to break up some of the folder one

src/view/prsTreeDataProvider.ts
```diff
@@ -121,7 +120,8 @@ export class PullRequestsTreeDataProvider implements vscode.TreeDataProvider<Tre
 	public async updateQueries() {
 		this._queries = await this.isVSO()
 			? []
-			: vscode.workspace.getConfiguration(SETTINGS_NAMESPACE, this._prManager.repository.rootUri).get<IQueryInfo[]>(QUERIES_SETTING) || [];
+			// TODO: is there a better scope to use?
+			: vscode.workspace.getConfiguration(SETTINGS_NAMESPACE, this._prManager.folderManagers[0]?.repository.rootUri).get<IQueryInfo[]>(QUERIES_SETTING) || [];
```
RMacfarlane, line 124: I'm wondering if we could push the `queries` logic into `WorkspaceFolderNode` so that each one can use its own rootUri

RMacfarlane: I guess this still needs to be here for the single folder case though

src/github/pullRequestManager.ts
alexr00: Yes, absolutely. Since I have a dependent PR, I will merge both into alexr00/multifolder then rename.

src/view/prsTreeDataProvider.ts
alexr00: Moved queries into `WorkspaceFolderNode` and reduced the duplicate code too.

PR #2032, opened 2020-07-30
Get "Changes in Pull Request" view working with multirepo

alexr00:

PR #2036, opened 2020-08-03
Convert issues features over to multirepo

alexr00:


alexr00:
This is the final large change to convert the extension over to a multirepo model. It builds upon https://github.com/microsoft/vscode-pull-request-github/pull/2032.

PR #2109, opened 2020-09-03
Space-delimit commit description lines

### LINKED ISSUE #2107: Replace new line chars with a space when squashing commit description
Squashing on to one line is expected, but I wanted all newlines to be replaced with a space so that it's easier to read.

_Originally posted by @alexr00 in https://github.com/microsoft/vscode-pull-request-github/issues/1943#issuecomment-686332852_

heaths:
Fixes #2107

### DIFF
--- a/src/github/folderRepositoryManager.ts
+++ b/src/github/folderRepositoryManager.ts
@@ -1570,6 +1570,9 @@ export const titleAndBodyFrom = (message: string): { title: string, body: string
 
 		body: idxLineBreak === -1
 			? ''
-			: message.slice(idxLineBreak + 1),
+			: message
+				.slice(idxLineBreak + 1)
+				.trim()
+				.replace(/\n+/g, ' '),
 	};
 };
--- a/src/github/folderRepositoryManager.ts
+++ b/src/github/folderRepositoryManager.ts
@@ -1572,7 +1572,6 @@ export const titleAndBodyFrom = (message: string): { title: string, body: string
 			? ''
 			: message
 				.slice(idxLineBreak + 1)
-				.trim()
-				.replace(/\n+/g, ' '),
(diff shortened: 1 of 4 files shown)


src/github/folderRepositoryManager.ts
```diff
@@ -1540,6 +1540,9 @@ export const titleAndBodyFrom = (message: string): { title: string, body: string
 
 		body: idxLineBreak === -1
 			? ''
-			: message.slice(idxLineBreak + 1),
+			: message
+				.slice(idxLineBreak + 1)
+				.trim()
+				.replace(/\n+/g, ' '),
```
alexr00, line 1546: This has the unfortunate side effect of causing my multiline commit message to have the newlines replaced with a space even when I have `"githubPullRequests.pullRequestDescription": "commit"` set. I think this is worse than the quick input box for `"custom"` having the newlines removed.

alexr00 (changes requested):
See comment!

src/github/folderRepositoryManager.ts
heaths: Ah, yes. That is worse. Fixing that is possible, but requires refactoring the code. What is your current release timeframe that would include my original PR? If soon (please? 😉), I would prefer to hold off on fixing this PR to avoid any destabilization, especially given that this multi-line squashing behavior is pre-existing (again, though: I'm happy to address it).

alexr00: Let's wait until next week then, since we will likely release this week.

heaths: I've resolved this. Please see the updated PR.

alexr00 (approved):
Thanks for the fix!

PR #2118, opened 2020-09-08
fix: Issues titles contaning " or '.

### LINKED ISSUE #2117: Issues titles containing " or ' produce invalid branch names.
- Extension version: 0.19.0
- VSCode Version: 1.48.2
- OS: Ubuntu

Steps to Reproduce:

1. Create an issue with title containing `"` or `'`
2. Setup: `"githubIssues.issueBranchTitle": "${user}/${sanitizedIssueTitle}"`
3. Click on work on issue.

Git fails to checkout as `"`, `'` and more symbols  are not stripped out from the regex, see:
https://github.com/microsoft/vscode-pull-request-github/blob/master/src/issues/util.ts#L377

dipasqualew:
The sanitiser function regex wasn't replacing `"` and `'` which break git commands.

The regex has been updated.

Unit tests covering the bug have been added.

Closes #2117

### DIFF
--- a/src/issues/util.ts
+++ b/src/issues/util.ts
@@ -364,6 +364,15 @@ export async function createGithubPermalink(gitAPI: GitApiImpl, positionInfo?: N
 	return { permalink: `https://github.com/${new Protocol(upstream.fetchUrl).nameWithOwner}/blob/${log[0].hash}${pathSegment}#L${range.start.line + 1}-L${range.end.line + 1}`, error: undefined };
 }
 
+export function sanitizeIssueTitle(title: string): string {
+	const regex = /[~^:;'".,~#?%*[\]@\\{}]|\/\//g;
+
+	return title
+		.replace(regex, '')
+		.trim()
+		.replace(/\s+/g, '-');
+}
+
 const VARIABLE_PATTERN = /\$\{(.*?)\}/g;
 export async function variableSubstitution(value: string, issueModel?: IssueModel, defaults?: PullRequestDefaults, user?: string): Promise<string> {
 	return value.replace(VARIABLE_PATTERN, (match: string, variable: string) => {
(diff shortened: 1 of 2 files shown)


dipasqualew:
Hi @RMacfarlane!
Could this PR be reviewed & merged?

I really like the extension and simplifies my workflow a big deal - I love to create branches from issues. However, as you can see, sensitisation of input is currently broken.

Cheers

src/issues/util.ts
```diff
@@ -364,6 +364,13 @@ export async function createGithubPermalink(gitAPI: GitApiImpl, positionInfo?: N
 	return { permalink: `https://github.com/${new Protocol(upstream.fetchUrl).nameWithOwner}/blob/${log[0].hash}${pathSegment}#L${range.start.line + 1}-L${range.end.line + 1}`, error: undefined };
 }
 
+export function sanitizeIssueTitle(title: string): string {
+	return title
+		.replace(/\W+/g, '-')
```
RMacfarlane, line 369: I think this is a little bit too aggressive - if the title is entirely non English characters, it would be completely blanked out. How about keeping the old regex but adding \', \", and \. to it?

RMacfarlane (changes requested):
Thanks for this PR and thanks for writing tests!

src/issues/util.ts
dipasqualew: I think you are right! I restored the original regex and added a few more characters to it.

```
%
.
,
:
;
'
"
~
#
```

RMacfarlane (approved):
Looks great, thanks again!

dipasqualew:
Thank you! The extension is great and hopefully this fix will help me use VS Code to generate branches.

Cheers

PR #2120, opened 2020-09-09
More GitHub git provider and add entry point

alexr00:


src/common/logger.ts
```diff
@@ -27,8 +27,8 @@ class Log {
 			case LogLevel.Off:
 				return;
 			case LogLevel.Debug:
-				const hrtime = process.hrtime();
-				const timeStamp = `${hrtime[0]}s ${Math.floor(hrtime[1] / 1000000)}ms`;
+				const hrtime = new Date().getTime()/1000;
```
alexr00, line 30: @RMacfarlane I'm not sure if this is the best replacement.

src/extensionWeb.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	const version = vscode.extensions.getExtension(EXTENSION_ID)!.packageJSON.version;
+	setTelemetry(new TelemetryReporter(EXTENSION_ID, version, aiKey));
+	context.subscriptions.push(telemetry);
+
+	PersistentState.init(context);
+	const credentialStore = new CredentialStore(telemetry);
+	await credentialStore.initialize();
+
+	context.subscriptions.push(registerGithubGitProvider(credentialStore, apiImpl));
```
RMacfarlane, line 34: It looks like the only difference between this entry point and `extension.ts` is this registration. I think breaking them up is ok, but one suggestion I have to minimize code duplication is to instead create two versions of `/gitProviders/api` in the `env` folder and have different implementations of `registerBuiltInGitProvider`

src/common/logger.ts
RMacfarlane: I think this is ok! Measuring in milliseconds seems good enough

RMacfarlane (approved):

src/github/repositoriesManager.ts
```diff
@@ -103,9 +103,9 @@ export class RepositoriesManager implements vscode.Disposable {
 
 	getManagerForFile(uri: vscode.Uri): FolderRepositoryManager | undefined {
 		for (const folderManager of this.folderManagers) {
-			const managerPath = folderManager.repository.rootUri.fsPath;
-			const testUriRelativePath = uri.fsPath.substring(managerPath.length);
-			if (path.join(managerPath, testUriRelativePath) === uri.fsPath) {
+			const managerPath = folderManager.repository.rootUri.path;
+			const testUriRelativePath = uri.path.substring(managerPath.length > 1 ? managerPath.length + 1 : managerPath.length);
```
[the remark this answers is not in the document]
alexr00: Correct.

package.json
```diff
@@ -1265,6 +1263,7 @@
     "@octokit/types": "^5.1.0",
     "apollo-boost": "^0.1.22",
     "apollo-link-context": "^1.0.12",
+    "cross-fetch": "^3.0.5",
```
alexr00, line 1266: @RMacfarlane FYI, since version 3.* of node-fetch remove browser support I had to add a depenency on cross-fetch. It's only used in the browser since I don't know if it will work exactly as expected everywhere.

src/extensionWeb.ts
alexr00: Done!

PR #2121, opened 2020-09-09
Use newer node version so that graphql queries work again

alexr00:
Graphql API calls weren't working.

### DIFF
--- a/browser.webpack.config.js
+++ b/browser.webpack.config.js
@@ -34,7 +34,6 @@ function getExtensionConfig(env) {
 		},
 		resolve: {
 			alias: {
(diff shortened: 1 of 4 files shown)


RMacfarlane (approved):
Thanks for fixing this!

RMacfarlane:
Actually - this causes the webworker build to fail if you run `yarn compile-web`. I thought the issue was that it was using `cjs` there which is why I did the downgrade, there must just be something missing in the webworker webpack to handle this

RMacfarlane:
Nevermind, there isn't an issue with `cjs`, the `node-fetch` alias just needs to be removed from `browser.webpack.config.js`

alexr00:
Everything builds if I remove the node-fetch alias. Thanks!

PR #2122, opened 2020-09-09
Replace svgs with codicons and remove unused resources

alexr00:
This change reduces our total resource count to 22, 13 of which are svgs.


RMacfarlane (approved):
Nice!

PR #2154, opened 2020-09-24
Support to navigate to a TODO location in a file from an issue inside VS Code

### LINKED ISSUE #2045: Support to navigate to a TODO location in a file from an issue inside VS Code
I add a TODO and that creates an issue, then I can assign it with @ but how do I make it so that an assignee can go directly to that TODO in code? We are mainly working on .ps1 files. 

I see an option to start working on the issue and to open it on the web, but nothing to jump to that section in the file where that TODO was created it.

alexr00:
Fixes #2045

### DIFF
--- a/package.json
+++ b/package.json
@@ -698,6 +698,11 @@
         "command": "issue.signinAndRefreshList",
         "title": "Sign in and Refresh",
         "category": "GitHub Issues"
+      },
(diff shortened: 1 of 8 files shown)


RMacfarlane (approved):
Looks great!

PR #2158, opened 2020-09-25
Automatically detect PR from branch

alexr00:


src/view/reviewManager.ts
```diff
@@ -210,10 +210,15 @@ export class ReviewManager {
 		}
 
 		const branch = this._repository.state.HEAD;
-		const matchingPullRequestMetadata = await this._folderRepoManager.getMatchingPullRequestMetadataForBranch();
+		let matchingPullRequestMetadata = await this._folderRepoManager.getMatchingPullRequestMetadataForBranch();
 
 		if (!matchingPullRequestMetadata) {
 			Logger.appendLine(`Review> no matching pull request metadata found for current branch ${this._repository.state.HEAD.name}`);
+			matchingPullRequestMetadata = await this._folderRepoManager.getMatchingPullRequestMetadataFromGitHub();
```
RMacfarlane, line 217: I think it might be good to call `PullRequestGitHelper.associateBranchWithPullRequest` after this - while this doesn't really matter for the web case, on desktop it's helpful to have the git config updated with a mapping between branches and prs. When we run the "checkout" action we check the config to see if there's already a local branch, so I think we could get into a state where we accidentally create a duplicate local branch

RMacfarlane (approved):

src/view/reviewManager.ts
alexr00: Good point!

PR #2160, opened 2020-09-25
Multi root workspace: order of folders in view does not match order in explorer

### LINKED ISSUE #2157: Multi root workspace: order of folders in view does not match order in explorer
I am using a multi-root workspace with `vscode` and then `vscode-docs` folder as order, but in the view I see `vscode-docs` first and then `vscode`.

//cc @alexr00 

image

alexr00:
Fixes #2157


RMacfarlane (approved):
Looks good!

PR #2161, opened 2020-09-25
Adopt changes to the decoration provider API

RMacfarlane:
Latest file decoration API: https://github.com/microsoft/vscode/issues/54938#issuecomment-698957318


alexr00:
I saw the breakage before I saw your PR and already pushed a similar change :(

Issue #1906, opened 2020-06-15
Enable user to copy to the clipboard the link to the PR


dkent600:
On the description page I would love to be able to copy to the clipboard the PR link.  Would save me having to click the link, go to the browser and copy the link from the browser URL input field.  The whole idea of this extension is to keep the workflow within VS Code.

![image](https://user-images.githubusercontent.com/1821666/84650690-c1731300-aed6-11ea-8ea9-4f3d548c2ba5.png)

Thanks!

rishab-pdx:
@dkent600 If I'm right you're talking about providing a ```copy to clipboard``` action next to the title of PR. @RMacfarlane I would like to take this if no one has taken this yet. 

Ideally, I think there needs to be a copy icon(a notepad SVG) next to the edit icon which when clicked copies the link to clipboard. 
But for now, I have added a button that lets users copy to the link. Let me know if a notification toast or message box should be displayed on the success of copying the link.

![issue-#1960](https://user-images.githubusercontent.com/7886919/85795287-df1d5500-b6ec-11ea-95e8-1bd9811686d6.gif)

ashrodan:
This looks good @rishab-pdx have you got a PR in the works?

Would of thought we could use the same icon from github or opensource icon?
<img width="35" alt="Inline_issue_action_to_go_to_query_by_alexr00_·_Pull_Request__1633_·_microsoft_vscode-pull-request-github" src="https://user-images.githubusercontent.com/8232214/94328836-1b689600-fff9-11ea-812f-25e6f33703c9.png">

ashrodan:
Looks like a dupe of https://github.com/microsoft/vscode-pull-request-github/issues/1172 @RMacfarlane

rishab-pdx:
@ashrodan I could not get back earlier. There is a [PR](https://github.com/microsoft/vscode-pull-request-github/compare/master...rishab-pdx:rishab-issue%231906) in work which makes use of ```vscode's clipboard api```. If you could provide the ```icon``` for copy then we can get something together and push it out.

rishab-pdx:
#2203 addresses this! @RMacfarlane @ashrodan FYI

rishab-pdx:
This is addressed in #1906

partlythomas:
So #2203 is apparently reversed and the copy link is no longer visible?

Issue #1172, opened 2019-05-23
Add ability to copy link(s) to PR and commits


tnrich:
I often want to send a link of the PR I just made to my co-workers who I've requested to review it. Unfortunately it is a little difficult to do that right now. It would be awesome if there were a way to copy the link to the github PR from this view in vscode: 

![image](https://user-images.githubusercontent.com/2730609/58217090-0cabf980-7cb6-11e9-8653-f2dd5e6e01db.png)

Currently I have to actually click the link and then copy the link from the browser window that opens. 

Love the extension in general! 
Thanks!

sh-tech:
I think so too 👍

phatmann:
The Jira extension has a copy link icon next to their links. Works well and looks clean. We should do the same.

phatmann:
Another (easier?) solution is to add a Copy Link command to the contextual menu for PRs in the Explorer.

PR #2167, opened 2020-09-27
[WIP] Feature/copy pr link in header

ashrodan:
## Summary
As discussed in the below issues, this PR attempts to resolve this by adding a copy icon adjacent to the title. When the user clicks the copy icon, the PR URL is copied to the user's clipboard.

Resolves: https://github.com/microsoft/vscode-pull-request-github/issues/1172
Resolves: https://github.com/microsoft/vscode-pull-request-github/issues/1906

*Light (Visual Studio)*
_Extension_Development_Host__-_Pull_Request__755_—_business-intelligence
*Dark (Visual Studio)*
_Extension_Development_Host__-_Pull_Request__755_—_business-intelligence

## RFC
- links is wrapped in `<a>` tag rather than `<button>` (currently done this to avoid the inherent styling that comes with `<button>`)
- Usage of `navigator` to copy to clipboard (unsure if this works on other OS's)
- Copy icon SVG doesn't have the same contrast for darker themes 
- No clear feedback after the user hovers/clicks the icon 

## Other solutions thought of
- Discreetly add a hidden button to the PR title when clicked add to the clipboard
- After creating the PR automatically add the PR link to the clipboard with a notification (just like VS codeshare does)


ashrodan:
Keen to hear feedback on the direction of this @RMacfarlane

RMacfarlane:
Sorry for the delay, I've just gotten back from vacation. This looks good, I like the approach you're taking!

I'm also not sure how robust using `navigator` is. One potential alternative is to instead use VSCodes `clipboard` API. In the onClick, you could post a message, and then receive it in `pullRequestOverview.ts` and use vscode.env.clipboard.writeText`. This might be overkill, though. @mjbvz, do you know if it's safe to use `navigator.clipboard` directly?

For the contrast issue, I think you need to add `fill: var(--vscode-foreground)` to the css for styling the svg, that variable should update automatically when changing the theme. Maybe you already know about it, but you can run the command "Developer: Open Webview Developer Tools" when the description page is open and use this to play around with the styles of the page.

As you said, having feedback when the icon is clicked would be nice. I like the UI GitHub has for this, where the button briefly turns into a checkmark and then back. Or we could show a notification.

rishab-pdx:
#2203 addresses this! @RMacfarlane @ashrodan FYI

RMacfarlane:
Hey @ashrodan, since I haven't heard back from you, I've gone ahead and merged https://github.com/microsoft/vscode-pull-request-github/pull/2203. Thanks for taking a look at this, though!

PR #2198, opened 2020-10-06
Provide warning when token is insufficient for users

### LINKED ISSUE #2190: Git commit input field does not show @ suggestions anymore
I am using the nightly version of PR extension in selfhost web and typing `@` in the git commit input field no longer completes names.

//cc @alexr00

alexr00:
Fixes #2190


alexr00:
Example of notification:
image

RMacfarlane:
I think this should actually be done on the VSCode side - we can verify the pasted token has the requested scopes there

RMacfarlane (approved):
I went ahead and added a check there: https://github.com/microsoft/vscode/commit/64d64c70ff6cad948e70876565c99618a3491157

but having this as well doesn't hurt!

PR #2199, opened 2020-10-06
Codespaces: Created issues is empty?

### LINKED ISSUE #2185: Codespaces: Created issues is empty?
Using build 2020.9.41222 on codespaces (disabled stable), I'm unable to see this issue https://github.com/JacksonKearl/mixedCase/issues/2 in the created issues tree... it shows up fine in desktop. I tried out a few custom query strings, and I only seem to get any results with "default"

alexr00:
Fixes #2185

### DIFF
--- a/src/github/githubRepository.ts
+++ b/src/github/githubRepository.ts
@@ -423,7 +423,7 @@ export class GitHubRepository implements vscode.Disposable {
 			const { data } = await query<IssuesSearchResponse>({
 				query: schema.Issues,
 				variables: {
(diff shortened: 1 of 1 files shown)


RMacfarlane (approved):
Oh, there's a slight API difference specifically in search between web and desktop because of the tokens that are used. The web uses a new special type of "repo" access token that are meant to only work with a particular repo, GitHub has a limitation that when searching with these tokens, you always need to set what type of search you want

PR #2223, opened 2020-10-22
"Start Working On Issue" should automatically assign the Issue to me

### LINKED ISSUE #2096: "Start Working On Issue" should automatically assign the Issue to me
It seems like when I click the "Start Working on Issue" arrow, the Issue in question should be automatically assigned to me inside of GitHub.  Otherwise, I'm forced to visit the Issue's website to change the assignment which kinda defeats the purpose of the plug-in.

(This seems kind of obvious so maybe it is supposed to work this way already?  If so, it is not working for me for some reason.)

alexr00:
Fixes #2096


RMacfarlane (approved):
Nice, looks good to me!

PR #2225, opened 2020-10-22
Use dropdown instead of radio buttons for selecting type of comment

RMacfarlane:
Some styling updates to the focused view, largely changing the control for choosing the comment type

### DIFF
--- a/webviews/components/comment.tsx
+++ b/webviews/components/comment.tsx
@@ -16,6 +16,7 @@ import { editIcon, deleteIcon, commentIcon } from './icon';
 import { GithubItemStateEnum } from '../../src/github/interface';
 import { useStateProp } from '../common/hooks';
 import emitter from '../common/events';
+import { Dropdown } from './dropdown';
 
 export type Props = Partial<IComment & PullRequest> & {
 	headerInEditMode?: boolean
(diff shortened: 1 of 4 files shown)


webviews/components/comment.tsx
```diff
@@ -318,100 +319,45 @@ export function AddComment({ pendingCommentText, state, hasWritePermission, isIs
...
-		selectedValue = e.target.value as ReviewType;
-		updatePR({ pendingReviewType: e.target.value as ReviewType });
-	}
-
-	return <form id='comment-form'
-		ref={form}
-		className='comment-form main-comment-form'
-		onSubmit={onSubmit}>
+	const availableActions = pr.isAuthor
+		? { 'comment': 'Comment '}
```
alexr00, line 352: Why doesn't `{ 'comment': 'Comment '}` more closely match `COMMENT_METHODS`? I would expect it to be `{ comment: 'Comment'}` (no quotes and no space).

webviews/components/dropdown.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+	const onMethodChange = e => {
+		selectMethod(e.target.value);
+		setOptionsVisible(false);
+		const primaryButton = document.getElementById('confirm-button');
+		primaryButton.focus();
+	}
+
+	const onKeyDown = e => {
+		if (e.keyCode === KEYCODES.esc && areOptionsVisible) {
```
alexr00, line 34: Since this series of `if`s are mutually exclusive, it's slightly easier to read to make them `else if`s.

alexr00: I see now that this a copy-paste. 🤷‍♀️

alexr00 (approved):

webviews/components/comment.tsx
RMacfarlane: Thanks, fixed!

webviews/components/dropdown.tsx
RMacfarlane: Restructured this to a `switch`, which I hope is easier to read

PR #2226, opened 2020-10-23
Do not offer to delete default branch after merge

### LINKED ISSUE #1911: Should not prompt to delete master branch after merge
If a change is committed to <some fork>:master, and then merged, the PR (when viewed from within this plugin in VS Code) prompts you to delete the master branch. This prompt would obviously be fine if it was anything other than master, but deleting master is not really what anyone wants (particularly as selecting the "delete on remote" option puts the remote fork in a confused state where GitHub web interface thinks the forking process is still underway and the only way to fix that is to push master back to the remote fork).

Indeed, the GitHub web interface suggests deleting the **fork**, not the master branch in such situations.

Version information:
- Extension version: 0.16.0
- VSCode Version: 1.45.1 
- OS: Windows 10

Steps to Reproduce:

1. Fork repository.
2. Commit to master of fork.
3. Make a PR from within VS Code using this extension
4. Merge the PR
5. Refresh the PR in VS Code
6. See that it suggests deleting the master branch (which is inconsistent with the GitHub web interface)

Note: If this has to be coded as a special case, it may rear it's head again when [master is replaced with a different term](https://www.zdnet.com/article/github-to-replace-master-with-alternative-term-to-avoid-slavery-references/).

RMacfarlane:
Fixes #1911

### DIFF
--- a/src/github/activityBarViewProvider.ts
+++ b/src/github/activityBarViewProvider.ts
@@ -15,7 +15,7 @@ import { ReviewEvent as CommonReviewEvent } from '../common/timelineEvent';
 import { getNonce, IRequestMessage, WebviewBase } from '../common/webview';
 import { parseReviewers } from './utils';
 
-export class PullRequestViewProvider extends WebviewBase implements vscode.WebviewViewProvider  {
+export class PullRequestViewProvider extends WebviewBase implements vscode.WebviewViewProvider {
 	public static readonly viewType = 'github:activePullRequest';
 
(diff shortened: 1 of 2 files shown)


alexr00 (approved):
It would be nice if we did offer to delete the fork, that sounds like a separate work item.

RMacfarlane:
Agreed! Created https://github.com/microsoft/vscode-pull-request-github/issues/2230 to track it, I believe we need to request additional permissions to be able to do this

PR #2227, opened 2020-10-23
Create PR - "choose target branch" should show list of branches

### LINKED ISSUE #814: Create PR - "choose target branch" should show list of branches
- Create PR
- Try to send PR to a branch with a long name that isn't `master`
- I have to copy/paste it or type from memory.

I should see a list of branches like "Git: Checkout to..."

image

RMacfarlane:
Fixes #814


src/view/reviewManager.ts
```diff
@@ -729,6 +729,48 @@ export class ReviewManager {
...
+	private async _chooseTargetBranch(targetRemote: RemoteQuickPickItem, pullRequestDefaults: PullRequestDefaults): Promise<string | undefined> {
+		const base: string = targetRemote.remote
+			? (await this._folderRepoManager.getMetadata(targetRemote.remote.remoteName)).default_branch
+			: pullRequestDefaults.base;
+
+		let target: string | undefined;
+		try {
+			const branches = await this._folderRepoManager.listBranches(targetRemote.owner, targetRemote.name);
+			target = await new Promise(async (resolve, reject) => {
+				const branchPicker = vscode.window.createQuickPick();
```
alexr00, line 741: Can you use the simpler `showQuickPick` here instead?

alexr00 (approved):

src/view/reviewManager.ts
RMacfarlane: I don't think so, I don't see a way to set an initial value with the options for it

PR #2238, opened 2020-10-28
Provide option to toggle between tree view and flat list view in "Changes in Pull Request" 

### LINKED ISSUE #2177: Changes in pull request: provide option to see changes as flat list
Testing #2136

Similar to how the changes view in SCM can toggle between flat list and tree:

image

ejkim1996:
Closes #2177 
The "Changes In Pull Request" view now displays an icon to let users switch between seeing the list of files changed in `tree` or `flat` list view, to match the behavior in the SCM view.

Screen Shot 2020-11-05 at 4 46 23 PM
Screen Shot 2020-11-05 at 4 46 11 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -512,9 +512,15 @@
         "category": "GitHub Pull Requests"
       },
       {
-        "command": "pr.toggleFileListLayout",
-        "title": "Toggle File List Layout",
-        "icon": "$(gear)",
+        "command": "pr.setFileListLayoutAsTree",
+        "title": "Toggle View Mode",
+        "icon": "$(list-tree)",
+        "category": "GitHub Pull Requests"
+      },
+      {
+        "command": "pr.setFileListLayoutAsFlat",
+        "title": "Toggle View Mode",
+        "icon": "$(list-flat)",
         "category": "GitHub Pull Requests"
       },
       {
@@ -996,8 +1002,13 @@
           "group": "navigation"
         },
         {
-          "command": "pr.toggleFileListLayout",
-          "when": "view =~ /prStatus/",
+          "command": "pr.setFileListLayoutAsTree",
+          "when": "view =~ /prStatus/ && fileListLayout:flat",
+          "group": "navigation"
+        },
+        {
+          "command": "pr.setFileListLayoutAsFlat",
+          "when": "view =~ /prStatus/ && !fileListLayout:flat",
           "group": "navigation"
         },
         {
--- a/package.json
+++ b/package.json
@@ -844,6 +844,14 @@
           "when": "gitOpenRepositoryCount != 0 && github:authenticated && github:hasGitHubRemotes"
         },
         {
+          "command": "pr.setFileListLayoutAsTree",
+          "when": "false"
+        },
+        {
+          "command": "pr.setFileListLayoutAsFlat",
+          "when": "false"
+        },
+        {
           "command": "pr.refreshChanges",
           "when": "false"
         },
--- a/package.json
+++ b/package.json
@@ -512,6 +512,12 @@
         "category": "GitHub Pull Requests"
       },
       {
+        "command": "pr.toggleFileListLayout",
+        "title": "Toggle File List Layout",
+        "icon": "$(gear)",
+        "category": "GitHub Pull Requests"
+      },
+      {
         "command": "pr.refreshChanges",
         "title": "Refresh",
         "icon": "$(refresh)",
@@ -986,6 +992,11 @@
         },
         {
           "command": "pr.refreshChanges",
+          "when": "view =~ /prStatus/",
+          "group": "navigation"
+        },
+        {
+          "command": "pr.toggleFileListLayout",
           "when": "view =~ /prStatus/",
           "group": "navigation"
         },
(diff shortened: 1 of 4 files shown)


RMacfarlane:
This looks great!

For the icon, you can see a list of icons that are available from VSCode here: https://microsoft.github.io/vscode-codicons/dist/codicon.html

Any of these should work with the "$(icon-name)" syntax. Maybe the `list-tree` icon would be best here?

ejkim1996:
Thank you for the feedback and the link to the icons!

It seems like in the SCM view the icon changes between the `list-tree` icon and the `list-flat` icon. I think it would be nice to replicate that in this extension as well. I tried to look at the `vscode` repository to see how it was set up there, but the setup for icons seemed different and I couldn't figure it out. Could you point me to the right direction on how we could make it work?

RMacfarlane:
Having it change between icons would be great!

There isn't a way to update the icon of a command dynamically, so I think to make this work, we actually need to add two commands instead of one that acts as a toggle. You can control when a command is shown by using a "when" property in the `package.json` for it, and then creating a context key for which one should be shown based on the current setting. An example of setting a context key is here: https://github.com/microsoft/vscode-pull-request-github/blob/c32a50c004c0f0cfdd0f1714ad21b1f3f94916c4/src/view/prChangesTreeDataProvider.ts#L49

ejkim1996:
Thank you for your help @RMacfarlane !

I've made the changes you suggested and got the icons to toggle between the `list-tree` and `list-flat` icons. I've updated the main comment to include the screenshots so you can see how it looks, and marked the PR as "Ready for review"

package.json
```diff
@@ -511,6 +511,18 @@
         "icon": "$(refresh)",
         "category": "GitHub Pull Requests"
       },
+      {
+        "command": "pr.setFileListLayoutAsTree",
```
RMacfarlane, line 515: I think these should be hidden in the command palette, you can do that by adding them in the "commandPalette" section below with "when": "false"

src/commands.ts
```diff
@@ -594,6 +594,24 @@ export function registerCommands(context: vscode.ExtensionContext, reposManager:
 		});
 	}));
 
+	context.subscriptions.push(vscode.commands.registerCommand('pr.setFileListLayoutAsTree', async _ => {
+		vscode.workspace.getConfiguration('githubPullRequests').update('fileListLayout', 'tree');
+		await vscode.commands.executeCommand(
```
RMacfarlane, line 599: We also need to set this value on startup, so that if you switch it to flat, the correct command is shown when you reload VS Code.

Somewhere in the `init` method in `extension.ts` would be a good place to put that

RMacfarlane (changes requested):
Looks really good! Just had a couple of comments

ejkim1996:
> Looks really good! Just had a couple of comments

@RMacfarlane Thanks for the review, I've made the suggested changes! 😄

src/extension.ts
```diff
@@ -90,6 +90,12 @@ async function init(context: vscode.ExtensionContext, git: GitApiImpl, credentia
 	context.subscriptions.push(reviewsManager);
 	tree.initialize(reposManager);
 	registerCommands(context, reposManager, reviewManagers, telemetry, credentialStore, tree);
+	const layout = vscode.workspace.getConfiguration('githubPullRequests').get<string>('fileListLayout');
+	if (layout === 'flat') {
+		await vscode.commands.executeCommand('pr.setFileListLayoutAsFlat');
```
RMacfarlane, line 95: I think it would be better to just directly do `setContext` here to avoid unneccessarily updating the setting

RMacfarlane (changes requested):
Works great! One last thing and then this will be ready to merge - there's an `onDidChangeConfiguration` listener in `prChangesTreeDataProvider.ts` that responds to changes in the layout setting. I think we should also update the context key there, so that the icon changes if someone changes the setting from the settings UI instead of using the button

ejkim1996:
Thanks @RMacfarlane! I've made the requested changes. I also noticed that since we're updating the context key in the `onDidChangeConfiguration` listener, we can remove the `setContext` command from the `setFileListLayoutAs` commands, so that change has been made as well.

src/commands.ts
```diff
@@ -594,6 +594,14 @@ export function registerCommands(context: vscode.ExtensionContext, reposManager:
 		});
 	}));
 
+	context.subscriptions.push(vscode.commands.registerCommand('pr.setFileListLayoutAsTree', _ => {
+		vscode.workspace.getConfiguration('githubPullRequests').update('fileListLayout', 'tree');
```
RMacfarlane, line 598: Sorry, one last thing that I missed! There is an optional argument to `update` that we should set to `true` at the end of this. This controls whether the setting is updated for the workspace, or for the user's overall settings. If the update it done just for the workspace, it creates a file in that folder. We just want to update the user's settings

RMacfarlane (changes requested):

src/commands.ts
ejkim1996: No worries! Thank you for explaining the reasoning behind it. I've made the changes pushed them

RMacfarlane (approved):
Thank you!

PR #2250, opened 2020-11-03
Conversation comment source links don't work

### LINKED ISSUE #1370: Conversation comment source links don't work
Issue Type: <b>Bug</b>

In conversation view, the source comments source locations are decorated as links, but do nothing when clicked or ctrl-clicked.

Extension version: 0.11.2
VS Code version: Code 1.38.1 (b37e54c98e1a74ba89e03073e5a3761284e3ffb0, 2019-09-11T13:35:15.005Z)
OS version: Windows_NT x64 10.0.18362

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-2600 CPU @ 3.40GHz (8 x 3392)|
|GPU Status|2d_canvas: enabled<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>oop_rasterization: unavailable_off<br>protected_video_decode: unavailable_off<br>rasterization: unavailable_off<br>skia_deferred_display_list: disabled_off<br>skia_renderer: disabled_off<br>surface_synchronization: enabled_on<br>video_decode: enabled<br>viz_display_compositor: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|15.88GB (0.62GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|
</details>

RMacfarlane:
Fixes #1370

Previously, opening a file from the description page worked by trying to find that file within the tree. There are a couple of issues with that approach - the overview page gets initialized with the description node, which could either be from the "Pull Requests" tree or the "Changes in PR" tree, but changing the checkout state of the PR doesn't update this. This is especially hard if the page starts in review mode, and then exits, there isn't a good way to find the PR within the other tree, and also it could be in multiple places.

Intead of doing that, I've decoupled this from looking at the tree at all. Now it should use the comment and try to match it with a file change, and simply use that to open a diff. I needed to make the PullRequestModel have access to the FolderRepositoryManager to make this work, which is kind of ugly. (or at least, rippled into a larger change). Open to suggestions on that.

One thing the previous approach did nicely was selecting the relevant file in the tree while opening it. I want to add this separately for the "Changes in PR" tree, it could listen to changes in open text documents and reveal stuff accordingly. I think doing the same for the other tree is impossible, since the PR could be in multiple places in the tree.

Also, I realized that navigating to "outdated" comments wasn't working previously, and have changes those to not be links.

### DIFF
--- a/src/github/githubRepository.ts
+++ b/src/github/githubRepository.ts
@@ -8,7 +8,7 @@ import { Octokit } from '@octokit/rest';
 import * as OctokitTypes from '@octokit/types';
 import Logger from '../common/logger';
 import { Remote, parseRemote } from '../common/remote';
-import { IAccount, RepoAccessAndMergeMethods, IMilestone, Issue } from './interface';
+import { IAccount, RepoAccessAndMergeMethods, IMilestone, Issue, PullRequest } from './interface';
 import { PullRequestModel } from './pullRequestModel';
 import { CredentialStore, GitHub } from './credentials';
 import { AuthenticationError } from '../common/authentication';
@@ -35,8 +35,8 @@ export interface IssueData extends ItemsData {
 	hasMorePages: boolean;
 }
 
-export interface PullRequestData extends IssueData {
-	items: PullRequestModel[];
+export interface PullRequestData extends ItemsData {
+	items: PullRequest[];
 }
 
 export interface MilestoneData extends ItemsData {
@@ -280,10 +280,10 @@ export class GitHubRepository implements vscode.Disposable {
 							return null;
 						}
 
-						return new PullRequestModel(this._telemetry, this, this.remote, convertRESTPullRequestToRawPullRequest(pullRequest, this));
+						return convertRESTPullRequestToRawPullRequest(pullRequest, this);
 					}
 				)
-				.filter(item => item !== null) as PullRequestModel[];
+				.filter(item => item !== null) as PullRequest[];
 
 			Logger.debug(`Fetch all pull requests - done`, GitHubRepository.ID);
 			return {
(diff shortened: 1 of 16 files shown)


src/github/githubRepository.ts
```diff
@@ -581,7 +581,7 @@ export class GitHubRepository implements vscode.Disposable {
 		}
 	}
 
-	async getPullRequest(id: number): Promise<PullRequestModel | undefined> {
+	async getPullRequest(id: number): Promise<PullRequest | undefined> {
```
alexr00, line 584: Why have all these methods changed from returning a PullRequestModel to a PullRequest?

src/github/pullRequestModel.ts
```diff
@@ -721,6 +723,43 @@ export class PullRequestModel extends IssueModel implements IPullRequestModel {
 		};
 	}
 
+	async openDiffFromComment(comment: IComment): Promise<void> {
```
alexr00, line 726: I'm probably missing something here, but all of the properties/methods on `this` in this method seem to be public. To avoid giving `PullRequestModel` a reference the folder repository manager could you make this method static or move it into a utility funciton?

alexr00 (approved):

src/github/pullRequestModel.ts
RMacfarlane: Good point! I made this method static and it really reduced this change set. thanks!

src/github/githubRepository.ts
RMacfarlane: This was a consequence of adding the folder manager to the PR - I was delaying creation of the model since I didn't have it here. Have now reverted this

PR #2301, opened 2020-11-30
Update nightly build yml to always build even when no changes

alexr00:
When proposed API changes but there have been no changes in vscode-pull-request-github then we don't get notified that there's a break. If we always build, even when there are no changes, then this should help us catch such problems earlier.


RMacfarlane (approved):
Nice!

PR #2335, opened 2020-12-29
Use a webview view for the create PR flow

RMacfarlane:
Changes the create PR flow to use a webview instead of the series of quickpicks.

I also plan to add another tree view that shows the diff between the current branch and the selected base branch, separately


src/github/createPRViewProvider.ts
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	private async getDescription(): Promise<string> {
+		const method = vscode.workspace.getConfiguration('githubPullRequests').get<PullRequestDescriptionSource>('pullRequestDescription', PullRequestDescriptionSourceEnum.Ask);
+
+		switch (method) {
+
+			case PullRequestDescriptionSourceEnum.Template:
+				const templateUris = await this._folderRepositoryManager.getPullRequestTemplates();
+				if (templateUris[0]) {
```
alexr00, line 87: I've been doing a length check before accessing `[0]`. TIL that this was an option! 😅

src/github/createPRViewProvider.ts
```diff
@@ -0,0 +1,258 @@
...
+					try {
+						const templateContent = await vscode.workspace.fs.readFile(templateUris[0]);
+						return templateContent.toString();
+					} catch (e) {
+						// Logger.appendLine(`Reading pull request template failed: ${e}`);
```
alexr00, line 92: Log line not needed?

src/github/createPRViewProvider.ts
```diff
@@ -0,0 +1,258 @@
...
+		const config = vscode.workspace.getConfiguration(PR_SETTINGS_NAMESPACE);
+		if (config.get(PR_TITLE) === titleSource) {
+			return;
+		}
+
+		const showPrompt = PersistentState.fetch('prompts', stateKey);
+		if (!showPrompt || showPrompt === PersistentState.MISSING) {
+			vscode.window.showInformationMessage(
+				`Would you like to always use the ${message} as the title of the pull request?`,
+				{ modal: true },
```
alexr00, line 153: Generally I think modal dialogs should only be used when action must be taken by the user. Since this is kind of an optional think, maybe it would be better as non-modal?

src/github/createPRViewProvider.ts
```diff
@@ -0,0 +1,258 @@
...
+		const nonce = getNonce();
+
+		return `<!DOCTYPE html>
+		<html lang="en">
+		<head>
+			<meta charset="UTF-8">
+			<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src vscode-resource: https:; script-src 'nonce-${nonce}'; style-src vscode-resource: 'unsafe-inline' http: https: data:;">
+			<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+			<title>Active Pull Request</title>
```
alexr00, line 250: `Create Pull Request` ?

src/github/folderRepositoryManager.ts
```diff
@@ -962,7 +953,7 @@ export class FolderRepositoryManager implements vscode.Disposable {
 				// There are unpushed commits
 				if (this._repository.state.HEAD?.ahead) {
 					// Offer to push changes
-					const shouldPush = await vscode.window.showInformationMessage(`There are currently no commits between '${params.base}' and '${params.head}'. Do you want to push your local commits and try again?`, 'Yes', 'Cancel');
+					const shouldPush = await vscode.window.showInformationMessage(`There are currently no commits between '${params.base}' and '${params.head}'. Do you want to push your local commits and try again?`, { modal: true }, 'Yes', 'Cancel');
```
alexr00, line 956: I like the modal changes here!

src/view/reviewManager.ts
```diff
@@ -641,260 +644,28 @@ export class ReviewManager {
...
-					ignoreFocusOut: true,
-					placeHolder: 'Select the pull request template to use'
-				});
-
-				// Treat user pressing escape as cancel
-				if (!targetTemplate) {
-					return;
+				if (createdPR) {
+					await this.updateState();
+					await vscode.commands.executeCommand('pr.openDescription', createdPR);
```
alexr00, line 661: Totally up to you, but I find these `executeCommand`s make it difficult to refactor and navigate the code. Instead, you could refactor the `pr.openDescription` command to call a function, then call that same function here.

webviews/createPullRequestView/index.css
```diff
@@ -0,0 +1,70 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+input, textarea {
+	padding: 5px !important;
```
alexr00, line 7: TIL about !important too. Cool.

package.json
```diff
@@ -1306,11 +1313,18 @@
     ]
   },
   "scripts": {
-    "vscode:prepublish": "webpack --env.production --config-name webviewIndex.js && webpack --env.production --config-name activityBar-webviewIndex.js && webpack --env.production --config-name extension && npx webpack-cli --config browser.webpack.config --env.production --config-name extension --mode none",
+    "webpack-description": "webpack --config-name webviewIndex.js",
+    "webpack-active-pr": "webpack --config-name activityBar-webviewIndex.js",
+    "webpack-create": "webpack --config-name createPR-webviewIndex.js",
+
+    "webpack:dev": "yarn run webpack-description --env.development && yarn run webpack-active-pr --env.development && yarn run webpack-create --env.production",
```
alexr00, line 1320: Copy paste error? At the end of the line, should it be `--env.development`?

alexr00: And thanks for breaking this up, it's way more readable.

alexr00 (approved):
I'm very excited about this change!

package.json
RMacfarlane: Thanks for catching that!

src/view/reviewManager.ts
RMacfarlane: Yeah, good point, the lack of type checking for the arguments is also a major downside. I've factored out a function to use

src/github/createPRViewProvider.ts
RMacfarlane: Thanks, commented it back in!

src/github/createPRViewProvider.ts
RMacfarlane: I think in this case it makes more sense to have it as a modal, because I think it would be easy to miss as a notification, and also this matches some other patterns we have, like when you move a ts file for the first time and you get a dialog asking if you always want to rename imports. I'll bring this up in the UX call, though

package.json
RMacfarlane: I realize I missed updating the `compile-web` command below, do you know why we sometimes use `webpack-cli` and sometimes use `webpack`? I'm also not sure why we set mode to `none` there and for the extension in prepublish, from looking at the documentation, the major difference between modes is if names get mangled or not. Maybe related to the problem with node-fetch?

PR #2340, opened 2021-01-06
Force nowrap for all timestamps in description

### LINKED ISSUE #2339: Prevent text wrap in commits list (description view)
Issue Type: <b>Feature Request</b>

Would you accept PR that prevents text wrap for timestamp in commits list?

In my case it looks like this:
image

Which will look much better if displayed in one line
image

If yes - can it be applied to all timestamps on description view (comment header etc.) or only commits list?

Extension version: 2021.1.4862
VS Code version: Code - Insiders 1.53.0-insider (98106c48a07d4d07f0f71b4db9b3ff156f223339, 2021-01-01T05:22:15.141Z)
OS version: Windows_NT x64 10.0.19041

IllusionMH:
Fixes #2339

@RMacfarlane regarding https://github.com/microsoft/vscode-pull-request-github/issues/2339#issuecomment-754111029 
Looks like there were already 2 places that had `nowrap` for timestamps (PR/Description header and comment header) and overflow most likely be caused by long author name and not only in headers but also in reviewers, PR header etc.
<details>
<summary>Overflow demo (large size from High DPI display 😢)</summary>

image

</details>

To keep this PR small and focused on single problem I haven't include updated to author name caused overflows since different places have different problems (overlap with icon that is shown on hover, hiding commit message etc.). I will try to look on author name overflow and overflows in headers for line comment later.

### DIFF
--- a/webviews/editorWebview/index.css
+++ b/webviews/editorWebview/index.css
@@ -709,6 +709,7 @@ code {
 .timestamp,
 .timestamp:hover {
 	color: inherit;
+	white-space: nowrap;
 }
 
 /** Theming */


RMacfarlane (approved):
Thanks, this looks good!

PR #2342, opened 2021-01-06
Remove buttons next to title input in create PR flow

RMacfarlane:
Based on feedback from today's UI sync, this removes the buttons that were next to the title input in the create PR flow. Now, if the setting for PR title or description are explicitly set by the user, we honor those, otherwise we imitate GitHub's behavior. So for the title, we use the branch name if there are multiple commits between the branchs, otherwise we use the last commit message. For the description, we use the PR template if it's available, or the body of the last commit message.


alexr00 (approved):
🚀

PR #2343, opened 2021-01-06
Exclude more files from package

### LINKED ISSUE #2341: Exclude more files from package
The nightly build has failed because we can only have 25 file in our package (since we also package for web): https://rebornix.visualstudio.com/Pull%20Request/_build/results?buildId=2090&view=logs&j=3bd52376-4866-57aa-dfdb-d04952b84d8d&t=14bd693e-34d1-5172-806f-7299de473e6e

We should see if more files can be excluded so we can be below the max.

RMacfarlane:
Fixes #2341

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -14,7 +14,7 @@ dist/browser/*.map
 .readme/**
 documentation/**
 .vscode-test/**
-webpack.config.js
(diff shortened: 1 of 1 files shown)


alexr00 (approved):
Thank you!

PR #2344, opened 2021-01-08
Add a diff view to the create PR flow

RMacfarlane:
Adds a new view that shows a diff between the selected and current branch. This adds a "createPullRequestHelper" which keeps track of both views and can send information from the webview to the diff view so that it can refresh information.

Also fixed an issue in create where I was only passing the branch name instead of owner:branchName, which causes issues for forks.

### DIFF
--- a/src/view/treeNodes/fileChangeNode.ts
+++ b/src/view/treeNodes/fileChangeNode.ts
@@ -341,6 +341,63 @@ export class GitFileChangeNode extends FileChangeNode implements vscode.TreeItem
 	}
 }
 
+/**
+ * File change node whose content is resolved from GitHub. For files not yet associated with a pull request.
+ */
+export class GitHubFileChangeNode extends TreeNode implements vscode.TreeItem {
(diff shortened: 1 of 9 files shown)


src/view/treeNodes/fileChangeNode.ts
```diff
@@ -341,6 +341,63 @@ export class GitFileChangeNode extends FileChangeNode implements vscode.TreeItem
...
+				break;
+
+			case GitChangeType.DELETE:
+				headURI = vscode.Uri.file(fileName).with({ scheme: 'github', query: JSON.stringify({ fileName, branch: headBranch, isEmpty: true }) });
+				break;
+		}
+
+		this.command = {
+			title: 'Open Diff',
+			command: 'vscode.diff',
```
alexr00, line 387: Glad to see it just uses vscode.diff!

alexr00 (approved):
Another change that I'm super excited about!

PR #2361, opened 2021-01-15
Open GitHub viewlet automatically when opening a PR from codespaces

### LINKED ISSUE #2360: Open GitHub viewlet automatically when opening a PR from codespaces

RMacfarlane:
Fixes #2360

### DIFF
--- a/src/common/webview.ts
+++ b/src/common/webview.ts
@@ -89,3 +89,16 @@ export class WebviewBase {
 		this._disposables.forEach(d => d.dispose());
 	}
 }
+
+export class WebviewViewBase extends WebviewBase {
+	public readonly viewType: string;
+	protected _view?: vscode.WebviewView;
+
+	public show() {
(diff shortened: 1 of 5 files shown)


src/common/webview.ts
```diff
@@ -88,4 +88,17 @@ export class WebviewBase {
...
+}
+
+export class WebviewViewBase extends WebviewBase {
+	public readonly viewType: string;
+	protected _view?: vscode.WebviewView;
+
+	public show() {
+		if (this._view) {
+			this._view.show();
+		} else {
```
alexr00, line 100: When is there not a `this._view`?

alexr00 (approved):

src/common/webview.ts
RMacfarlane: Before the webview has been resolved, so if it has been constructed but not revealed yet

PR #2362, opened 2021-01-15
Adds notification to offer PR create on publish

eamodio:
When a branch is published to a remote, we will show a notification and ask if the user wants to create a new pull request.

image


src/extension.ts
```diff
@@ -51,6 +54,22 @@ async function init(context: vscode.ExtensionContext, git: GitApiImpl, credentia
 		}
 	});
 
+	context.subscriptions.push(git.onDidPublish(async e => {
+		// Only notify on branch publish events
+		if (!e.branch || PersistentState.fetch(PROMPTS_SCOPE, PROMPT_TO_CREATE_PR_ON_PUBLISH_KEY) === false) {
```
RMacfarlane, line 59: I think we should also avoid showing this notification if you're already in the create view - if you don't have an upstream for you current branch, we show a button that will run the publish action from there.

You could move this to `reviewManager` to have access to the `CreatePullRequestHelper` there, and also direct access to the create method instead of executing the command. `CreatePullRequestHelper` doesn't expose anything right now that indicates if create is in progress, so that would need to be added

src/extension.ts
```diff
@@ -51,6 +54,27 @@ async function init(context: vscode.ExtensionContext, git: GitApiImpl, credentia
...
+			return;
+		}
+
+		const reviewManager = reviewManagers.find(reviewManager => reviewManager.repository.rootUri.toString() === e.repository.rootUri.toString());
```
RMacfarlane, line 63: there are a couple of tslint warnings here, `reviewManager` is a shadowed variable and some missing semicolons below

RMacfarlane (approved):
LGTM! just has a couple of linter warnings

src/extension.ts
Donna3344: 👌👌

PR #2364, opened 2021-01-17
Sanitizing issue titles lowercases text

cone56:
Branch names created using the `${sanitizedIssueTitle}` variable substitution will have a lowercase title

### DIFF
--- a/src/issues/util.ts
+++ b/src/issues/util.ts
@@ -421,7 +421,8 @@ export function sanitizeIssueTitle(title: string): string {
 	return title
 		.replace(regex, '')
 		.trim()
-		.replace(/\s+/g, '-');
+		.replace(/\s+/g, '-')
+		.toLowerCase();
 }
 
 const VARIABLE_PATTERN = /\$\{(.*?)\}/g;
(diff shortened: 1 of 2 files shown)


RMacfarlane:
@cone56 Can you explain a bit more why this change is needed?

cone56:
Sorry, I should have explained my rationale earlier:

- Ease of use: lowercase is easier to type in git or other CLI tools 
- Ecosystem: it would match other services that create branches from issue titles including Jira/Bitbucket and GitLab
- Personal preference: it'd match how I would name the branch if I were to create it manually

krlmlr:
Agree that lowercase is better.

While we're at it: should we also sanitize `()` and other special characters? Especially the former don't work well in `bash` .

cone56:
I'd be happy to update my PR to support parentheses if @RMacfarlane would provide an indication that it would be accepted

alexr00:
@cone56 I would prefer to leave `sanitizedIssueTitle` as is, and instead have a new variable `sanitizedLowercaseIssueTitle`. If you want to make a PR for the new variable then I would be happy to review it!

PR #2392, opened 2021-01-25
Adds webview debugging

eamodio:
Sets up webpack & launch config to support webview debugging in VS Code


RMacfarlane (dismissed):
Nice!

eamodio:
@RMacfarlane I updated this to work with the new build and it now should work with all webviews.

alexr00:
Thanks for the PR! We have since added support for webview debugging.

Issue #2412, opened 2021-01-26
Warning border on pull request title?


JacksonKearl:
Testing Microsoft/vscode#114967

Not clear why this is yellow -- in search view we use that for errors.
![image](https://user-images.githubusercontent.com/8586769/105898484-57732e00-5fce-11eb-8936-42b8c5f10cbf.png)

the description is blue as expected
![image](https://user-images.githubusercontent.com/8586769/105898527-6823a400-5fce-11eb-86d7-834e44fa0bae.png)

PR #2423, opened 2021-01-27
fix focus color for select and input

TylerLeonhardt:
fixes https://github.com/microsoft/vscode-pull-request-github/issues/2412

select and text input's should also use vscode focus color like textarea.


RMacfarlane (approved):
Perfect, thanks!

PR #2433, opened 2021-01-28
prevent description box resize from being too small

### LINKED ISSUE #2408: Description box can be resized
https://github.com/microsoft/vscode/issues/114967

PR flow, the descripiton box can be manually resized.
While I personally believe it would be nicer if it behaves just like the SCM commit input box (no manual resize, shift + enter for more lines) but just that it stars with two lines.

However if you prefer to keep the current manualy resize behavior I suggest that you introduce a min limit. Since currently I can resize it to be smaller than 1 line and than it looks very broken.

TylerLeonhardt:
Fixes #2408

This is the smallest you can make the body now:
image

### DIFF
--- a/webviews/createPullRequestView/index.css
+++ b/webviews/createPullRequestView/index.css
@@ -13,6 +13,8 @@ select {
 
 textarea {
 	height: 100px;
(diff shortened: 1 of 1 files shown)


RMacfarlane (approved):
Looks great, thanks!

PR #2451, opened 2021-02-03
PR: Using inputbox warnings as showing errors is not accessible

### LINKED ISSUE #2399: PR: Using inputbox warnings as showing errors is not accessible
https://github.com/microsoft/vscode/issues/114967

Try to create a PR with an empty title through the new create PR flow.
Notice the error under the input.

However turn on the screen reader and notice that nothing is being read out.

So if a user tries to create a PR with no title we should properly read out the error to the user.

RMacfarlane:
Fixes #2399


alexr00 (approved):
LGTM

Issue #339, opened 2018-08-30
Support resolving conversations


Tyriar:
<img width="789" alt="screen shot 2018-08-30 at 9 55 03 am" src="https://user-images.githubusercontent.com/2193314/44866689-c9e35680-ac3a-11e8-9211-735d7c88c1bb.png">

Needs support for deemphasizing outdated/resolved comments https://github.com/Microsoft/vscode-pull-request-github/issues/89

aoberoi:
I would like to add that I don't necessarily want "outdated" comments hidden, but i would like "resolved" comments hidden. I hope that the resolve button is available and that these two properties can be filtered separately.

myartsev:
+1 just tried out the extension and can see myself using it regularly if resolving conversations directly from VSCode was possible, otherwise, there's too much noise and I end up going directly to the GitHub website.

stgarf:
Same as the last two comments -- If in the Github PR UI the `resolved comments` and `outdated comments` are hidden when appropriate (e.g. resolved and/or outdated) but I also can click a small icon to reveal them to take a second-glance, this would be the icing on the cake for this extension. I basically don't need to leave VSCode... :)

JaimieMurdock:
I would also really like to see this feature. I'm still reviewing PRs through the website instead of VSCode because of it. As @dylmye pointed out, it is exposed by the API. Would anyone have a pointer on what files need to be edited to add this support? I might be willing to work on this, but would need some mentorship.

liang2zhu1:
The extension is awesome but this features is really blocking me from using that day-to-day

ghs180:
This extension would be perfect if it had this feature :).

judy:
Throwing my hat in here too. This extension is unusable to me until I can hide resolved comments/see only unresolved comments.

MrMage:
+1. It would be great if resolved comments could at least be indicated _somehow_ in the UI, or if the extension could remember the "collapsed" state across closing and re-opening a particular diff.

hasbro17:
+1 for remembering that a comment is resolved. Or at least remember if we've collapsed the comment locally within a review session.
Great extension overall except for some rough edges like this which break your focus while reviewing and force you back to the browser to see what comments are outdated.

MrMage:
[...]

PR #2471, opened 2021-02-05
Render resolved conversations differently on description page

RMacfarlane:
Fetch whether review threads are resolved, and show them collapsed on the description page.
Screen Shot 2021-02-05 at 12 00 45 PM

Screen Shot 2021-02-05 at 12 00 52 PM

First part of https://github.com/microsoft/vscode-pull-request-github/issues/339

### DIFF
--- a/src/github/utils.ts
+++ b/src/github/utils.ts
@@ -255,7 +255,7 @@ export function convertGraphQLEventType(text: string) {
 	}
 }
 
-export function parseGraphQLComment(comment: GraphQL.ReviewComment): IComment {
+export function parseGraphQLComment(comment: GraphQL.ReviewComment, isResolved: boolean): IComment {
 	const c: IComment = {
 		id: comment.databaseId,
 		url: comment.url,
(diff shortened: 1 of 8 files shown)


src/github/utils.ts
```diff
@@ -255,7 +255,7 @@ export function convertGraphQLEventType(text: string) {
 	}
 }
 
-export function parseGraphQLComment(comment: GraphQL.ReviewComment): IComment {
+export function parseGraphQLComment(comment: GraphQL.ReviewComment, isResolved: boolean): IComment {
```
alexr00, line 258: Why not update `GraphQL.ReviewComment` to include the `resolved` property so that you don't have to pass it as an additional param here?

alexr00 (approved):

src/github/utils.ts
RMacfarlane: The property is actually at the thread level instead of on individual comments, but I thought adding a concept of threads instead of the comments we use everywhere was too big of a refactoring

PR #2488, opened 2021-02-10
Add experimentation service

RMacfarlane:
Adopt the `vscode-tas-client` package for running experiments

### DIFF
--- a/src/experimentationService.ts
+++ b/src/experimentationService.ts
@@ -6,7 +6,13 @@
 import * as vscode from 'vscode';
 import * as path from 'path';
 import TelemetryReporter from 'vscode-extension-telemetry';
-import { getExperimentationService, IExperimentationTelemetry, TargetPopulation } from 'vscode-tas-client';
+import { getExperimentationService, IExperimentationService, IExperimentationTelemetry, TargetPopulation } from 'vscode-tas-client';
+
+/* __GDPR__
+	"query-expfeature" : {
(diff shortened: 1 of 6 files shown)


src/experimentationService.ts
```diff
@@ -0,0 +1,84 @@
...
+		default: return TargetPopulation.Public;
+	}
+}
+
+export async function createExperimentationService(context: vscode.ExtensionContext, experimentationTelemetry: ExperimentationTelemetry) {
+	const pkg = await getPackageConfig(context.extensionPath);
+	const product = await getProductConfig(vscode.env.appRoot);
+	const targetPopulation = getTargetPopulation(product);
+
+	return getExperimentationService(`${pkg.publisher}.${pkg.name}`, pkg.version, targetPopulation, experimentationTelemetry, context.globalState);
```
RMacfarlane, line 83: I'm not sure what the consequence the insiders and stable versions of the extension having different names here is, should they map to the same service?

sana-ajani: cool, we confirmed with Julie and Amy about showing this only in the stable version of the extension

sana-ajani:
Looks good for the most part! Only other thing I think we need is to add the GDPR classification for the "query-expfeature" telemetry event (like we do in the remote-containers extension: https://github.com/microsoft/vscode-remote-containers/blob/9c5f61a5ff6c390e148c237ab6b0e110f204f015/src/utils/telemetry.ts#L9).

RMacfarlane:
Thanks! Added that classification

PR #2522, opened 2021-02-23
Improves perceived performance of PR loading

eamodio:
Adds preload command which can be triggered by Codespaces to put the PR extension in a "pr" loading state (e.g. in the correct view with a loading welcome message).

Also defers a lot of activation code until post-activation to improve perceived performance, since views will be show much quicker in a loading state.

### DIFF
--- a/package.json
+++ b/package.json
@@ -789,17 +789,17 @@
       {
         "view": "pr:github",
         "when": "!github:initialized",
-        "contents": "Loading pull requests..."
+        "contents": "Loading..."
       },
       {
         "view": "issues:github",
         "when": "!github:initialized",
-        "contents": "Loading issues..."
+        "contents": "Loading..."
       },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -774,6 +785,21 @@
         "view": "github:compareChanges",
         "when": "github:noUpstream",
         "contents": "The comparing branch has no upstream remote\n[Publish branch](command:git.publish)"
+      },
+      {
+        "view": "pr:github",
+        "when": "!github:initialized",
+        "contents": "Loading pull requests..."
```
RMacfarlane, line 792: Actually, there's a point later when we are actually loading pull requests where we display "Loading...", so I think this text might be better as "Initializing..."

RMacfarlane (approved):
LGTM!

PR #2530, opened 2021-02-25
Add an option to configure the default handling of terminal links

icatalina:
This PR introduces a new setting `githubPullRequests.terminalLinks.default` which allows to pre-configure the handling o terminal links for all links in the terminal.

3 possible options:

- `github`: Opens the links directly in github. Equivalent to pressing `no`.
- `vscode`: Opens the links directly in vscode. Equivalent to pressing `yes`.
- `ask`: Prompt the user every time they click a link.

### DIFF
--- a/package.json
+++ b/package.json
@@ -193,7 +193,7 @@
 					"default": true,
 					"description": "When true, the option to delete the remote will be selected by default when deleting a branch from a pull request."
 				},
-				"githubPullRequests.terminalLinks.default": {
+				"githubPullRequests.terminalLinksHandler": {
 					"type": "string",
 					"enum": [
 						"github",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -193,6 +193,21 @@
...
 				},
+				"githubPullRequests.terminalLinks.default": {
+					"type": "string",
+					"enum": [
+						"github",
+						"vscode",
+						"ask"
+					],
+					"enumDescriptions": [
+						"Create the PR in github",
```
RMacfarlane, line 204: To be consistent with the rest of the UI and other settings, I think these descriptions should be:
```
Create the pull request on GitHub
Create the pull request in VS Code
Ask which of the above methods to use
```

RMacfarlane (changes requested):
Thanks, this looks good! I think the setting descriptions should be tweaked a bit and then this will be ready to merge.

package.json
icatalina: Sure thing, I'll update them.

RMacfarlane (approved):
I went ahead and updated them so that this can be merged, thank you for this PR!

icatalina:
Thanks, sorry it slipped my mind 🙇

PR #2542, opened 2021-02-26
Using ${user} twice in query fails

### LINKED ISSUE #2533: Using ${user} twice in query fails
- Extension version: 0.22.0
- VSCode Version: 1.52.1
- OS: Windows_NT x64 10.0.17763

Steps to Reproduce:

1. Create the query in settings.json `githubPullRequests.queries`:
    ```
    {
         "label": "Reviewed by me",
         "query": "reviewed-by:${user} -author:${user}"
    }
    ```
2. Get the error:

    > Fetching pull requests failed: The listed users cannot be searched either because the users do not exist or you do not have permission to view the users.

Additional context:
- The query does not fail if I replace the second `${user}` with my literal username
- Looks like this was reproduced before: https://github.com/microsoft/vscode-pull-request-github/issues/2144#issuecomment-705471757

RMacfarlane:
Fixes #2533

### DIFF
--- a/src/github/githubRepository.ts
+++ b/src/github/githubRepository.ts
@@ -32,6 +32,7 @@ import { PullRequestModel } from './pullRequestModel';
 import defaultSchema from './queries.gql';
 import {
 	convertRESTPullRequestToRawPullRequest,
+	getPRFetchQuery,
 	parseGraphQLIssue,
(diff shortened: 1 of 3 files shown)


alexr00 (approved):
LGTM

PR #2543, opened 2021-02-27
Fix some tests and fail the build if the tests fail

RMacfarlane:
The tests were failing when run from the cmd line, and there were a few of them that were also outright failing. I also renamed some deprecated functions, and added a flag so that if the tests fail in CI, the CI will fail


alexr00 (approved):
🎉

PR #2552, opened 2021-03-02
Allow picking compare branch in create pull request webview

### LINKED ISSUE #2270: Create pull request command doesn't allow picking the source
It always assume you want to create the PR for the current branch, and I wanted to create a PR for a not-current branch.

RMacfarlane:
Fixes #2270
Screen Shot 2021-03-02 at 3 05 54 PM


src/view/compareChangesTreeDataProvider.ts
```diff
@@ -21,91 +22,129 @@ export class CompareChangesTreeProvider implements vscode.TreeDataProvider<TreeN
...
+				return new GitHubFileChangeNode(
+					this._view,
+					file.filename,
+					file.previous_filename,
+					getGitChangeType(file.status),
+					data.merge_base_commit.sha,
+					this.compareBranchName,
+				);
+			});
+		} catch (e) {
```
alexr00, line 144: Is it expected that this ever happen, or does something need to go seriously wrong to hit this catch? If the former, it might be helpful to include a notification about how it can be fixed.

alexr00 (approved):

src/view/compareChangesTreeDataProvider.ts
RMacfarlane: This isn't expected to happen, there's not really an action to take to fix it. Changing to use `Logger`, though

PR #2600, opened 2021-03-22
Resolves timing issue with RemoteHub 🤞

eamodio:
Hopefully resolves a timing issue with RemoteHub where no repositories are detected

This avoids any awaits between setting up the folders and review managers, and the git events.


RMacfarlane (approved):
LGTM!

PR #2601, opened 2021-03-22
Diff from PR's file gets erroneously opened when working on the PR's branch

### LINKED ISSUE #2295: Diff from PR's file gets erroneously opened when working on the PR's branch
I've had `jackson/proposed-api-management-command` checked out for a while, which has a PR up for it but I haven't been actively using the GHPR extension. However, I still see that a diff for `main.js` keeps opening at strange times

1. Checkout `Microsoft/vscode@jackson/proposed-api-management-command`
2. Close all editors
2. Reload
2. Wait a bit
3. This pops up:
image

reproduces in github.vscode-pull-request-github@0.21.0 and the latest nightly

RMacfarlane:
Fixes #2295

Scopes down the diff opening to codespaces - do you think this is reasonable? There have already been a couple of issues about this since turning on "Focused Mode" so I think it would be good to open the diff less frequently

### DIFF
--- a/src/view/reviewManager.ts
+++ b/src/view/reviewManager.ts
@@ -363,7 +363,7 @@ export class ReviewManager {
 		Logger.appendLine(`Review> display pull request status bar indicator and refresh pull request tree view.`);
 		this.statusBarItem.show();
 		vscode.commands.executeCommand('pr.refreshList');
-		if (!silent && this._context.workspaceState.get(FOCUS_REVIEW_MODE)) {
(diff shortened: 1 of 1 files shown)


eamodio (approved):
Change looks fine to me, but ultimately, I wonder if we should just add that as a setting? As I don't think I want that behavior ever 😄

RMacfarlane:
Yeah, I think having a setting makes sense. I merged this change in for now but will create an issue to revisit this

Tyriar:
Thanks for the fix @RMacfarlane

PR #2602, opened 2021-03-22
Create experiment service on activation

RMacfarlane:
Create the experimentation service and send a request for our A/A experiment to make sure everything is setup correctly

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -16,7 +16,7 @@ import { Resource } from './common/resources';
 import { handler as uriHandler } from './common/uri';
 import { onceEvent } from './common/utils';
 import { EXTENSION_ID, FOCUS_REVIEW_MODE } from './constants';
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -180,6 +180,9 @@ async function init(
 		"startup" : {}
 	*/
 	telemetry.sendTelemetryEvent('startup');
+
+	const experimentationService = await createExperimentationService(context, telemetry);
+	experimentationService.isCachedFlightEnabled('githubaa');
```
eamodio, line 185: Is that string correct? (Just asking because it seems an odd one)

eamodio (approved):
If that string is fine, it LGTM

src/extension.ts
RMacfarlane: yeah, good question. this is based on the scorecard we created so is correct, but I think for an actual experiment we would create a more meaningful name

PR #2609, opened 2021-03-23
Fix default branch selection if current branch doesn't have a remote

### LINKED ISSUE #2608: do not auto select a default branch if current branch doesn't have a remote
Testing microsoft/vscode#119527

- create a new branch without pushing it to a remote (e.g. "aweinand/foo")
- make sure to be on that branch (shown in the status bar)
- in the "Create Pull Request" view create a new pull request

Observe: In the "Merge Change From" section the branch selector shows the first branch of the list of all branches ("TylerLeonhard/ignore-this") which is totally unrelated to the currently selected branch shown in the status bar.

This choice is confusing.

I suggest in this case to not select a branch in the branch selector and explain somewhere (e.g. as the placeholder text) that the current branch has not yet pushed to a remote.

RMacfarlane:
Fixes #2608

The branches for "source" should be branches on GitHub + the current branch if it's not yet published

### DIFF
--- a/src/github/createPRViewProvider.ts
+++ b/src/github/createPRViewProvider.ts
@@ -123,10 +123,12 @@ export class CreatePullRequestViewProvider extends WebviewViewBase implements vs
 		}
 
 		if (hasMultipleCommits) {
-			return this.compareBranch.name!;
+			return this.compareBranch.name ?? '';
 		} else {
-			return titleAndBodyFrom(await this._folderRepositoryManager.getTipCommitMessage(this.compareBranch.name!))
-				.title;
+			return this.compareBranch.name
+				? titleAndBodyFrom(await this._folderRepositoryManager.getTipCommitMessage(this.compareBranch.name))
+						.title
+				: '';
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/github/createPRViewProvider.ts
```diff
@@ -188,9 +188,11 @@ export class CreatePullRequestViewProvider extends WebviewViewBase implements vs
 		let branchesForCompare = branchesForRemote;
 		if (defaultCompareRemote.owner !== defaultBaseRemote.owner) {
 			branchesForCompare = await this._folderRepositoryManager.listBranches(defaultCompareRemote.owner, defaultCompareRemote.repositoryName);
-			if (!branchesForCompare.includes(this.compareBranch.name!)) {
-				branchesForCompare.push(this.compareBranch.name!);
-			}
+		}
+
+		if (!branchesForCompare.includes(this.compareBranch.name!)) {
```
alexr00, line 193: nit: is is worth having a check for the name to avoid the `!`? (here and in other places)

alexr00 (approved):

src/github/createPRViewProvider.ts
RMacfarlane: yeah, good point! added some additional checks for this

Issue #1793, opened 2020-05-12
Its possible connect to github enterprise


vaclavbenes:
Hi, its possible configure extension with different github. (Use of github enterprise)

GABeech:
It looks like they removed support for GHE with version 0.16. 

It took me a bit of digging, and it doesn't look like there was any notification that this support was going away - but I don't pay close attention to the issues here. 

I'm personally a bit confused about the reasoning (that they don't have resources to test against it). Especially since they appear to be using Octokit.rest which has first class support for GHE out of the box.

drayfar:
+1 I can no longer use this extension now that #1616 is in

deerawan:
I experienced this issue in `0.16`. As interim, we could install the previous version `0.15.0` in VS Code

![image](https://user-images.githubusercontent.com/1243921/81761833-55377500-950e-11ea-88d0-5151f2f2c04e.png)

dvsoukup:
I read the change log on this.  I can sympathize with the lack of hands to assist with development on the enterprise functionality.

That being said though, it's a no-go to do that.  A "solution" is not to just use a previous version (v0.15).  Many thousands of people whom use this extension are now effectively blocked from ever getting updates.  You rip the rug out from underneath them.

A nicer scenario would have been to say "hey, we can't continue to develop on enterprise connectivity.  But we'll leave in what we have in there now".  Let the versions continue to advance, but don't rip this out from folks.

You leave lots of folks scratching their heads, wasting hours trying to figure out why their tooling isn't working.  Only to eventually come here, and see this mysterious decision to pull the plug on GHE support.

I love this extension, 5-stars for sure.  But the decision to do this is of 1-star quality.

azinoviev:
> I love this extension, 5-stars for sure. But the decision to do this is of 1-star quality.

+1

I'm really upset about this decision. GitHub Enterprise integration was extremely useful feature.
If there's not enough resources to test it, it'd prefer to have it buggy, than not to have it at all.

Richard-Lynch:
+1, very unhappy to see this feature go, would love to know if there’s any effort which can be made by the community to get it back

alexanderkjeldaas:
@Richard-Lynch fork?

JunhwanPark:
I'm waiting for Github Enterprise support..............

khepin:
Locked out as well. Don't have the JS chops to contribute the required code sadly.

danecreekphotography:
[...]

PR #2615, opened 2021-03-24
Fix fetch header API to detect GitHub

kabel:


catskul:
Is this related to? https://github.com/microsoft/vscode-pull-request-github/issues/1793

Issue #2288, opened 2020-11-12
Use `vscode.open` and `vscode.diff` in file related tree commands


alexr00:
When `vscode.open` and `vscode.diff` are used as the command for `TreeItem`s VS Code will handle the open or diff as it does in the built in file explorer (i.e. the `TextDocumentShowOptions` which can be passed to those commands will be filled in correctly).

Depends on https://github.com/microsoft/vscode/issues/110498

alexr00:
I did a find all references on [command](https://github.com/Microsoft/vscode/blob/16cff9cff93539b87a0a16ccc8dd09ddf542e7d5/src/vs/vscode.d.ts#L9193-L9194) and I think we have updated them all!

PR #2625, opened 2021-03-30
Use `vscode.open` and `vscode.diff` in file related tree commands

alexr00:
The focus of this change is the command "pr.openDiffView".

Part of #2288

### DIFF
--- a/.eslintrc.base.json
+++ b/.eslintrc.base.json
@@ -23,7 +23,6 @@
 	"reportUnusedDisableDirectives": true,
 	"root": true,
 	"rules": {
-		"linebreak-style": ["error", "unix"],
 		"new-parens": "error",
 		"no-async-promise-executor": "off",
 		"no-console": "off",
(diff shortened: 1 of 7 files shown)


.eslintrc.base.json
```diff
@@ -23,7 +23,6 @@
 	"reportUnusedDisableDirectives": true,
 	"root": true,
 	"rules": {
-		"linebreak-style": ["error", "unix"],
```
alexr00, line 26: @eamodio just a heads up about removing this rule since it isn't cross-plat friendly.

src/view/treeNodes/pullRequestNode.ts
```diff
@@ -280,6 +280,7 @@ export class PRNode extends TreeNode implements CommentHandler, vscode.Commentin
 			const parentFileName = change.status === GitChangeType.RENAME ? change.previousFileName! : change.fileName;
 			if (change instanceof SlimFileChange) {
 				return new RemoteFileChangeNode(
```
alexr00, line 282: @RMacfarlane from what I understand a `SlimFileChange` resulting in a `RemoteFileChangeNode` means that the file is not present locally and opening a diff won't do anything. So the command that was set in `RemoteFileChangeNode` is wrong.

I tested it out, and the current behavior when I try to open a diff of a jpg file that doesn't exist locally I just get a slient failure.

With my change, there will be an error. 

What should the behavior here be? Show a notification that says the we can't open this file?

RMacfarlane: Yeah, I think showing a notification makes sense - we could have a button to open the file on GitHub instead

RMacfarlane (approved):

PR #2626, opened 2021-03-30
Adopt review threads in comment controllers

RMacfarlane:
A couple of things to call out about this:
- Previously, we got comment position information that was about the line it was in the diff/patch, and not the actual line in the file. The review thread data we get has the line in the file, so the amount of position mapping we need to do is much less - the only concern is if a PR is checked out and there are local changes to the file that has comments in it.
- The review comment controller used to manage comment threads on 'review', 'file', and 'pr' scheme files when a pr was checked out. Now, the pull request node always manages 'pr' threads (which are the files that are actually under that node) regardless of if the pr is checked out or not. This is because they now listen for updates to the same cache, instead of having separate caches.
- There's now a separate command for the "Add comment" button that's meant to immediately add a single comment. We used to check whether there was an existing review, and then either add a comment to that, or call the REST API for adding a single comment. Adding a review thread in the graphql API always creates a review or adds to an existing one, I couldn't see a way to just immediately submit the thread in one step. So now we create the review and immediately submit it - using the old API for submitting a single comment was too painful since this uses the old diff-based positioning.  

Some things that I will address separately:
- File decorations for commenting don't update, these should also read the new review thread cache and listen for updates
- Similarly, the description page should listen for thread updates as well


src/view/treeNodes/pullRequestNode.ts
```diff
@@ -346,96 +284,175 @@ export class PRNode extends TreeNode implements CommentHandler, vscode.Commentin
...
+		});
 
-				if (!fileChange || fileChange instanceof RemoteFileChangeNode) {
-					return;
+		if (added.length) {
+			const reviewThreads = this.pullRequestModel.reviewThreadsCache;
+			const threadsByPath = groupBy(reviewThreads, thread => thread.path);
+			added.forEach(editor => {
+				const { isBase, fileName } = fromPRUri(editor.document.uri);
+				if (threadsByPath[fileName]) {
```
alexr00, line 362: Is this the same as line 317 above? Might be nice to factor it into a sepearte function.

src/view/reviewManager.ts
```diff
@@ -482,6 +482,7 @@ export class ReviewManager {
 	private async getPullRequestData(pr: PullRequestModel & IResolvedPullRequestModel): Promise<void> {
 		try {
 			this._comments = await pr.getReviewComments();
+			await pr.getReviewThreads();
```
alexr00, line 485: Value not saved anywhere?

alexr00 (approved):
I'm not that familiar with the commenting code, so I don't feel like I was able to give this a super in depth review. Looks good though!

src/view/treeNodes/pullRequestNode.ts
RMacfarlane: Thanks, moved!

src/view/reviewManager.ts
RMacfarlane: Changed this to use a new `initializeReviewThreadsCache` function to hopefully make its purpose clearer

PR #2645, opened 2021-04-08
fix: Spelling mistakes

ffflorian:
There were a few spelling mistakes which I noticed, so I started searching for more :)

Note that there are a few breaking changes because some constants and functions had typos, too.


RMacfarlane (approved):
Thank you!

PR #2658, opened 2021-04-16
Change commented icon to speech bubble

### LINKED ISSUE #2589: The comment icon in the TreeView is unclear and tiny
In the treeview, the comment icon with a diamond is unclear that there are comments and quite tiny. Maybe a an icon like this would be better?
image

What it looks like currently:
image

RMacfarlane:
Fixes #2589


eamodio (approved):
LGTM!

PR #2668, opened 2021-04-21
Remove reading of package.json and product.json for setting up experimentation service

RMacfarlane:
Use vscode apis instead of reading these files.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -184,13 +184,14 @@ async function init(
 
 	await vscode.commands.executeCommand('setContext', 'github:initialized', true);
 
+	const experimentationService = await createExperimentationService(context, telemetry);
+	await experimentationService.initializePromise;
+	experimentationService.isCachedFlightEnabled('githubaa');
+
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -184,13 +184,14 @@ async function init(
 
 	await vscode.commands.executeCommand('setContext', 'github:initialized', true);
 
+	const experimentationService = await createExperimentationService(context, telemetry);
+	await experimentationService.initializePromise;
+	experimentationService.isCachedFlightEnabled('githubaa');
```
alexr00, line 189: This method call looks like it returns something that should be read. I guess it also has side effects though and the return value doesn't matter?

alexr00 (approved):

src/extension.ts
RMacfarlane: Yeah, that's right - normally we would use the return value to turn a feature on or off, but doing this also updates info about what flight a user is in so just tests that everything is set up correctly

PR #2698, opened 2021-04-29
Clicking on a file in a commit with outdated comments fails

### LINKED ISSUE #2697: Clicking on a file in a commit with outdated comments fails
1. Open a clone of microsoft/vscode
2. Checkout this PR: https://github.com/microsoft/vscode/pull/122239
3. In the Commits section of the Changes view, expand the commit "address pr comments"
4. Click on vscode.proposed.d.ts
5. Nothing appears to happen. If you open the dev tools you'll see an error.

alexr00:
Fixes #2697

### DIFF
--- a/src/view/treeNodes/fileChangeNode.ts
+++ b/src/view/treeNodes/fileChangeNode.ts
@@ -6,8 +6,7 @@
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { IComment, ViewedState } from '../../common/comment';
-import { DiffChangeType, DiffHunk } from '../../common/diffHunk';
-import { getDiffLineByPosition, getZeroBased } from '../../common/diffPositionMapping';
+import { DiffHunk } from '../../common/diffHunk';
 import { GitChangeType } from '../../common/file';
 import { asImageDataURI, EMPTY_IMAGE_URI, fromReviewUri, ReviewUriParams, toResourceUri } from '../../common/uri';
 import { groupBy } from '../../common/utils';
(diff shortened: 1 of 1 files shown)


alexr00:
@RMacfarlane, I really just copied some code that used to be there back in. I'm not sure that this is the correct fix, but it does cause clicking on the file to do what I expect it to.

src/view/treeNodes/fileChangeNode.ts
```diff
@@ -370,7 +371,16 @@ export class GitFileChangeNode extends FileChangeNode implements vscode.TreeItem
 			.sort((a, b) => a.line - b.line);
 
 		if (reviewThreadsForNode.length) {
-			options.selection = new vscode.Range(reviewThreadsForNode[0].line, 0, reviewThreadsForNode[0].line, 0);
```
RMacfarlane, line 373: it looks like I used the wrong property here - I wasn't paying attention to the fact that these are outdated review threads, so instead of `line` I think using `originalLine` would also fix this

Though I think there's still a bit of funkiness to whether comments are displayed on commits. This seems to work from the comments panel but not the tree. This should probably be filtered to not just outdated comments, but comments for the specific commit. I think that can be handled separately though

RMacfarlane (approved):

src/view/treeNodes/fileChangeNode.ts
alexr00: Yep, `originalLine` also fixes this. I've removed the re-added code and switched to just use `originalLIne`.

PR #2730, opened 2021-05-14
Fix unexpected behavior when deleting PR branch

### LINKED ISSUE #1838: Branch name is UNKNOWN after merging PR and deleting branch
image

- Extension version: 0.16.0
- VSCode Version: 1.45.1
- OS: MacOS 10.15.5

Steps to Reproduce:

1. Create a PR using the plugin
2. Merge the PR
3. select options to delete both the branches (remote and local)
4. Click on refresh button

The branch name is incorrect (`UNKNOWN`) after merging a PR and refreshing using the refresh button.

gerryhernandez:
This fixes #1838 and also fixes some idiosyncrasies  with the "delete branch" button.

With this change, the correct branch name will now be displayed in the PR header view, even if it was previously deleted. Furthermore, if both the local and remote heads are deleted, the button will no longer be rendered. When the button is clicked, branches that don't exist will no longer show up in the quick-pick menu. For example, if the remote branch is already deleted, clicking the "delete branch" button will now only show the local branch in the quick-pick menu.


RMacfarlane (approved):
Thank you, this change looks good!

PR #2731, opened 2021-05-15
Pull selected branch only when checking out PR

### LINKED ISSUE #2030: Don't fetch every branch in contributor's remote when checking out a pull request
Thanks for this very nice extension!!  It is really nice to review code in my code editor:)

I noticed that when I clicked the blue "Checkout" button in the pull request tab, the following happens:
1) The extension adds a remote named after the GitHub username of the pull-request author (this remote is *not* deleted when I "Exit review mode")
2) This new remote is fetched (this creates remote-tracking branches for *all* branches in the contributor's remote)
3) The the source branch of the pull request is checked out (this branch is *not* deleted when I "Exit review mode")

I don't like 2), it pollutes the output of `git branch -r`. 
I think the extension should only fetch the branch of the pull request, i.e. 
```bash
git fetch $PR_AUTHOR_REMOTE_URL $PR_BRANCH:refs/heads/pr/$AUTHOR_GITHUB_USERNAME/$PR_NUMBER
```

 In fact, adding a remote is not even necessary as Git can fetch a branch directly into `FETCH_HEAD`, and we can the checkout `FETCH_HEAD`in detached HEAD. 

I understand the current behaviour might be convenient for some, but maybe it could be configurable ?

cdce8p:
Change pull behavior during unshallowing step to only pull and unshallow selected branch.
This PR requires an update to the VS Code Git extension API: https://github.com/microsoft/vscode/pull/123954

Fixes #2030

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
 	"version": "0.26.0",
 	"publisher": "GitHub",
 	"engines": {
-		"vscode": "^1.55.0"
+		"vscode": "^1.57.0"
 	},
 	"categories": [
 		"Other"
--- a/src/api/api.d.ts
+++ b/src/api/api.d.ts
@@ -122,6 +122,12 @@ export interface FetchOptions {
 	depth?: number;
 }
 
+export interface PullOptions {
+	remote?: string;
+	ref?: string;
+	unshallow?: boolean;
+}
+
 export interface BranchQuery {
 	readonly remote?: boolean;
 	readonly pattern?: string;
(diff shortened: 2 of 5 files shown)


iamtangram:
asdasdas

cdce8p:
I no longer need this as I've built a custom script using Github CLI to do the checkout. That works perfectly fine.
If anyone wants to reuse this PR or parts of it, please feel free to do so.

Issue #2741, opened 2021-05-20
big pull request not loading


RobertGalatNordic:
Issue Type: <b>Bug</b>

I have a big pull request (>100 files changed) and when I click on `>` under Waiting For My Review,  to see the changes, it keeps loading, and nothing happens. I can checkout the PR, but I can not see the files changed, nor start a review. 

Extension version: 0.26.0
VS Code version: Code 1.56.2 (054a9295330880ed74ceaedda236253b4f39a335, 2021-05-12T16:45:26.313Z)
OS version: Linux x64 5.8.0-53-generic snap

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz (8 x 4199)|
|GPU Status|2d_canvas: unavailable_software<br>gpu_compositing: disabled_software<br>multiple_raster_threads: enabled_on<br>oop_rasterization: disabled_off<br>opengl: disabled_off<br>rasterization: disabled_software<br>skia_renderer: enabled_on<br>video_decode: disabled_software<br>vulkan: disabled_off<br>webgl: unavailable_software<br>webgl2: unavailable_software|
|Load (avg)|2, 3, 2|
|Memory (System)|15.46GB (1.36GB free)|
|Process Argv|--force-user-env --no-sandbox --unity-launch --no-sandbox --crash-reporter-id 8f4baa7d-87f6-456d-9200-7180f882bc82|
|Screen Reader|no|
|VM|0%|
|DESKTOP_SESSION|ubuntu|
|XDG_CURRENT_DESKTOP|Unity|
|XDG_SESSION_DESKTOP|ubuntu|
|XDG_SESSION_TYPE|x11|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368cf:30146710
vsreu685:30147344
python383cf:30185419
pythonvspyt700cf:30270857
pythonvspyt602:30300191
vspor879:30202332
vspor708:30202333
vspor363:30204092
vstes627:30244334
pythonvspyt639:30300192
pythontb:30283811
vspre833:30267464
pythonptprofiler:30281270
vsdfh931cf:30280410
vshan820:30294714
vscorecescf:30290706
pythondataviewer:30285071
vscus158:30286553
vscgsv2:30301612
vscorehov:30302759
bridgeflightcf:30302070
vscod805cf:30301675

```

</details>

RMacfarlane:
[...]

PR #2766, opened 2021-06-10
Fix broken pull request files pagination

habibm:
Fixes https://github.com/microsoft/vscode-pull-request-github/issues/2741.

Pagination was broken causing infinite loop of paging requests for PRs with more than 100 files.


RobertGalatNordic (approved):

alexr00 (approved):
Thanks for the fix!

PR #2838, opened 2021-07-15
Send refresh event to the treeview

### LINKED ISSUE #2299: Edit PR title, does not (auto) update title of PR in treeview (PR in list on left side)
- Extension version:v0.21.2
- VSCode Version:1.51
- OS:Windows

Steps to Reproduce:
1. Open Pull request
2. Go to PR description.
3. Edit title of pr.
4. Wait for sync? nothing happens
5. Refresh PR: still the PR in list has old title.
6. Solution: Refresh the PR list from its treeview control. (I would expect it to happen when title is edited automatically)

sumitani:
Sending the event to refresh the treeview
Fixes #2299

### DIFF
--- a/src/github/issueModel.ts
+++ b/src/github/issueModel.ts
@@ -146,6 +146,8 @@ export class IssueModel<TItem extends Issue = Issue> {
 				},
 			});
 
+			vscode.commands.executeCommand('pr.refreshList');
+
 			return data!.updatePullRequest.pullRequest;
 		} catch (e) {
 			throw new Error(formatError(e));
--- a/src/github/pullRequestOverview.ts
+++ b/src/github/pullRequestOverview.ts
@@ -191,6 +191,8 @@ export class PullRequestOverviewPanel extends IssueOverviewPanel<PullRequestMode
 
 				const continueOnGitHub = isCrossRepository && isInCodespaces();
 
+				vscode.commands.executeCommand('pr.refreshList');
+
 				Logger.debug('pr.initialize', PullRequestOverviewPanel.ID);
 				this._postMessage({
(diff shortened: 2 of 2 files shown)


alexr00:
Thank you for the PR! I am not a fan of calling `vscode.commands.executeCommand` to execute code paths that the extension itself owns since it makes refactoring difficult and obscures where a call is coming from. I recognize that this extension already does this in many places, but some day I want to delete all of them :). Because of this, I don't want to accept any PRs that do this since it is just more debt to tackle later.

sumitani:
@alexr00 I agree, I was worried about the architectural issues. Is there a way to do this in a cleaner way than `vscode.commands.executeCommand` that I can use as a reference? (other parts of the code that you consider better than this approach).

alexr00:
@sumitani thank you for understanding 😌. I don't think any of the objects that you can currently access from the places the "refresh" needs to happen provide you with access to a method that will refresh the tree. 
Here's the method that should refresh the tree: 
https://github.com/microsoft/vscode-pull-request-github/blob/152a77f780214a50845aaea9c17c76dff9366c89/src/view/prsTreeDataProvider.ts#L133-L134

I'm fairly confident that we'll need an event in [`IssueModel`](https://github.com/microsoft/vscode-pull-request-github/blob/a1dc3abf08bd755dd1b3388e4aa4a9703102ee33/src/github/issueModel.ts#L23-L24) for when a property, such as the title has changed. (and if you want to make a PR for just this part I would accept it for review)

Beyond that, I'm less sure what the best direction is. There are a lot of places where that same command is executed, and all of them should be replaced. I'd need to spend a lot more time deciding how to refactor before making any further changes.

PR #2842, opened 2021-07-18
chore: replace deprecated vscode-test with @vscode/test-electron

iChenLei:
vscode-test be marked as [deprecated](https://www.npmjs.com/package/vscode-test) on npmjs.org(yarnpkg.com), so just replace vscode-test with latest @vscode/test-electron for `vscode-pull-request-github`'s CI. cc @alexr00 @RMacfarlane


alexr00 (approved):
Thank you!

Issue #2350, opened 2021-01-11
PR description title hover icons shift the title


Tyriar:
If going with this show on hover style, maybe show the buttons as an overlay with opaque background on top of the title title, instead of causing the reflow:

![recording (24)](https://user-images.githubusercontent.com/2193314/104222197-ceb59900-53f6-11eb-819c-dfaa1e2b4981.gif)

pnavarrc:
I encountered this bug today, maybe moving the issue number to the left would work too. The problem is that is impossible to click on the PR number as it will move up and down depending on what the user is hovering!

burkeholland:
I'm not seeing this behavior in  0.29.0....


https://user-images.githubusercontent.com/686963/129456222-5cd7298e-ba75-4692-8946-e3369aa0c0e3.mp4

PR #2849, opened 2021-07-19
Always show action bar in PR overview header

### LINKED ISSUE #2515: Make edit and copy buttons beside the PR name always visible
These buttons only show on however and aren't discoverable outside of hover. I'm not sure how intuitive it is to hover over that area to discover the buttons.

On hover buttons: 
image

Regular:
image

joshuaobrien:
Fixes #2515 and #2350


alexr00 (approved):
Looks good, thank you!

PR #2887, opened 2021-08-05
updating README since Github Enterprise is now supported

bhatvikrant:


alexr00:
Thank you for the PR, but we have now updated the readme!

PR #2909, opened 2021-08-13
Fixes #2259

### LINKED ISSUE #2259: issueBranchTitle not Working as Expected
No matter what I do I cannot get this pluging to accept and use this formatting in the devcontainer.json (and all other settings.json) **if the branch was previously created with the default format**

"githubIssues.issueBranchTitle": "${user}-issue${issueNumber}",

The branch is always created in the default format. Even if I delete the "bad" branch. Is there metadata for this and how do I delete it?

It seems like once you've created a branch with any name then the new settings are impossible to take effect on a branch that was previously created. It seems like some metadata is stored somewhere? Is that deletable?

burkeholland:
Check to see if new branch name per config is different from what is in state. If so, and branch has been deleted, update branch name to new config calculated value.

### DIFF
--- a/src/issues/currentIssue.ts
+++ b/src/issues/currentIssue.ts
@@ -123,19 +123,6 @@ export class CurrentIssue {
 		return false;
 	}
 
-	private async renameBranch(oldBranch: string | undefined, newBranch: string): Promise<Boolean> {
-		try {
-			if (oldBranch && await this.branchExists(oldBranch)) {
-				await vscode.commands.executeCommand('git.renameBranch', [newBranch]);
-			}
-		} catch (e) {
-			vscode.window.showErrorMessage(
-				`Unable to rename branch ${oldBranch}. Git error: ${e.error}`,
-			);
-		}
-		return true;
-	}
-
 	private async createOrCheckoutBranch(branch: string): Promise<boolean> {
 		try {
 			if (await this.branchExists(branch)) {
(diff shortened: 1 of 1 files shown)


src/issues/currentIssue.ts
```diff
@@ -201,13 +201,19 @@ export class CurrentIssue {
...
+			undefined,
+			await this.getUser(),
+		);
+		if (this._branchName !== branchNameConfig) {
+			const branchExists = await this.branchExists(this._branchName!);
+			if (!branchExists) {
+				this._branchName = branchNameConfig;
+			}
+		}
 		if (!this._branchName) {
```
alexr00, line 216: We also need to do this `if` if `this._branchName` changed in the previous block since it includes branch name validation.

alexr00 (changes requested):
Just one change needed I think! I still think we should also offer to "move to"/"rename" the old branch if it exists, but that can be a separate PR since this is a good change regardless.

src/issues/currentIssue.ts
alexr00: @burkeholland this is the only change I see that is needed before merging!

alexr00: I was thinking something like this, since we still needed to validate the branch name:

https://github.com/microsoft/vscode-pull-request-github/pull/2909/files#diff-e0d767b263cbb178ca6cf7902e1934d49d08810ea2f4c26d97d23b98e5643210R231-R236

alexr00 (approved):
I pushed the change, it all looks good. Thank you!

PR #2929, opened 2021-08-21
burkeholland-issue2740

### LINKED ISSUE #2740: Extension does not respect remote link type for upstream
- Extension version:0.26.0
- VSCode Version:1.56.2 x64 system install
- OS:10.0.19043.1021 x64 pro zh-CN

Steps to Reproduce:

1.Clone https://github.com/lychichem/winget-pkgs, which is forked from microsoft/winget-pkgs, by using ssh url, with this extension in VSCode
2.The upstream automatically added uses https but not ssh

This may be a small issue for developers outside mainland China, but it's a severe issue for developers in mainland China, as Github/Gitlab is not so robust under https (They suffer from TLS RST attack, especially when it is near June 4th). So developers in mainland China usually cannot use git with an https remote.

burkeholland:
Fixes #2740

Respect the protocol of existing remotes. Right now, just looking at the first remote so whatever the first one is, so will the upstream be.

### DIFF
--- a/src/github/folderRepositoryManager.ts
+++ b/src/github/folderRepositoryManager.ts
@@ -9,7 +9,7 @@ import type { Branch, Repository, UpstreamRef } from '../api/api';
 import { GitApiImpl, GitErrorCodes, RefType } from '../api/api1';
 import { GitHubManager } from '../authentication/githubServer';
 import Logger from '../common/logger';
-import { Protocol } from '../common/protocol';
+import { Protocol, ProtocolType } from '../common/protocol';
 import { parseRepositoryRemotes, Remote } from '../common/remote';
 import { ITelemetry } from '../common/telemetry';
 import { EventType, TimelineEvent } from '../common/timelineEvent';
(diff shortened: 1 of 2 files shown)


src/github/folderRepositoryManager.ts
```diff
@@ -553,7 +553,11 @@ export class FolderRepositoryManager implements vscode.Disposable {
 					const upstreamAvailable = !this.repository.state.remotes.some(remote => remote.name === 'upstream');
 					const remoteName = upstreamAvailable ? 'upstream' : metadata.parent.owner?.login;
 					if (remoteName) {
-						await this.repository.addRemote(remoteName, metadata.parent.clone_url);
+						// check the remotes to see what protocol is being used
+						const isSSH = this.repository.state.remotes[0].pushUrl?.startsWith('git');
```
alexr00, line 557: Can you do `this.githubRepositories[0].remote.gitProtocol.type === ProtocolType.SSH` instead of checking if the push url starts with `'git'`?

alexr00 (changes requested):

alexr00 (approved):
Looks good, will merge after the August release!

PR #2967, opened 2021-08-27
Focus border is used on assignees when it's not focused

### LINKED ISSUE #2956: Focus border is used on assignees when it's not focused
I thought this was focused:

image

This results in this weird state:

image

Normally this doesn't show in the UI as focused can only be on one item, for example:

image

Ext: v2021.8.35110

burkeholland:
Fixes #2956


alexr00 (approved):
Looks good, I'll merge it after the August release!

PR #2968, opened 2021-08-27
Fixes Commit hash should use a monospace font #2957

### LINKED ISSUE #2957: Commit hash should use a monospace font
Current:

image

.com

image

Consider using `font-family: var(--vscode-editor-font-family);`

Ext: v2021.8.35110

burkeholland:
Fixes #2957


alexr00 (changes requested):
I see some slightly off positioning/padding/sizing with this:
image

PR #2972, opened 2021-08-30
fix: make pending checks "show" button / link accessible

### LINKED ISSUE #2973: Pending checks "show" button/link is not accessible
- Extension version: 0.29.2 (latest)
- VSCode Version: 1.59.1
- OS: Darwin x64 19.6.0

Steps to Reproduce:

1. Create a PR using the extension
2. Wait for the Pull Request view to show up.
3. Hit tab to go through the elements

**Expected result**

The "Show" button will be focused at some point

**Actual result**

It doesn't.

**Why?**

Seems like it's because the button is actually an `a` tag without `href`.

https://user-images.githubusercontent.com/3723951/131308127-59cf6b41-5f95-408f-9042-bd982cb2aad1.mov

moshfeu:
Fix #2973

[More info](https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/mouse-and-keyboard-events/links/link-behavior/#:~:text=Do%20not%20use%20an%20anchor%20with%20an%20onclick%20event%20without%20an%20href%20element%20event)

A note: It would be better to use `button` element here and style it as link but this is a quick fix.

### DIFF
--- a/webviews/components/merge.tsx
+++ b/webviews/components/merge.tsx
@@ -52,7 +52,7 @@ const StatusChecks = ({ pr }: { pr: PullRequest }) => {
 				<div className="status-item">
 					<StateIcon state={status.state} />
 					<div>{getSummaryLabel(status.statuses)}</div>
-					<a aria-role="button" onClick={toggleDetails}>
+					<a href="javascript:void(0)" aria-role="button" onClick={toggleDetails}>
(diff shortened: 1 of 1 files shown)


alexr00 (approved):
Thanks for the fix! I will merge in the next couple days after we have made our August release.

Issue #2951, opened 2021-08-25
Pending check circle should be horizontally aligned with no conflicts tick


Tyriar:
Current:

![image](https://user-images.githubusercontent.com/2193314/130790463-96fbcade-a6bd-4e4d-9878-2e4ff242d1f6.png)

Proposed:

![image](https://user-images.githubusercontent.com/2193314/130790527-d10b3ec9-c859-45ab-83c0-16bb8d69f535.png)

Ext: v2021.8.35110

alexr00:
@burkeholland FYI in case you want to take this one.

burkeholland:
Modified the dot to match the official octicon "dot-16" which is horizontally aligned. Now the dots are not vertically aligned in details...

![image](https://user-images.githubusercontent.com/686963/131365932-d9e56bfb-fee1-4cc8-8965-ffabafbab82f.png)

Maybe we already have an issue opened. If not, I'll open one.

burkeholland:
Moved fix to #2971

Issue #2950, opened 2021-08-25
Tick SVG doesn't look right


Tyriar:
![image](https://user-images.githubusercontent.com/2193314/130789802-ef9d16c5-d269-44cf-a522-57eb56abcf60.png)

Using the actual SVG from github.com looks nicer:

![image](https://user-images.githubusercontent.com/2193314/130790291-aabcf645-6508-49be-88f8-29eda7e7715b.png)

cc @misolori 

Ext: v2021.8.35110

Tyriar:
It also doesn't align with the codicon:

![image](https://user-images.githubusercontent.com/2193314/130793198-2f90e4df-3abb-4c8c-bee9-c458e50c61c5.png)

alexr00:
@burkeholland FYI in case you want to take this one.

Tyriar:
Can fix that in https://github.com/microsoft/vscode-pull-request-github/issues/2951

PR #2975, opened 2021-08-30
Fixes Pending check circle should be horizontally aligned with no conflicts tick #2951

burkeholland:


burkeholland:
Fixed moved to #2950

PR #2980, opened 2021-09-01
GitHub permalinks: Deduplicate the line number if the range is one line.

lgarron:
This will generate URLs like:

`https://github.com/microsoft/vscode-pull-request-github/blob/39d333d9931d685d01078e27c4e6583eda55fc27/src/issues/util.ts#L495`

... instead of:

`https://github.com/microsoft/vscode-pull-request-github/blob/39d333d9931d685d01078e27c4e6583eda55fc27/src/issues/util.ts#L495-L495`

### DIFF
--- a/src/issues/util.ts
+++ b/src/issues/util.ts
@@ -497,7 +497,7 @@ export async function createGithubPermalink(
 			return '';
 		}
 		let hash = `#L${range.start.line + 1}`;
-		if (range.start.line === range.end.line) {
+		if (range.start.line !== range.end.line) {
 			hash += `-L${range.end.line + 1}`;
 		}
 		return hash;
--- a/src/issues/util.ts
+++ b/src/issues/util.ts
@@ -492,10 +492,19 @@ export async function createGithubPermalink(
 		}
 	}
 	const pathSegment = uri.path.substring(repository.rootUri.path.length);
-	const rangeString = range ? `#L${range.start.line + 1}-L${range.end.line + 1}` : '';
+	const rangeString = () => {
+		if (!range) {
+			return '';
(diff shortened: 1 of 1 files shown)


src/issues/util.ts
```diff
@@ -492,10 +492,19 @@ export async function createGithubPermalink(
 		}
 	}
 	const pathSegment = uri.path.substring(repository.rootUri.path.length);
-	const rangeString = range ? `#L${range.start.line + 1}-L${range.end.line + 1}` : '';
+	const rangeString = () => {
```
lgarron, line 495: This could theoretically be kept one one line, or shortened by conditionally updating a variable in the main parent block. However, I thought this would be the easiest to read, since the calculation is entirely inside block without side effects.

src/issues/util.ts
```diff
@@ -492,10 +492,19 @@ export async function createGithubPermalink(
...
+		if (!range) {
+			return '';
+		}
+		let hash = `#L${range.start.line + 1}`;
+		if (range.start.line === range.end.line) {
```
alexr00, line 500: Should be `!==` instead of `===`?

alexr00 (changes requested):
Thanks for the PR! Just one comment.

src/issues/util.ts
lgarron: 🤦‍♀️ That's even what I wrote above!
```suggestion
		if (range.start.line !== range.end.line) {
```

lgarron: Thanks!

alexr00 (approved):
Looks good, thank you!

PR #2993, opened 2021-09-10
Burkeholland-issue1635

### LINKED ISSUE #1635: Navigate from pull request query to settings json
Similar to https://github.com/microsoft/vscode-pull-request-github/issues/1622, have the same behavior for the queries in the PR tree

burkeholland:
Fixes #1635

### DIFF
--- a/package.json
+++ b/package.json
@@ -651,6 +651,12 @@
 				"icon": "$(collapse-all)"
 			},
 			{
+				"command": "pr.editQuery",
+				"title": "Edit Query",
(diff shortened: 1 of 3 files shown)


burkeholland:
I tried to do this as close to how it was done for the Issues, but the way the Issues TreeView is implemented seems quite different from the PR's. I'm guessing the PR's came first? In any event, this works the same way that the Issues edit query works if not implemented differently.

alexr00 (changes requested):

PR #3005, opened 2021-09-24
Skipped status check is shown as pending

### LINKED ISSUE #2580: Skipped status check is shown as pending
- Extension version: v0.24.0
- VSCode Version: 1.54.3
- OS: Windows 10

Steps to Reproduce:

1. Configure status checks for pull requests so that one of them would be skipped. For example, using the GitHub Actions and [`jobs.<job_id>.if'](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif).
1. Create a PR.
1. Wait for the checks to be executed.
1. Make sure that one of the checks was skipped: 
image
1. Open the same PR in VS Code and see that the same check is in the pending state: 
image

burkeholland:
Fixes #2580


alexr00 (approved):
Looks good, thank you!

PR #3008, opened 2021-09-27
Fixes #3007

### LINKED ISSUE #3007: Text not visible when PR check in pending state
Moreover, while one of the checks is really in pending state, I've caught the following strange situation:
image

_Originally posted by @anantakrishna in https://github.com/microsoft/vscode-pull-request-github/issues/2580#issuecomment-801997735_

burkeholland:


webviews/components/merge.tsx
```diff
@@ -360,7 +360,7 @@ const StatusCheckDetails = ({ statuses }: Partial<PullRequest['status']>) => (
 					<StateIcon state={s.state} />
 					<Avatar for={{ avatarUrl: s.avatar_url, url: s.url }} />
 					<span className="status-check-detail-text">
-						{s.context} {s.description ? `— ${s.description}` : ''}
+						{s.context}{s.description ? `${s.description}` : ''}
```
alexr00, line 363: Let's leave the dash in here.

webviews/components/user.tsx
```diff
@@ -12,7 +12,7 @@ export const Avatar = ({ for: author }: { for: Partial<PullRequest['author']> })
 		{author.avatarUrl ? (
 			<img className="avatar" src={author.avatarUrl} alt="" />
 		) : (
-			<Icon className="avatar-icon" src={require('../../resources/icons/dark/github.svg')} />
+			<img className="avatar" src='../github.png' />
```
alexr00, line 15: I don't actually see a `github.png` in this repo. Is this change needed?

src/github/pullRequestModel.ts
```diff
@@ -982,9 +982,9 @@ export class PullRequestModel extends IssueModel<PullRequest> implements IPullRe
 					return {
 						id: context.id,
 						url: context.targetUrl,
-						avatar_url: context.avatarUrl,
+						avatar_url: context.checkSuite.app?.logoUrl,
```
alexr00, line 985: Execution will only fall into this `else` if we have determined that the run is not a `CheckRun`, so at this point `context` is `any`. It seems though that we are assuming that really we do want pending runs to be treated as being a `CheckRun`. Would it make sense instead to change this:

https://github.com/burkeholland/vscode-pull-request-github/blob/fbff4e8582138f9272ac0be3f8a81eeb2294cc8a/src/github/graphql.ts#L593-L595

to be:

```ts
export function isCheckRun(x: CheckRun | StatusContext): x is CheckRun {
	return !!(x as CheckRun).checkSuite;
}
```
?

alexr00 (changes requested):
Just a few questions:

PR #3049, opened 2021-10-20
Fix typo

38elements:
I think "a GitHub permalink" is correct.


alexr00 (approved):

alexr00:
Thanks for the doc fix!

PR #3084, opened 2021-11-05
Fix generated branch name requiring quotes in bash

taisph:
If an issue name contains parenthesis characters, they are currently
added to the generated branch name, causing bash to complain about the
name if it is not quoted.

I.e. `git checkout test/some(other)branch`
bash: syntax error near unexpected token `('

Stripping the parenthesis characters from the branch name fixes this.


alexr00 (approved):
Looks good, thank you!

PR #3091, opened 2021-11-08
Move to File System Provider for better workbench editor integration

rebornix:
Fix microsoft/vscode#133892. When we started the project, the FS API is not rich enough yet and the `TextDocumentContentProvider` fits our need pretty well. However the challenge with `TextDocumentContentProvider` is it won't work with notebooks as it's not FS based, thus users can't open the change of a notebook in a rich notebook diff editor.

In this PR we expolored moving to FS provider completely in the GHPRI, just like what we did with Issue file.


src/view/readonlyFileSystemProvider.ts
```diff
@@ -0,0 +1,52 @@
...
+
+	readDirectory(_uri: vscode.Uri): [string, vscode.FileType][] {
+		return [];
+	}
+
+	createDirectory(_uri: vscode.Uri): void {
+		/** no op */
+	}
+
+	async readFile(_uri: vscode.Uri): Promise<Uint8Array> {
```
alexr00, line 37: Might be nice to make this class abstract and then make `readFile` abstract, since currently the `ReadonlyFileSystemProvider` isn't really usable on its own.

src/view/compareChangesTreeDataProvider.ts
```diff
@@ -173,8 +174,29 @@ export class CompareChangesTreeProvider implements vscode.TreeDataProvider<TreeN
...
+			path: params.fileName,
+			ref: params.branch,
+		});
+
+		const contents = (fileContent.data as any).content ?? '';
+		const buff = buffer.Buffer.from(contents, (fileContent.data as any).encoding);
+		return buff;
+	}
 
 	async provideTextDocumentContent(uri: vscode.Uri, _token: vscode.CancellationToken): Promise<string> {
```
alexr00, line 201: Looks like there are still references to `GitHubContentProvider` that expect it to `provideTextDocumentContent`, but I think to solve the notebook issue we need to replace all of those with the file system provider, correct?

alexr00 (changes requested):
Thanks for the good exploration! I took a quick look, and the idea makes sense. @rebornix, do you want to address my initial comments, or would you prefer that I take over this PR later?

alexr00 (approved):
I've made the suggested changes and confirmed that diffs from all file system providers still look good.

PR #3209, opened 2021-12-26
Modified Authentication module to infer port number from URI Schema

sheenjustin:
instead of hardcoding 443

### DIFF
--- a/src/authentication/configuration.ts
+++ b/src/authentication/configuration.ts
@@ -42,4 +42,15 @@ export const HostHelper = class {
 			return `/api/v3${path}`;
 		}
 	}
+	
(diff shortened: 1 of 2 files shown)


alexr00:
@sheenjustin can you describe the issue that this would fix?

src/authentication/configuration.ts
```diff
@@ -42,4 +42,15 @@ export const HostHelper = class {
 			return `/api/v3${path}`;
 		}
 	}
+	
+	// TODO: Add support for URIs with explicit port number e.g. {schema}://{domain}:{port}/{repo}
+	// for now just hardcode http = 80, https = 443
+	public static getHostPort(host: IHostConfiguration | vscode.Uri): number {
```
alianZhen, line 48: test

PR #3212, opened 2021-12-29
Fix react warnings in timeline

blindpirate:
Previously, there's a warning "Each child in a list should have
a unique key prop" in `Timeline.render`, because there's no `databaseId`
field in `CommitEvent`. In this case, we use `CommitEvent.id` instead.

Xnip2021-12-29_07-32-28

Xnip2021-12-29_07-52-05


alexr00 (approved):
Looks good, thank you for fixing!

PR #3214, opened 2021-12-30
Fix more React warnings of incorrect `props.key`

blindpirate:
There are some warnings in `CommentThread` because of misuse of
`props.key`: `CommentThread`/`Diff` had no `key` prop, and
`CommentThread` incorrectly forwareded `key` prop to child component.

Xnip2021-12-30_13-18-55

------

Update: found another missing `key prop` in `MergedEvent`:

image

Xnip2021-12-30_14-41-10

### DIFF
--- a/webviews/components/timeline.tsx
+++ b/webviews/components/timeline.tsx
@@ -132,12 +132,12 @@ const ReviewEventView = (event: ReviewEvent) => {
 	);
 };
 
-function CommentThread({ key, thread, eventId }: { key: string; thread: IComment[]; eventId: number }) {
+function CommentThread({ thread, eventId }: { thread: IComment[]; eventId: number }) {
 	const comment = thread[0];
 	const [revealed, setRevealed] = useState(!comment.isResolved);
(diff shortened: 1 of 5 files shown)


webviews/components/timeline.tsx
```diff
@@ -132,12 +132,12 @@ const ReviewEventView = (event: ReviewEvent) => {
 	);
 };
 
-function CommentThread({ key, thread, eventId }: { key: string; thread: IComment[]; eventId: number }) {
+function CommentThread({ thread, eventId }: { thread: IComment[]; eventId: number }) {
 	const comment = thread[0];
 	const [revealed, setRevealed] = useState(!comment.isResolved);
 	const { openDiff } = useContext(PullRequestContext);
 	return (
-		<div key={key} className="diff-container">
```
blindpirate, line 140: > special props (ref and key) are used by React, and are thus not forwarded to the component.

alexr00 (approved):
Thank you for the fixes!

Issue #2426, opened 2021-01-27
Visually out of place drop down in pull requests views with Safari 


mjbvz:
**Repro**
1. Using the latest nightly in a codespace loaded in Safari
2. Create a pull request

**Bug**
In the pull requests view, the drop downs are using the standard aqua-styling from safari:

<img width="404" alt="Screen Shot 2021-01-27 at 2 32 01 PM" src="https://user-images.githubusercontent.com/12821956/106063150-9c6d9200-60ac-11eb-9729-b1905eaf4725.png">

This looks out of place compared to our other dropdowns:

<img width="240" alt="Screen Shot 2021-01-27 at 2 32 33 PM" src="https://user-images.githubusercontent.com/12821956/106063209-ac857180-60ac-11eb-8fb3-7c9e1e5447db.png">

alexr00:
@burkeholland in case you want to work on this!

Yash621:
@mjbvz can I work on this ?

alexr00:
@Yash621 you are welcome to work on this!

Yash621:
@alexr00 can you guide me on this ,like how should I go about it ?

alexr00:
The dropdowns are here:

https://github.com/microsoft/vscode-pull-request-github/blob/b4bb8a6c63109e838ae54797ce608c61123f19b3/webviews/createPullRequestView/app.tsx#L12-L49

I haven't done any investigation into this issue, but I assume there is some styling that can be done to make the dropdowns look more like VS Code dropdowns. Styles are here: https://github.com/microsoft/vscode-pull-request-github/blob/09a489f8dd15d0a6c98e927879e863d234ebd68a/webviews/createPullRequestView/index.css

alexr00:
Looks like we lost some of the original styling (the arrow on the right of the select is gone):
![image](https://user-images.githubusercontent.com/38270282/148943252-a6d96546-5d68-43d4-8e60-9938db3b3f13.png)

The select now also has rounded corners, which isn't consistent with other selects in VS Code:
![image](https://user-images.githubusercontent.com/38270282/148943369-c0d95544-ae81-4cd1-834b-9466bca0a957.png)

The colors look good though. Do you know if there's a way to fix the lost styling?

saifahn:
@alexr00 

Hey there!

[...]

PR #3218, opened 2021-12-30
fixed Visually out of place drop down in pull requests views with Safari

Yash621:
@mjbvz  @alexr00 This PR is with respect to issue #2426 

I have fixed the styling of the drop downs ,please review it and let me know if any changes are required.

### DIFF
--- a/webviews/createPullRequestView/index.css
+++ b/webviews/createPullRequestView/index.css
@@ -85,3 +85,7 @@ body {
 .selector-group .wrapper {
 	padding-right: 5px;
 }
+.select-style {
+	background-color: #3d3c3d;
+	-webkit-appearance: none;
+}
--- a/webviews/createPullRequestView/index.css
+++ b/webviews/createPullRequestView/index.css
@@ -11,6 +11,9 @@ textarea {
 
 select {
 	width: 100%;
+	background-color: var(--vscode-dropdown-background);
+	color: var(--vscode-dropdown-foreground);
+	-webkit-appearance: none;
 }
 
 textarea {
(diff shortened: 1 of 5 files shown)


webviews/createPullRequestView/index.css
```diff
@@ -84,4 +84,8 @@ body {
 
 .selector-group .wrapper {
 	padding-right: 5px;
-}
\ No newline at end of file
+}
+.select-style {
+	background-color: #3d3c3d;
```
alexr00, line 89: Instead of hardcoding a value, we should use one of the VS Code theme colors so the solution works with all themes. The same one we use for the select options should be ok:

https://github.com/yash621/vscode-pull-request-github/blob/61b93b263b36f1b7d8c93096d439e3853aca7428/webviews/activityBarView/index.css#L115-L116

webviews/createPullRequestView/index.css
```diff
@@ -84,4 +84,8 @@ body {
 
 .selector-group .wrapper {
 	padding-right: 5px;
-}
\ No newline at end of file
+}
+.select-style {
```
alexr00, line 88: Can these styles be added to the `select` selector instead of making a new class?

https://github.com/yash621/vscode-pull-request-github/blob/9dabb657592114854a9402a21e0520e54cd98dd9/webviews/createPullRequestView/index.css#L12-L14

alexr00 (changes requested):
Thanks for the PR! In addition to the changes I've requested, can you share some screenshots and gifs of what the dropdowns look like with the changes across a couple themes?

Yash621:
@alexr00  I have done the changes suggested ,please review them ,though there was no script to run the project locally so i am unable to attach the screenshot

PR #3238, opened 2022-01-08
Add command to open all diffs

### LINKED ISSUE #3047: Feature Request: View All Changes In PR
I would find a command that allows me to open all changes in a pull request really valuable. Currently I open each file's diff from the PR Changes view. This is fine, but gets a bit tedious with larger pull request (O(N) in terms of my input). If I'm wanting to compare changes to the working tree GitLens has a `Open All Changes (difftool)` (O(1) in terms of my input) command. Which is fantastic as I can open each diff as a tab and work through the changes in the commit, closing tabs as I'm satisfied with the changes.

I would like to be able to be able to do something similar for my PR review process:
- Close All Editors
- Open All Changes in PR
- Review each file, adding comments, marking as reviewed, and finally closing that file when I'm satisfied.

I hope this isn't a duplicate. I see that the folks working on GitLens received a similar request, however, they see this project as more in scope for implementation. See https://github.com/Axosoft/vscode-gitlens/issues/1020

joshuaobrien:
This PR adds a command that opens all of the files touched by the active PR.

Closes #3047

### DIFF
--- a/package.json
+++ b/package.json
@@ -472,8 +472,8 @@
 				"icon": "$(globe)"
 			},
 			{
-				"command": "pr.openAllChanges",
-				"title": "Open All Changes",
+				"command": "pr.openAllDiffs",
+				"title": "Open All Diffs",
 				"category": "GitHub Pull Requests"
 			},
 			{
@@ -928,7 +928,7 @@
 					"when": "gitHubOpenRepositoryCount != 0 && github:inReviewMode"
 				},
 				{
-					"command": "pr.openAllChanges",
+					"command": "pr.openAllDiffs",
 					"when": "gitHubOpenRepositoryCount != 0 && github:inReviewMode"
 				},
 				{
--- a/package.json
+++ b/package.json
@@ -472,8 +472,8 @@
 				"icon": "$(globe)"
 			},
 			{
-				"command": "pr.openAllChangesLocally",
-				"title": "Open All Changes Locally",
+				"command": "pr.openAllChanges",
+				"title": "Open All Changes",
 				"category": "GitHub Pull Requests"
 			},
 			{
@@ -928,7 +928,7 @@
 					"when": "gitHubOpenRepositoryCount != 0 && github:inReviewMode"
 				},
 				{
-					"command": "pr.openAllChangesLocally",
+					"command": "pr.openAllChanges",
 					"when": "gitHubOpenRepositoryCount != 0 && github:inReviewMode"
 				},
 				{
--- a/package.json
+++ b/package.json
@@ -472,6 +472,11 @@
 				"icon": "$(globe)"
 			},
 			{
+				"command": "pr.openAllChangesLocally",
+				"title": "Open All Changes Locally",
+				"category": "GitHub Pull Requests"
+			},
+			{
 				"command": "pr.refreshPullRequest",
 				"title": "Refresh Pull Request",
 				"category": "GitHub Pull Requests"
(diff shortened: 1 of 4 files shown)


src/view/treeNodes/fileChangeNode.ts
```diff
@@ -253,8 +254,8 @@ export class FileChangeNode extends TreeNode implements vscode.TreeItem {
 		return this;
 	}
 
-	openFileCommand(): vscode.Command {
-		return openFileCommand(this.filePath);
+	openFileCommand(opts: Pick<vscode.TextDocumentShowOptions, 'preview'>): vscode.Command {
```
joshuaobrien, line 257: When we're opening all of the files, we likely don't want to open them as a preview. In order to support this, callers can now specify whether or not files should open as a preview.

src/view/treeNodes/fileChangeNode.ts
```diff
@@ -17,11 +17,12 @@ import { FileViewedDecorationProvider } from '../fileViewedDecorationProvider';
 import { DecorationProvider } from '../treeDecorationProvider';
 import { TreeNode, TreeNodeParent } from './treeNode';
 
-export function openFileCommand(uri: vscode.Uri): vscode.Command {
+export function openFileCommand(uri: vscode.Uri, options: Pick<vscode.TextDocumentShowOptions, 'preview'>): vscode.Command {
```
alexr00, line 20: Why not just use `vscode.TextDocumentShowOptions` instead of `Pick<vscode.TextDocumentShowOptions, 'preview'>`?

src/commands.ts
```diff
@@ -138,6 +138,38 @@ export function registerCommands(
...
+						)
+					)
+					: activePullRequestsWithRootUri[0];
+
+				if (!activePullRequestWithRootUri) {
+					return;
+				}
+
+				const { activePr, rootUri } = activePullRequestWithRootUri;
+				(await activePr.getFileChangesInfo())
```
alexr00, line 165: `getFileChangesInfo` can be expensive. Instead of of calling `getFileChangesInfo` again (it's called as part of setting up the compare changes tree), you should be able to use the folder manager to get the review manager using  
https://github.com/joshuaobrien/vscode-pull-request-github/blob/a721a36d310522bcda872915886d0f48e6a9657f/src/view/reviewManager.ts#L976-L977
then, there's a change on `main` that adds a `ReviewModel` to the `ReviewManager`:
https://github.com/microsoft/vscode-pull-request-github/blob/6711e556f502bbd13cc9b9e9de8475eea746b488/src/view/reviewManager.ts#L40-L41 
if you add a getter for `_reviewModel` in `ReviewManager`, then you can the `localFileChanges` from the `ReviewModel` and call the `openFileCommand` method on each of the `localFileChanges`.

package.json
```diff
@@ -456,6 +456,11 @@
 				"category": "GitHub Pull Requests",
 				"icon": "$(globe)"
 			},
+			{
+				"command": "pr.openAllChanges",
+				"title": "Open All Changes",
```
alexr00, line 461: What do you think about making this command "Open All Diffs"? I usually prefer to view the diffs for each file, and I think the original issue was also interested in diffs.

alexr00 (changes requested):
Thanks for the PR, it's looking good! I pushed a change with a few nits changed and left some questions/suggestions.

package.json
joshuaobrien: Oh, yeah 🤦. I'll refactor 👍

alexr00 (approved):
Looks great, thank you!

PR #3244, opened 2022-01-11
fixed Visually out of place drop down in pull requests views with Safari

Yash621:
This PR is with respect to issue #2426

@alexr00 I have removed all the changes that are not related to fixing the bug,please review them.

let me know if any changes are required :)


alexr00:
Looks like we lost some of the original styling (the arrow on the right of the select is gone):

image

The select now also has rounded corners, which isn't consistent with other selects in VS Code:
image

The colors look good though. Do you know if there's a way to fix the lost styling?

Yash621:
@alexr00  sure , i would look into it and get back to you !

alexr00 (changes requested):
Thanks for the update! I'm seeing some style discrepancies, which I've left a comment on.

You can always push more commits to this PR and it will update, instead of creating a new PR each time :).

Yash621:
@alexr00  I have tried to fix the styling but i am not getting like why the arrows have disappeared ,can you guide me on this ?

alexr00:
@Yash621 I don't know why the arrows have disappeared off the top of my head so I don't have any guidance on this. I can do some digging into this later!

Yash621:
@alexr00 Please see once ,and let me know as I have tried some approches but none is working so please let me know :)

Yash621:
@alexr00  did you got any time to do some digging on why the arrows have disappeared ?

alexr00:
@Yash621 not yet!

saifahn:
I believe it is due to `-webkit-appearance: none;`, which removes the glossy default Safari styling, but it also removes the arrows. You can recreate the arrows with CSS, though.

Is this still a problem looking to be solved? I'd be interested in working on it if so :)

alexr00:
@saifahn, as far as I know https://github.com/microsoft/vscode-pull-request-github/issues/2426 is still a problem. If you want to take a crack at it please do!

PR #3247, opened 2022-01-13
Azure DevOps - update release/pre-release pipelines

lszomoru:


package.json
```diff
@@ -1544,7 +1543,7 @@
 		"browsertest": "yarn run browsertest:preprocess && node ./dist/browser/test/runTests.js",
 		"test:preprocess-gql": "node scripts/preprocess-gql --in src/github/queries.gql --out out/src/github/queries.gql",
 		"test:preprocess-svg": "node scripts/preprocess-svg --in ../resources/ --out out/resources",
-		"update-dts": "pushd \"src/@types\" && npx vscode-dts main && npx vscode-dts dev && popd",
+		"update-dts": "cd \"src/@types\" && npx vscode-dts main && npx vscode-dts dev",
```
alexr00, line 1546: Do we need to `cd ../../` at the end of `update-dts`?

azure-pipeline.release.yml
```diff
@@ -1,73 +1,53 @@
-trigger: none
+name: $(Date:yyyyMMdd)$(Rev:.r)
+
+trigger:
+  branches:
+    include:
+      - main
```
alexr00, line 6: When doing patch releases, we release from `release/major.minor` branches. Ex: `release/0.34`.

azure-pipeline.release.yml
```diff
@@ -1,73 +1,53 @@
-trigger: none
+name: $(Date:yyyyMMdd)$(Rev:.r)
+
+trigger:
+  branches:
```
alexr00, line 4: Does this mean that when I push a tag to one of the allowed branches the release pipeline will get triggered?

package.json
lszomoru: As far as I could tell, that was not necessary as the working directory is reset on each task

azure-pipeline.release.yml
lszomoru: Do we want CI to run on the release branches as well on each commit? Makes sense to me.

azure-pipeline.release.yml
lszomoru: I do not believe that tags are branch specific. When you push a tag, that will automatically run the "publish" phase.

azure-pipeline.release.yml
alexr00: I see now. This just means that it will only run automatically. I can still manually run on release branches. That's fine, no need to change.

alexr00 (approved):
Looks great, thank you! I might remove the trigger that causes it to run on every commit at some point. All the commits go through me now, and I make PRs for most things, so it seems a bit wasteful. I'll leave it for a month or so and see if it's useful.

PR #3307, opened 2022-02-09
Fix empty files tree

### LINKED ISSUE #3294: After making a change to a PR that was made on a fork the "Files" tree is empty
image

Refreshing the tree doesn't bring the files back.

alexr00:
Fixes #3294

### DIFF
--- a/src/view/prChangesTreeDataProvider.ts
+++ b/src/view/prChangesTreeDataProvider.ts
@@ -73,6 +73,12 @@ export class PullRequestChangesTreeDataProvider extends vscode.Disposable implem
 		reviewModel: ReviewModel,
 		shouldReveal: boolean,
 	) {
+		if (this._pullRequestManagerMap.has(pullRequestManager)) {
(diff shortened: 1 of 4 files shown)


src/view/prChangesTreeDataProvider.ts
```diff
@@ -73,6 +73,12 @@ export class PullRequestChangesTreeDataProvider extends vscode.Disposable implem
 		reviewModel: ReviewModel,
 		shouldReveal: boolean,
 	) {
+		if (this._pullRequestManagerMap.has(pullRequestManager)) {
+			const existingNode = this._pullRequestManagerMap.get(pullRequestManager);
+			if (existingNode && (existingNode.pullRequestModel === pullRequestModel)) {
+				return;
+			}
+		}
```
alexr00, line 81: Previously, the pull request node would get re-added to the tree view, but the old node wouldn't get disposed of. This led to tree nodes that didn't exist in the tree any more trying to refresh themselves, which in turn let to exceptions that prevented the tree from completing it's refresh.

rebornix (approved):

PR #3440, opened 2022-03-30
Support creating `insiders.vscode.dev` permalinks

joyceerhl:
Currently on insiders.vscode.dev, Copy GitHub Permalink generates a github.com URL. On vscode.dev it generates vscode.dev URLs, and on github.dev it generates github.dev URLs.


alexr00 (approved):
Thank you for the fix!

Issue #3324, opened 2022-02-16
How to have enterprise github URL in Copy as Permalink feature ?


utkarsh22garg:
When I try to do copy as permalink, it copies URL with github.com/.... but I would like to have github enterprise URL instead when copying. 

Is there a way to achieve so ?

- Extension version: 0.36.1
- VSCode Version: 1.64.1 (system setup)
- OS: Windows_NT x64 10.0.19042

alexr00:
Fixed by https://github.com/microsoft/vscode-pull-request-github/pull/3460

PR #3460, opened 2022-04-09
Support the copied permalink for Github Enterprise URL

hoontae24:
This PR is with respect to issue #3324

## About Copying a Permalink:
- The problem is `copyGithubPermalink` is always copying as a permalink based on `github.com`
- If the remote upstream's fetch url is not based on "github.com", the copied permalink should be started upstream's url origin.
- It helps using Github Enterprise Server.

## Applied Conditions
- You have to fill the Github Enterprise URI on VSCode config.(The attribute's name is `github-enterprise.uri`)
    - image

- The Host  of your git remote repository's upstream equals to the host of Github Enterprise URI
- If the remote url is based on SSH, it checks equality of `host`(`git@myenterprise.com` => `myenterprise.com`) to the host of Github Enterprise URI

**Let me know if there are any problems.**

### DIFF
--- a/src/issues/util.ts
+++ b/src/issues/util.ts
@@ -15,7 +15,7 @@ import { GithubItemStateEnum, User } from '../github/interface';
 import { IssueModel } from '../github/issueModel';
 import { PullRequestModel } from '../github/pullRequestModel';
 import { RepositoriesManager } from '../github/repositoriesManager';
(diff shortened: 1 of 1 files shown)


alexr00 (approved):
Looks great, thank you for the PR!

Issue #3432, opened 2022-03-24
Copying GitHub head fails with nothing being copies


digitarald:

alexr00:
> Right-click package.json and pick Copy GitHub Head Link

Is this in the explorer view or the tab heading?

digitarald:
@alexr00 in this case it was _Explorer_.

alexr00:
I see https://github.com/digitarald/theCatSaidNode/blob/main/package.json get copied. This isn't quite right, since the default branch in that repo is `master`, but it's not empty either.

joyceerhl:
I looked into this while investigating https://github.com/microsoft/vscode-pull-request-github/issues/3402 in case the fix for that was related.

This issue has a different cause: digitarald/theCatSaidNode is a fork, and GHPRI is using the parent repo's default_branch for the HEAD link generation: https://github.com/microsoft/vscode-pull-request-github/blob/1f2bb6ecf6eb92b1a6650162f2e954213afab387/src/github/folderRepositoryManager.ts#L1102-L1110

Opened a PR to fix this: https://github.com/microsoft/vscode-pull-request-github/pull/3477

PR #3477, opened 2022-04-19
Use HEAD branch name to generate HEAD links

joyceerhl:
Fix https://github.com/microsoft/vscode-pull-request-github/issues/3432

### DIFF
--- a/src/issues/util.ts
+++ b/src/issues/util.ts
@@ -575,9 +575,12 @@ export async function createGitHubLink(
 	if (!folderManager) {
 		return { permalink: undefined, error: 'Current file does not belong to an open repository.', originalFile: undefined };
 	}
-	const headBranchName = folderManager.repository.state.HEAD?.name;
-	if (!headBranchName) {
-		return { permalink: undefined, error: 'Not currently on a branch.', originalFile: undefined };
+	let branchName = folderManager.repository.state.HEAD?.name;
+	if (!branchName) {
+		// Fall back to default branch name if we are not currently on a branch
+		const origin = await folderManager.getOrigin();
+		const metadata = await origin.getMetadata();
+		branchName = metadata.default_branch;
 	}
 	const upstream = getSimpleUpstream(folderManager.repository);
 	if (!upstream?.fetchUrl) {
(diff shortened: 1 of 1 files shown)


src/issues/util.ts
```diff
@@ -575,14 +575,17 @@ export async function createGitHubLink(
...
+		return { permalink: undefined, error: 'Not currently on a branch.', originalFile: undefined };
+	}
 	const upstream = getSimpleUpstream(folderManager.repository);
 	if (!upstream?.fetchUrl) {
 		return { permalink: undefined, error: 'Repository does not have any remotes.', originalFile: undefined };
 	}
 	const pathSegment = uri.path.substring(folderManager.repository.rootUri.path.length);
 	return {
-		permalink: `https://github.com/${new Protocol(upstream.fetchUrl).nameWithOwner}/blob/${defaults.base
+		permalink: `https://github.com/${new Protocol(upstream.fetchUrl).nameWithOwner}/blob/${headBranchName
```
joyceerhl, line 588: Two callouts here:
1. This PR would lead to a behavior change for any existing users who expect this command to generate a link for the repo's default branch regardless of the current checkout, since it will now use the current branch name instead. I think this is what I would expect Copy GitHub HEAD link to do but just wanted to call that out here.
2. Additionally, when invoked on a PR, this would now generate a HEAD link for the fork repo rather than the upstream (which I also think is more reasonable than generating a link for the PR target repo, but I may be missing some context here for why we'd favor the current behavior).

alexr00: For (1), the goal was to have it use the default branch but I talked to Erich (who originally requested this command) and he likes that it would be the current branch. 

For (2), We do try to use the upstream repo for links, but for the HEAD link I think it makes sense to use the fork.

alexr00: If `headBranchName` is `undefined`, I think it makes sense to fall back to the default branch though.

alexr00 (changes requested):
I think we should fall back the default branch if there is no current branch. What do you think?

src/issues/util.ts
joyceerhl: Yeah, I was concerned that we generate the link for a file or folder that no longer exists on the default branch, and would have preferred to disable this command altogether when on a tag or commit. Though maybe that is OK since this is by definition not a permalink--we'd also have this problem when generating a link on a branch if that file or folder is subsequently deleted.

PR #3507, opened 2022-05-02
Show pull request's close button for author

blindpirate:
It's possible to create a pull request when the
author has no write permission in a repository.
In this case, the author should be ablt to see
the close button.


alexr00 (approved):
Thanks for the PR!

connor4312 (approved):

Issue #3371, opened 2022-03-03
Unable to Access Enterprise Git Pull Requests


harshanarayana:
```bash
❯ git remote --v
origin	git@github3.abcd-comp.com:myuser/some-repo.git (fetch)
origin	git@github3.abcd-comp.com:myuser/some-repo.git (push)
origin-https	https://github3.abcd-comp.com/my-org/some-repo.git (fetch)
origin-https	https://github3.abcd-comp.com/my-org/some-repo.git (push)
upstream	git@github3.abcd-comp.com:my-org/some-repo.git (fetch)
upstream	git@github3.abcd-comp.com:my-org/some-repo.git (push)
```

With the repo configuration as above, and following VSCode settings

```json
{
    "github-enterprise.uri": "https://github3.abcd-comp.com",
    "githubPullRequests.remotes": [
        "origin",
        "upstream",
    ]
}
```

I am unable to see any PRs being listed in the VSCode. 

![Screenshot 2022-03-03 at 7 00 28 PM](https://user-images.githubusercontent.com/1404867/156574477-a21deb84-c655-457c-9ce5-30d329a1b5b8.png)

```bash
[2022-03-03 18:56:35.177] [exthost] [info] Extension host with pid 2298 started
[2022-03-03 18:56:35.237] [exthost] [info] ExtensionService#_doActivateExtension vscode.configuration-editing, startup: false, activationEvent: 'onLanguage:jsonc'
[2022-03-03 18:56:35.241] [exthost] [info] ExtensionService#_doActivateExtension vscode.json-language-features, startup: false, activationEvent: 'onLanguage:jsonc'
[2022-03-03 18:56:35.252] [exthost] [info] ExtensionService#_doActivateExtension vscode.typescript-language-features, startup: false, activationEvent: 'onLanguage:jsonc'
[2022-03-03 18:56:35.260] [exthost] [info] ExtensionService#_doActivateExtension vscode.microsoft-authentication, startup: false, activationEvent: 'onAuthenticationRequest:microsoft'
[2022-03-03 18:56:35.272] [exthost] [info] ExtensionService#_doActivateExtension vscode.debug-auto-launch, startup: true, activationEvent: '*'
[2022-03-03 18:56:35.274] [exthost] [info] ExtensionService#_doActivateExtension vscode.git-base, startup: true, activationEvent: '*', root cause: vscode.github
[...]

PR #3565, opened 2022-05-21
Issue #3371 | Updated getAuthSessionOptions in case of GitHub Enterprise AuthProvider

jpspringall:
Resolves Unable to Access Enterprise Git Pull Requests #3371

### DIFF
--- a/src/github/credentials.ts
+++ b/src/github/credentials.ts
@@ -80,6 +80,11 @@ export class CredentialStore implements vscode.Disposable {
 			}
 		}
 		getAuthSessionOptions = { ...getAuthSessionOptions, ...{ createIfNone: false } };
+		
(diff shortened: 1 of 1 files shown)


alexr00 (approved):
Thank you for the fix!

hediet:
What is the difference between
```ts
getAuthSessionOptions = { ...getAuthSessionOptions, ...{ createIfNone: true, silent: false } };
```
and
```ts
getAuthSessionOptions = { ...getAuthSessionOptions, createIfNone: true, silent: false };
```
?

alexr00:
@hediet I would expect those to do the exact same thing.

jpspringall:
> @hediet I would expect those to do the exact same thing.

That is correct both do the same thing, I just copied the style that was there before

Issue #3382, opened 2022-03-08
PR view should emphasize assigning to yourself like on .com


Tyriar:
Assigning in extension requires filtering/scrolling:

![image](https://user-images.githubusercontent.com/2193314/157268585-b82af13f-4d8d-4d7a-b619-f9f7ce65001e.png)

Assigning via .com doesn't:

![image](https://user-images.githubusercontent.com/2193314/157268714-015417dc-1e87-4ec9-949c-e5d6f75f81d6.png)

alexr00:
Duplicate of https://github.com/microsoft/vscode-pull-request-github/issues/3627

PR #3601, opened 2022-06-09
Implement quick self assign link as available on .com (#3382)

leopoldsedev:
This does not exactly implement what has been suggested in #3382 , but instead implements the one-click button to self-assign that is also available on github.com. This streamlines the self-assign process even further than the suggested feature.
A further change might be to also implement what was originally suggested (i.e. putting yourself on top of the picker list).


alexr00 (approved):
I just added the `canEdit` check. Thank you for the PR!

connor4312 (approved):

PR #3712, opened 2022-07-11
Parallel API requests

benibenj:
All 5 requests are now performed in parallel. I have not seen any reason why they should have some sort of dependency between each other. By paralyzing the requests, a 3-4-fold speedup can be gained. It's now also caching the current user by accessing the `CredentialStore`

### DIFF
--- a/src/view/treeNodes/pullRequestNode.ts
+++ b/src/view/treeNodes/pullRequestNode.ts
@@ -6,7 +6,7 @@
 import * as vscode from 'vscode';
 import { Repository } from '../../api/api';
 import { getCommentingRanges } from '../../common/commentingRanges';
-import { SlimFileChange } from '../../common/file';
+import { InMemFileChange, SlimFileChange } from '../../common/file';
 import Logger from '../../common/logger';
 import { FILE_LIST_LAYOUT } from '../../common/settingKeys';
 import { fromPRUri, Schemes } from '../../common/uri';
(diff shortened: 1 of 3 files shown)


alexr00:
Suggested edit:
```diff
diff --git a/src/github/pullRequestModel.ts b/src/github/pullRequestModel.ts
index 498ee7f3..c920a1c6 100644
--- a/src/github/pullRequestModel.ts
+++ b/src/github/pullRequestModel.ts
@@ -783,7 +783,7 @@ export class PullRequestModel extends IssueModel<PullRequest> implements IPullRe
 
 			this.diffThreads(reviewThreads);
 			this._reviewThreadsCache = reviewThreads;
-
+			this._onDidChangeReviewThreads.fire({ added: reviewThreads, changed: [], removed: [] });
 			return reviewThreads;
 		} catch (e) {
 			Logger.appendLine(`Failed to get pull request review comments: ${e}`);

```

src/view/treeNodes/pullRequestNode.ts
```diff
@@ -74,9 +74,13 @@ export class PRNode extends TreeNode implements vscode.CommentingRangeProvider {
 				return [descriptionNode];
 			}
 
-			await this.pullRequestModel.initializeReviewThreadCache();
-			await this.pullRequestModel.initializePullRequestFileViewState();
-			this._fileChanges = await this.resolveFileChangeNodes();
+			[, , this._fileChanges, ,] = await Promise.all([
```
alexr00, line 77: I did a quick bit of testing of this, and I noticed that the little speech bubble character that we use as a file decoration for files with comments was no longer showing with this change. I suggested an edit that fixes the issue. This was a long standing bug that was only just revealed now that these methods are run in parallel. In short, this method was getting called when the comments threads changed, but not when they were originally set: https://github.com/microsoft/vscode-pull-request-github/blob/4c077d1ad6c5f8ad54c6adb0253bce87a6d9655e/src/view/treeDecorationProvider.ts#L12-L19

alexr00 (changes requested):
There's one change I suggested to fix a bug that this reveal, otherwise looks good.

PR #3725, opened 2022-07-14
Fix comment event firing twice

alexr00:


benibenj (approved):
Pipeline test passes now

PR #3729, opened 2022-07-15
Assignees get QuickPickItems async

benibenj:
Getting the QuickPickItems for assignees is currently not done asynchronously which leads to a delay before the dropdown opens.

Changed to: getting the QuickPickItems is done asynchronously which leads to the dropdown opening directly and the assignees appearing as soon as they have been loaded.


jrieken (approved):

joaomoreno (commented):
👏

PR #3737, opened 2022-07-19
Add progress notification to create

### LINKED ISSUE #3531: Investigate performance of "Create" view
The "Create PR" view is slow to load and slow to create a PR. We should see if there are opportunities to improve the load performance and the create performance.

alexr00:
Fixes #3531

### DIFF
--- a/src/github/createPRViewProvider.ts
+++ b/src/github/createPRViewProvider.ts
@@ -384,68 +384,79 @@ export class CreatePullRequestViewProvider extends WebviewViewBase implements vs
 	}
 
 	private async create(message: IRequestMessage<CreatePullRequest>): Promise<void> {
-		try {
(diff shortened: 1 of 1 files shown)


joaomoreno:
@alexr00 Is a notification the best we can do here? Can't we somehow show progress in the view itself?

joaomoreno:
@alexr00 A good example of what I mean is @benibenj's fix in https://github.com/microsoft/vscode-pull-request-github/pull/3729: that operation was slow so we had to show progress somewhere. The fix was to show the UI prematurely and show progress there. An alternative fix would've been to use a notification for progress and open quick open once done; this would've been a bit jarring.

Maybe we need a restructuring in the view such that we can open it immediately and show progress within... but it would definitely improve the overall experience. What do you think?

PR #3738, opened 2022-07-19
Benibenj/changesSinceReview

benibenj:
- Changes since Review button on each PR Description
- When diffs are open then they will close and reopen in the same tab order with the new diffs (only open if there are diffs)

- Only thing missing is the Icon


src/github/pullRequestOverview.ts
```diff
@@ -173,6 +174,7 @@ export class PullRequestOverviewPanel extends IssueOverviewPanel<PullRequestMode
 					requestedReviewers,
 					repositoryAccess,
 					branchInfo,
+					LatestReviewCommitInfo
```
alexr00, line 177: nit: `L` to `l`.

src/view/treeNodes/descriptionNode.ts
```diff
@@ -32,12 +32,27 @@ export class DescriptionNode extends TreeNode implements vscode.TreeItem {
...
 	}
+
+	updateContextValue(): void {
+		this.contextValue = 'description' +
+			(this.pullRequestModel.hasChangesSinceLastReview ? ':changesSinceReview' : '') +
+			(this.pullRequestModel.showChangesSinceReview ? ':active' : ':inactive');
+	}
+
+	protected registerSinceReviewChange() {
+		this.pullRequestModel.onDidChangeChangesSinceReview(_ => {
```
alexr00, line 53: This event listener should be disposed so that we don't leak listeners. The `TreeNode` has a way to do this:

https://github.com/microsoft/vscode-pull-request-github/blob/f2ea88b0d3fab2e6c2c8a6952b628abca5eeaa83/src/view/treeNodes/fileChangeNode.ts#L120-L126

src/view/treeNodes/pullRequestNode.ts
```diff
@@ -89,6 +89,8 @@ export class PRNode extends TreeNode implements vscode.CommentingRangeProvider {
 				);
 			}
 
+			descriptionNode.updateContextValue();
```
alexr00, line 92: I see this is also called in the constructor of the `DescriptionNode`. Is the call here also needed? Is it enough that you also listen for `pullRequestModel.onDidChangeChangesSinceReview` in the `DescriptionNode`?

src/view/treeNodes/repositoryChangesNode.ts
```diff
@@ -78,6 +79,41 @@ export class RepositoryChangesNode extends DescriptionNode implements vscode.Tre
 		return this;
 	}
 
+	protected registerSinceReviewChange() {
+		this.pullRequestModel.onDidChangeChangesSinceReview(directlyAfterActivation => {
```
alexr00, line 83: Same note as before about disposing this event listener.

src/view/treeNodes/repositoryChangesNode.ts
```diff
@@ -78,6 +79,41 @@ export class RepositoryChangesNode extends DescriptionNode implements vscode.Tre
...
+			this.updateContextValue();
+			vscode.commands.executeCommand('pr.refreshChanges').then(() => {
+				this.reopenNewDiffs(directlyAfterActivation);
+			});
+		});
+	}
+
+	public async reopenNewDiffs(directlyAfterActivation: boolean | void) {
```
alexr00, line 91: nit: I don't see `repenNewDiffs` used outside this class so it can be private.

src/view/reviewManager.ts
```diff
@@ -661,6 +697,16 @@ export class ReviewManager {
...
 	}
 
+	private async updateContentChanges(pr: PullRequestModel & IResolvedPullRequestModel): Promise<void> {
+		try {
+			const contentChanges = await pr.getFileChangesInfo();
+			this._reviewModel.localFileChanges = await this.getLocalChangeNodes(pr, contentChanges);
+		} catch (e) {
+			Logger.appendLine(`Review> ${e}`);
+		}
+		return Promise.resolve(void 0);
```
alexr00, line 707: I think you don't need this since this is an `async` method.

alexr00 (changes requested):
Change is looking good! I did a quick review so that you don't have to wait too long before getting feedback.

src/github/pullRequestModel.ts
```diff
@@ -417,6 +434,24 @@ export class PullRequestModel extends IssueModel<PullRequest> implements IPullRe
 		}
 	}
 
+	async getViewerLatestReviewCommit(): Promise<{ sha: string } | undefined> {
+		Logger.debug(`Fetch viewers latest review commit`, IssueModel.ID);
+		const { query, remote, schema } = await this.githubRepository.ensure();
+
+		const { data } = await query<LatestReviewCommitResponse>({
```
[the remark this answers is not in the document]
benibenj: We can integrate it into the PullRequest for GraphQL, however, in some places we are still using the REST API to get the information of a PR and the viewers latest review commit is not part of the response data of the REST PR get Request. (```GithubRepository.getPullRequestsForCategory```)

src/view/treeNodes/descriptionNode.ts
```diff
@@ -32,12 +38,63 @@ export class DescriptionNode extends TreeNode implements vscode.TreeItem {
...
+			});
+		}).flat();
+
+		await Promise.all(proms);
+	}
+
+	public async reopenTab(tab: vscode.Tab, pullRequest: PullRequestModel) {
+		if (tab.input instanceof vscode.TabInputTextDiff) {
+			if ((tab.input.original.scheme === Schemes.Pr) && (tab.input.modified.scheme === Schemes.Pr)) {
+				const changes = await (this.parent as PRNode).getFileChanges(true); // HERE THE 'UGLY' WORKAROUND
```
benibenj, line 84: Here is where we are looking for a solution as we've discussed. After refreshing we reopen the diffs but the parent node has not yet received the new changes.

alexr00: Can we move basically all of this, including the `onDidChangeChangesSinceReview` listener, into `PRNode`? That way, we can explicitly call and await `getFileChanges` before doing a `refresh`.

src/github/pullRequestModel.ts
```diff
@@ -1201,6 +1274,10 @@ export class PullRequestModel extends IssueModel<PullRequest> implements IPullRe
 			files = data.files as IRawFileChange[];
 		}
 
+		if (oldHasChangesSinceReview !== this.hasChangesSinceLastReview) {
+			this._onDidChangeChangesSinceReview.fire({ afterActivation: oldHasChangesSinceReview === undefined });
```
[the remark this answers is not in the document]
benibenj: This is done for the case when the user reopens VSCode and the last time he had the diffTab open in the changesSinceReview mode. In that case it will reopen the diff in the default state (show all Diffs)

src/view/treeNodes/descriptionNode.ts
benibenj: Yes, this works a lot nicer now in the PRNode

src/view/reviewManager.ts
```diff
@@ -556,6 +600,42 @@ export class ReviewManager {
 		return Promise.resolve(void 0);
 	}
 
+	public async updateChangedFiles(): Promise<void> {
```
alexr00, line 603: I don't think this is called anywhere now.

src/view/treeNodes/repositoryChangesNode.ts
```diff
@@ -61,6 +63,7 @@ export class RepositoryChangesNode extends DescriptionNode implements vscode.Tre
 	}
 
 	async getChildren(): Promise<TreeNode[]> {
+		await this._reviewManager.refreshSinceReview;
```
alexr00, line 66: Ideally, now that the other dependency on `ReviewManager` is gone, we'd be able to remove this one too. I do see that this makes showing progress work easily though. Let's save it for a separate PR during debt week and find a way to add something to `ReviewModel`.

src/view/treeNodes/descriptionNode.ts
```diff
@@ -32,12 +35,19 @@ export class DescriptionNode extends TreeNode implements vscode.TreeItem {
...
 		};
 
-		this.contextValue = 'description';
+
 		this.tooltip = `Description of pull request #${pullRequestModel.number}`;
 		this.accessibilityInformation = { label: `Pull request page of pull request number ${pullRequestModel.number}`, role: 'button' };
 	}
 
 	getTreeItem(): vscode.TreeItem {
+		this.updateContextValue();
```
alexr00, line 44: `updateContextValue` can be private now.

src/view/treeNodes/descriptionNode.ts
alexr00: Looks good!

alexr00 (changes requested):
Just a few more comments, all pretty small!

src/view/treeNodes/repositoryChangesNode.ts
benibenj: I'll remove it, but will have to figure out another way to show progress.

PR #3740, opened 2022-07-19
Update SCM path parsing

alexr00:
In response to https://github.dev/microsoft/vscode/pull/155624

### DIFF
--- a/src/issues/issueCompletionProvider.ts
+++ b/src/issues/issueCompletionProvider.ts
@@ -30,7 +30,7 @@ export class IssueCompletionProvider implements vscode.CompletionItemProvider {
 		private stateManager: StateManager,
 		private repositoriesManager: RepositoriesManager,
 		private context: vscode.ExtensionContext,
-	) {}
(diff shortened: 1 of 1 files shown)


src/issues/issueCompletionProvider.ts
```diff
@@ -182,7 +182,7 @@ export class IssueCompletionProvider implements vscode.CompletionItemProvider {
 			const configuration = vscode.workspace
 				.getConfiguration(ISSUES_CONFIGURATION)
 				.get('issueCompletionFormatScm');
-			if (document.uri.path.match(/scm\/git\/scm\d\/input/) && typeof configuration === 'string') {
+			if (document.uri.path.match(/git\/scm\d\/input/) && typeof configuration === 'string') {
```
lszomoru, line 185: Would it make sense to instead check that:
1. document URI schema is `vscode`
2. document URI authority is `scm`

lszomoru (approved):

PR #3781, opened 2022-07-27
show correct CSLR files when checkedout

### LINKED ISSUE #3771: Files still shown even after clicking show files changed since last review
Testing #3763

Was looking at a PR that looks like so:

image

That should have nothing in the list because I believe I'm in the show since last change mode

Click on the 'Show All Changes'

Now it shows nothing. 

image

Looks like the state is backwards.

benibenj:
Fixes #3771

We have the problem that when we change the CSLR state both the DescriptionNode in the All Prs view and the Current PR view refresh which leads to the DN of current PR view to fetch the correct diffs. However the Description Node of the All Pr view does not fetch the changes as it is the active PR and goes and directly gets the changes form the PRModel. However, this DN node might access the changes before the oder node could even start fetching which will lead to showing the old/same diffs or access the changes while the other node is fetching which leads to showing no diff at all as they have been cleared before being fetched.

The change solves this problem but refreshes slow (Only for active PR in ALL PR view). There currently also isn't a progress indication. 

The only other option I can think of is to just having both nodes fetch the file change info. This would lead to a faster refresh for the All PR view DN, however, this is not an elegant solution.

### DIFF
--- a/src/github/pullRequestModel.ts
+++ b/src/github/pullRequestModel.ts
@@ -120,9 +120,6 @@ export class PullRequestModel extends IssueModel<PullRequest> implements IPullRe
 	private _onDidChangeChangesSinceReview = new vscode.EventEmitter<void>();
 	public onDidChangeChangesSinceReview = this._onDidChangeChangesSinceReview.event;
 
-	public onDidChangeChangesSinceReviewActiveEmitter = new vscode.EventEmitter<void>();
-	public onDidChangeChangesSinceReviewActive = this.onDidChangeChangesSinceReviewActiveEmitter.event;
-
 	private _comments: IComment[] | undefined;
 	private _onDidChangeComments: vscode.EventEmitter<void> = new vscode.EventEmitter();
 	public readonly onDidChangeComments: vscode.Event<void> = this._onDidChangeComments.event;
--- a/src/github/pullRequestModel.ts
+++ b/src/github/pullRequestModel.ts
@@ -120,6 +120,9 @@ export class PullRequestModel extends IssueModel<PullRequest> implements IPullRe
 	private _onDidChangeChangesSinceReview = new vscode.EventEmitter<void>();
 	public onDidChangeChangesSinceReview = this._onDidChangeChangesSinceReview.event;
 
+	public onDidChangeChangesSinceReviewActiveEmitter = new vscode.EventEmitter<void>();
+	public onDidChangeChangesSinceReviewActive = this.onDidChangeChangesSinceReviewActiveEmitter.event;
+
 	private _comments: IComment[] | undefined;
 	private _onDidChangeComments: vscode.EventEmitter<void> = new vscode.EventEmitter();
 	public readonly onDidChangeComments: vscode.Event<void> = this._onDidChangeComments.event;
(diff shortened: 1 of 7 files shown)


src/view/treeNodes/pullRequestNode.ts
```diff
@@ -126,6 +126,13 @@ export class PRNode extends TreeNode implements vscode.CommentingRangeProvider {
 	protected registerSinceReviewChange() {
 		this._disposables.push(
 			this.pullRequestModel.onDidChangeChangesSinceReview(_ => {
+				if (!this.pullRequestModel.isActive) {
+					this.refresh();
+				}
+			})
+		);
+		this._disposables.push(
+			this.pullRequestModel.onDidChangeChangesSinceReviewActive(async _ => {
```
alexr00, line 135: Will refershing on this event work? 

https://github.com/microsoft/vscode-pull-request-github/blob/58e7c742befe7519abc21c082f5475a985bdb66f/src/view/reviewModel.ts#L12-L13

src/github/pullRequestModel.ts
```diff
@@ -120,6 +120,9 @@ export class PullRequestModel extends IssueModel<PullRequest> implements IPullRe
 	private _onDidChangeChangesSinceReview = new vscode.EventEmitter<void>();
 	public onDidChangeChangesSinceReview = this._onDidChangeChangesSinceReview.event;
 
+	public onDidChangeChangesSinceReviewActiveEmitter = new vscode.EventEmitter<void>();
```
alexr00, line 123: I'd like to avoid this pattern of having a public emitter. It leaks class internal state (when events should be fired) externally. I've suggested another option in another comment.

src/view/treeNodes/pullRequestNode.ts
alexr00: To get the review model into this class, you could pass it in here

https://github.com/microsoft/vscode-pull-request-github/blob/58e7c742befe7519abc21c082f5475a985bdb66f/src/view/reviewsManager.ts#L55-L56

alexr00 (changes requested):

src/view/treeNodes/pullRequestNode.ts
alexr00: If this doesn't end up working, we can add another event to the review model: `onDidChangeObsoleteFileChanges`, which should do the trick.

src/view/prsTreeDataProvider.ts
```diff
@@ -112,6 +113,11 @@ export class PullRequestsTreeDataProvider implements vscode.TreeDataProvider<Tre
 				});
 			}),
 		);
+		this._disposables.push(
+			...reviewModels.map(model => {
+				return model.onDidChangeLocalFileChanges(_ => { this.refresh(); });
```
alexr00, line 118: This refreshes the whole tree, instead of just the one PR node. Is this intentional?

alexr00 (approved):

src/view/prsTreeDataProvider.ts
benibenj: Calling refresh on the PR node calls refresh on the parent. This propagates all the way up to the tree instance. We discussed this once and decided it would be fine as refreshing only the node created an infinite refreshing state.

PR #3787, opened 2022-07-28
Error message for milestones

benibenj:
The `formatError` function does not consider some of the Validation Errors and returns an empty string error.

When creating a milestone with a name that is already taken but closed it can be very confusing. I added an Error message for that case. It could be done a bit more elegantly by also requesting closed milestones, however, this would require some more changes that I don't want to force into this release.

We could also consider lateron to give the option to reopen a milestone if the user enters a name for an already existing but closed milestone.


alexr00 (dismissed):

alexr00 (approved):
Thanks for adding a test!

PR #3796, opened 2022-07-29
Changelog

benibenj:
#3794 was causing all sorts of git problems when reverting the one commit or rebasing, so to not waste to much time I just created a new branch and PR.


connor4312:
Assigning to August assuming it's not a candidate

alexr00 (approved):

PR #3797, opened 2022-07-29
Don't open first diff when collapsing PR

benibenj:
Fixes the problem of opening diffs when collapsing PR. Might even improve the experience with the feature.

- Only open First diff when no diff is open and switching into CSLR mode
- Still reopens open diffs when toggeling CSLR


connor4312:
Assigning to August assuming it's not a candidate

alexr00 (approved):
👍

PR #3806, opened 2022-08-02
Don't get issues when only using milestones

### LINKED ISSUE #3791: Don't get issues when only using milestones
When we get milestones to show in the PR overview, we are actually getting a lot of the issues that are assigned to those milestones too.

This is an oversight. Instead, we should make a new GraphQL query that just get's the milestone names without also getting the issues associated with them.

alexr00:
Fixes #3791

### DIFF
--- a/src/github/queries.gql
+++ b/src/github/queries.gql
@@ -836,7 +836,20 @@ query MaxIssue($owner: String!, $name: String!) {
 	}
 }
 
-query GetMilestones($owner: String!, $name: String!, $assignee: String!) {
+query GetMilestones($owner: String!, $name: String!, $states: [MilestoneState!]!) {
+	repository(owner: $owner, name: $name) {
+			milestones(first: 12, orderBy: { direction: DESC, field: DUE_DATE }, states: $states) {
+				nodes {
+					dueOn
+					title
+					createdAt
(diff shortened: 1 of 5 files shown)


src/github/queries.gql
```diff
@@ -836,7 +836,20 @@ query MaxIssue($owner: String!, $name: String!) {
 	}
 }
 
-query GetMilestones($owner: String!, $name: String!, $assignee: String!) {
+query GetMilestones($owner: String!, $name: String!, $states: [MilestoneState!]!) {
+	repository(owner: $owner, name: $name) {
+			milestones(first: 12, orderBy: { direction: DESC, field: DUE_DATE }, states: $states) {
```
benibenj, line 841: What if a repo has more than 12 open milestones? This might not be a good habit but I'm sure there will be users who do have more. Also, in the case that both open and closed Milestones want to be fetched, it will be common that there are more than 12 milestones (vscode has almost 200).

benibenj (approved):

src/github/queries.gql
alexr00: Fair point. I'll up to 100 in another PR. If we need to add paging we can always add it later.

PR #3832, opened 2022-08-09
Handle depricated GHE versions LatestReviewCommit

benibenj:
Fixes candidate #3831


alexr00 (approved):
Verified that the changes since last review feature still works with this change.

PR #3853, opened 2022-08-15
fix: use ssh url for ssh protocol upstream

yin1999:
The git protocol is not supported by GitHub, see: https://github.blog/2021-09-01-improving-git-protocol-security-github/

When we use this extension in a repo which is using ssh protocol (such as: `git@github.com:microsoft/vscode-pull-request-github`), the upstream will be set as `git://xxxxxx`. But it's not supported on GitHub.

We should set ssh protocol instead.

## Steps to reproduce

1. Clone a fork to locale with ssh protocol `git clone git@github.com:<user>/vscode-pull-request-github.git`
2. Open this fork folder in vs code with vscode-pull-request-github extension
3. open file `.git/config`, we can see the upstream url is set to `git://github.com/microsoft/vscode-pull-request-github.git` (with git protocol which is not supported by GitHub now. And I think it should be ssh protocol which is the same as the fork's protocol)

### DIFF
--- a/src/github/folderRepositoryManager.ts
+++ b/src/github/folderRepositoryManager.ts
@@ -601,7 +601,7 @@ export class FolderRepositoryManager implements vscode.Disposable {
 						// check the remotes to see what protocol is being used
 						const isSSH = this.gitHubRepositories[0].remote.gitProtocol.type === ProtocolType.SSH;
 						if (isSSH) {
-							await this.repository.addRemote(remoteName, metadata.parent.git_url);
+							await this.repository.addRemote(remoteName, metadata.parent.ssh_url);
 						} else {
 							await this.repository.addRemote(remoteName, metadata.parent.clone_url);
 						}


apulll:
ytgrfed

apulll:
ytgrfedw

apulll (approved):
hygfds

yin1999:
Is there any problem on this?

alexr00 (approved):
Thanks for the PR and for the reminder! the `ssh_url` still used the `git` protocol in my testing, but it still makes sense to switch over to it in case that changes.

I will merge this PR later this week after we have released the new version of the extension.

yin1999:
> Thanks for the PR and for the reminder! the `ssh_url` still used the `git` protocol in my testing, but it still makes sense to switch over to it in case that changes.

Hi @alexr00, I'm not really sure about this, I've tested on my windows, and this works for me (if I just use my `fix-ssh-url` branch).

https://user-images.githubusercontent.com/15844309/193789604-bd496f1b-42be-42fb-9317-c00bd2d594a5.mp4

But, when I merge the newest main branch into this branch (`git rebase main`), the upstream will be set as `git` protocol:

https://user-images.githubusercontent.com/15844309/193789845-62163097-516d-4958-84e7-a53542d13b72.mp4

And I can't debug with the rebased branch, so I can't figure out what's wrong with this.

alexr00:
@yin1999 I'm still seeing the `git` protocol in your first video:

image

yin1999:
> This is still `ssh` protocol, because we can omit `ssh://` prefix:

See also: <https://git-scm.com/book/en/v2/Git-Internals-Transfer-Protocols#_ssh_2>

alexr00:
Thanks for the explanation! I tested again with a rebase to bring in the latest changes, and I get the correct format:
```
[remote "upstream"]
	url = git@github.com:microsoft/vscode-pull-request-github.git
	fetch = +refs/heads/*:refs/remotes/upstream/*
```

I'm still planning to merge this right after the next release.

PR #3854, opened 2022-08-15
Resolve and Unresolve

benibenj:
- Review Threads can be resolved/unresolved from the PullRequest Overview.
- ReviewThreads can be hidden.

### DIFF
--- a/src/common/timelineEvent.ts
+++ b/src/common/timelineEvent.ts
@@ -39,14 +39,16 @@ export interface CommentEvent {
 	createdAt: string;
 }
 
+export interface reviewResolveInfo {
+	threadId: string;
+	canResolve: boolean;
+	canUnresolve: boolean;
+	isResolved: boolean;
+}
+
 export interface ReviewEvent {
 	id: number;
-	reviewThread?: {
-		threadId: string;
-		canResolve: boolean;
-		canUnresolve: boolean;
-		isResolved: boolean;
-	}
+	reviewThread?: reviewResolveInfo
 	event: EventType;
 	comments: IComment[];
 	submittedAt: string;
--- a/src/common/timelineEvent.ts
+++ b/src/common/timelineEvent.ts
@@ -41,6 +41,12 @@ export interface CommentEvent {
 
 export interface ReviewEvent {
 	id: number;
+	reviewThread?: {
+		threadId: string;
+		canResolve: boolean;
+		canUnresolve: boolean;
+		isResolved: boolean;
+	}
 	event: EventType;
 	comments: IComment[];
 	submittedAt: string;
--- a/src/common/timelineEvent.ts
+++ b/src/common/timelineEvent.ts
@@ -39,7 +39,7 @@ export interface CommentEvent {
 	createdAt: string;
 }
 
-export interface reviewResolveInfo {
+export interface ReviewResolveInfo {
 	threadId: string;
 	canResolve: boolean;
 	canUnresolve: boolean;
@@ -48,7 +48,7 @@ export interface reviewResolveInfo {
 
 export interface ReviewEvent {
 	id: number;
-	reviewThread?: reviewResolveInfo
+	reviewThread?: ReviewResolveInfo
 	event: EventType;
 	comments: IComment[];
 	submittedAt: string;
--- a/src/github/pullRequestModel.ts
+++ b/src/github/pullRequestModel.ts
@@ -997,6 +997,20 @@ export class PullRequestModel extends IssueModel<PullRequest> implements IPullRe
 			}
 		});
 
+		reviewThreads.forEach(thread => {
+			if (!thread.prReviewDatabaseId || !reviewEventsById[thread.prReviewDatabaseId]) {
+				return;
+			}
+			const prReviewThreadEvent = reviewEventsById[thread.prReviewDatabaseId];
+			prReviewThreadEvent.reviewThread = {
+				threadId: thread.id,
+				canResolve: thread.viewerCanResolve,
+				canUnresolve: thread.viewerCanUnresolve,
+				isResolved: thread.isResolved
+			};
+
+		});
+
 		const pendingReview = reviewEvents.filter(r => r.state.toLowerCase() === 'pending')[0];
 		if (pendingReview) {
 			// Ensures that pending comments made in reply to other reviews are included for the pending review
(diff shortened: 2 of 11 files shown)


src/github/pullRequestModel.ts
```diff
@@ -997,6 +997,20 @@ export class PullRequestModel extends IssueModel<PullRequest> implements IPullRe
...
+			const prReviewThreadEvent = reviewEventsById[thread.prReviewDatabaseId];
+			prReviewThreadEvent.reviewThread = {
+				threadId: thread.id,
+				canResolve: thread.viewerCanResolve,
+				canUnresolve: thread.viewerCanUnresolve,
+				isResolved: thread.isResolved
+			};
+
+		});
+
```
alexr00, line 1013: Why is this needed? The `PullRequestComments` query, which is what `reviewThreads` seems to come from, looks like it should have these properties already:

https://github.com/microsoft/vscode-pull-request-github/blob/da74def171b2bd0ff0123e015c3a67fb7fbe4ba8/src/github/queries.gql#L378-L386

webviews/common/context.tsx
```diff
@@ -176,6 +176,13 @@ export class PRContext {
 
 	public openDiff = (comment: IComment) => this.postMessage({ command: 'pr.open-diff', args: { comment } });
 
+	public toggleResolveComment = (reviewThread: ReviewEvent['reviewThread'], thread: IComment[]) => {
+		if (!reviewThread) {
+			return;
+		}
+		this.postMessage({ command: 'pr.resolve-comment-thread', args: { threadId: reviewThread.threadId, toResolve: reviewThread.isResolved, thread } });
```
alexr00, line 183: Might need to await this `postMessage` and then call `updatePR`. That could be where the unexpected UI behavior we were seeing is coming from.

src/github/pullRequestModel.ts
benibenj: We are getting the information from `PullRequestComments`, but need the TimelineEvent to know about this information. And as the TimelineEvents API does not know about the threadId and resolve permissions, we need to add this information to the event.

webviews/common/context.tsx
benibenj: Actually I was only missing the `updatePR` call. I was changing the values in the timeline but I did not update the state which caused the inconsistent state.

webviews/common/context.tsx
```diff
@@ -176,6 +176,17 @@ export class PRContext {
...
 	public openDiff = (comment: IComment) => this.postMessage({ command: 'pr.open-diff', args: { comment } });
 
+	public toggleResolveComment = (reviewThread: ReviewEvent['reviewThread'], thread: IComment[], newResolved: boolean) => {
+		if (!reviewThread) {
+			return;
+		}
+		reviewThread.isResolved = newResolved;
+		reviewThread.canUnresolve = newResolved;
+		reviewThread.canResolve = !newResolved;
+		this.updatePR({});
```
alexr00, line 186: The usual patter for `updatePR` is to pass the properteis of the PR that changed into `updatePR`. Does just calling it without passing in the updated `{ events }` work?

alexr00: I guess you're modifying the original object, so that's why it would work. I find this to be a little tricky to understand and not self-documenting (I have to make sure I understand that `reviewThread` is the same object that is already in `this.pr`, there are no clues about what is happening). Instead, is it possible to use the usual `const response = await this.postMessage` then pass some property of the `response` into `updatePR`?

benibenj: It seemed to work but I just noticed it actually doe not. I'm trying to figure out how to only update the state of one array item

webviews/common/context.tsx
```diff
@@ -176,6 +176,17 @@ export class PRContext {
 
 	public openDiff = (comment: IComment) => this.postMessage({ command: 'pr.open-diff', args: { comment } });
 
+	public toggleResolveComment = (reviewThread: ReviewEvent['reviewThread'], thread: IComment[], newResolved: boolean) => {
```
alexr00, line 179: nit: I would prefer we break `ReviewThread` into a separate type.

src/common/timelineEvent.ts
```diff
@@ -39,8 +39,16 @@ export interface CommentEvent {
 	createdAt: string;
 }
 
+export interface reviewResolveInfo {
```
alexr00, line 42: The style in this codebase is that interface names are Pascal case.

alexr00: And the signature [here](https://github.com/microsoft/vscode-pull-request-github/blob/9859e75b24f99d8402fc443a7bd286e6ed0bc1c5/webviews/common/context.tsx#L179) should use the new type.

webviews/common/context.tsx
alexr00: Would it make sense to call `PullRequestModel`'s `getTimelineEvents` method after doing the resolve in `PullRequestOverViewPanel`? That would also have the benefit of updating state that the overview uses so that we're up-to-date in both the extension and the webview as well as giving you the correct item to include in the message response.

alexr00 (changes requested):

src/github/pullRequestModel.ts
```diff
@@ -938,17 +939,25 @@ export class PullRequestModel extends IssueModel<PullRequest> implements IPullRe
...
-			const { data } = await query<TimelineEventsResponse>({
-				query: schema.TimelineEvents,
-				variables: {
-					owner: remote.owner,
-					name: remote.repositoryName,
-					number: this.number,
-				},
-			});
+			const [{ data }, latestReviewCommitInfo, currentUser, reviewThreads] = await Promise.all([
+				await query<TimelineEventsResponse>({
```
alexr00, line 943: unneeded await?

webviews/common/context.tsx
```diff
@@ -5,7 +5,7 @@
 
 import { createContext } from 'react';
 import { IComment } from '../../src/common/comment';
-import { EventType, isReviewEvent, ReviewEvent } from '../../src/common/timelineEvent';
+import { EventType, isReviewEvent, ReviewEvent, ReviewResolveInfo, TimelineEvent } from '../../src/common/timelineEvent';
```
alexr00, line 8: lint error with unused `ReviewResolveInfo`. I'll try to see why the build isn't catching this.

alexr00 (dismissed):

src/github/pullRequestModel.ts
benibenj: yes, thanks for noticing

alexr00 (approved):
👍

PR #3873, opened 2022-08-22
Update post-commit command

lszomoru:


alexr00 (approved):
Discussed with Lad:
- Description is shown when there's enough space
- Title is shown when space is limited.

Issue #3900, opened 2022-08-29
Buttons don't match VS Code's new style


Tyriar:
Current PR view:

![image](https://user-images.githubusercontent.com/2193314/187226356-3f9ec9b4-70c1-4ac9-994e-48962414de34.png)

Expected round corners:

![image](https://user-images.githubusercontent.com/2193314/187226427-a748cb2b-92aa-45bd-a984-0635a3be38e4.png)

alexr00:
Actually, let's take this as a candidate. Should be very low risk.

Issue #3785, opened 2022-07-27
Unify drop-down buttons


hediet:
The pull request extension also has a drop down button, but it works a bit different than the commit drop down button. Maybe it makes sense to unify these two ideas?

![image](https://user-images.githubusercontent.com/2931520/181295639-7acdc072-00a0-419f-b930-26e9407b9393.png)

alexr00:
I'm fine with bringing the GHPRI button into consistency with the way this button works in VS Code core/the commit button.

hediet:
I like about the GHPRI button that it remembers the last choice!

lszomoru:
@hediet, it's on the plan for next milestone. Stay tuned...

alexr00:
I will keep remembering the last choice in GHPRI, but I will also make it only 1 click to change choice as the git button is.

daviddossett:
I'll chat w/ @hawkticehurst about the split button today. The webview toolkit is built on top of [FAST](https://github.com/microsoft/fast/) which doesn't yet provide a split button base component out of the box. Nitty gritty details aside, we'll see if there's a workaround and will report back.

hawkticehurst:
Catching up on old issues and the quick update about split button is now that a context menu API is being finalized for use in webviews the possibility to create our own split button is really opening up.

Assuming that officially lands sooner than later I'd love to ship a split button before the end of this year.

alexr00:
This has been fixed!

Issue #3035, opened 2021-10-13
Use `Webview UI Toolkit for Visual Studio Code` for webview controls


daviddossett:
## Proposal

Hi all! 👋 

Now that we've released the [Webview UI Toolkit for Visual Studio Code](https://github.com/microsoft/vscode-webview-toolkit), I think it would great to swap the bespoke/styled UI components in the extension with their counterparts in the toolkit. This would reduce the load on maintainers to have to keep up with visual/theme/accessibility consistency with VS Code itself, among other advantages. 

You can view the components in a Storybook environment [here](https://microsoft.github.io/vscode-webview-ui-toolkit/?path=/story/docs-introduction--page).

From what I've seen, the primary webviews in the extension use inputs, buttons (icon, primary, secondary), tags, checkboxes, and links. I believe the toolkit covers all of those components spare the split button currently used for selecting a merge method. That component [is in our backlog](https://github.com/microsoft/vscode-webview-ui-toolkit/issues/39) to be added soon. 

## References

<img width="369" alt="CleanShot 2021-10-13 at 11 15 40@2x" src="https://user-images.githubusercontent.com/25163139/137192291-5c166450-dd46-4ba9-8714-93e398bfb5e0.png">

<img width="908" alt="CleanShot 2021-10-13 at 11 17 08@2x" src="https://user-images.githubusercontent.com/25163139/137192320-9dc1f3f3-c675-4f0b-92dc-da5e0229e75a.png">

cc @hawkticehurst @misolori

PR #3904, opened 2022-08-30
Explore use of vscode-webview-ui-toolkit

Thomas1664:
I saw a few issues related to [vscode-webview-ui-toolkit](https://github.com/microsoft/vscode-webview-ui-toolkit) and I think using their components can make life easier for this extension:
#3035, #3785, #3900

However, I ran into a few issues when integrating components from the toolkit:

- Icon buttons aren't properly formatted
  I think the best solution here would be to use codicons but the problem is that they don't provide a React component which makes them harder to use and to integrate.
- Button groups require changing the CSS
  I whish the UI toolkit provided a way to handle this as button groups are a common use case.


hawkticehurst:
Hey @Thomas1664, maintainer of the toolkit here! 👋

I was catching up on old issues and saw this (and some other GHPRI extension issues related to the toolkit) linked and wanted to check in.

It's been on @daviddossett and I's to-do list for a long time to try and get this extension onboarded to the toolkit. We're hoping to [ship a split button by the end of the year](https://github.com/microsoft/vscode-pull-request-github/issues/3785#issuecomment-1247513011) so as we're working towards that I'd love to hear more about any other blockers you have in adopting the toolkit and how we can help resolve those.

As a starter, it would be great to learn a bit more about the icon button and button group issues you were discussing above. Could you give some more context and details about this? (p.s. screenshots of examples are also always appreciated too)

Thanks in advance!

Thomas1664:
Hi @hawkticehurst,

The reason why I want to have a `ButtonGroup` component is because there is no space between buttons. Bootstrap for example offers a similar feature via `btn-group`.

image

The icon button thing can be dropped because it isn't the fault of the toolkit that the icon isn't displayed in the center of the button but I wished there was an `Icon` component that allows the use of vscodeicons via React.

daviddossett:
Thanks for looking into this @Thomas1664. Once I've closed the larger overhaul PRs tied to #4029 I'll come back to this to see if I can resolve the issue with the icon button. I'm less worried about the button group for now but agree that it could be useful to have that in the future. 

@hawkticehurst codicons should work just fine with the React versions of the toolkit, no?

src/github/issueOverview.ts
```diff
@@ -390,7 +390,8 @@ export class IssueOverviewPanel<TItem extends IssueModel = IssueModel> extends W
 		<title>Pull Request #${number}</title>
 	</head>
 	<body class="${process.platform}">
-		<div id=app></div>
+		<div id=app>
+		</div>
```
mmlinhang, line 394: 111

webviews/components/comment.tsx
```diff
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
```
rasul-1008, line 3: @ @

Mouhand999067 (approved):
Done

alexr00:
It would still be cool to adopt the toolkit; however, given the age of this PR I think it's dead. We can always re-open if there's further interest.

PR #3909, opened 2022-09-01
Don't error out when deleting branches

alexr00:
when the branch doesn't exist

### DIFF
--- a/src/github/folderRepositoryManager.ts
+++ b/src/github/folderRepositoryManager.ts
@@ -1672,9 +1672,17 @@ export class FolderRepositoryManager implements vscode.Disposable {
 						try {
 							await Promise.all(
 								picks.map(async pick => {
-									await this.repository.deleteBranch(pick.label, true);
-								}),
-							);
+									try {
+										await this.repository.deleteBranch(pick.label, true);
+									} catch (e) {
+										if ((typeof e.stderr === 'string') && (e.stderr as string).includes('not found')) {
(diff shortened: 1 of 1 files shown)


roblourens (approved):

src/github/folderRepositoryManager.ts
```diff
@@ -1672,9 +1672,17 @@ export class FolderRepositoryManager implements vscode.Disposable {
...
 							await Promise.all(
 								picks.map(async pick => {
-									await this.repository.deleteBranch(pick.label, true);
-								}),
-							);
+									try {
+										await this.repository.deleteBranch(pick.label, true);
+									} catch (e) {
+										if ((typeof e.stderr === 'string') && (e.stderr as string).includes('not found')) {
+											// TODO: The git extension API doesn't support removing configs
```
roblourens, line 1679: I'm curious what "removing configs" means here?

alexr00:
The git config file has entries like this in it:
```
[branch "alexr00/defaultMerge"]
	remote = origin
	merge = refs/heads/alexr00/defaultMerge
	github-pr-owner-number = "microsoft#vscode#149237"
```
That last property, `github-pr-owner-number` is added by GHPRI, and we use it list all the branches that we can offer to delete (among other things). In this catch block, we now know that the branch has been deleted, but for some reason we still have a `branch` entry in the config. I don't know how that can happen, but at the very least we should remove the `github-pr-owner-number` sub-entry at this point.

PR #3915, opened 2022-09-02
Update pr vmImage

alexr00:
The old one is being deprecated and we just got hit with the warning brown out.
https://github.com/actions/runner-images/issues/5583


alexr00:
@benibenj FYI in case you see PR builds failing.

bpasero (approved):

PR #3916, opened 2022-09-02
Mark file as viewed with checkmark

benibenj:


src/view/treeNodes/directoryTreeNode.ts
```diff
@@ -4,13 +4,15 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as vscode from 'vscode';
+import { ViewedState } from '../../common/comment';
```
alexr00, line 7: Unused import

src/view/treeNodes/fileChangeNode.ts
```diff
@@ -14,6 +14,7 @@ import { FolderRepositoryManager, SETTINGS_NAMESPACE } from '../../github/folder
 import { IResolvedPullRequestModel, PullRequestModel } from '../../github/pullRequestModel';
 import { FileChangeModel, GitFileChangeModel, InMemFileChangeModel, RemoteFileChangeModel } from '../fileChangeModel';
 import { DecorationProvider } from '../treeDecorationProvider';
+import { DirectoryTreeNode } from './directoryTreeNode';
```
alexr00, line 17: Now we have a dependency cycle where `FileChangeNode` depends on `DirectoryTreeNode` and vice versa. It looks like this dependency exists for `updateParentCheckbox`. I see that you've already added `updateCheckbox` to `TreeNode`.  Can `updateParentCheckbox` be added to `TreeNode` as well?

src/view/treeNodes/fileChangeNode.ts
```diff
@@ -98,7 +101,7 @@ export class FileChangeNode extends TreeNode implements vscode.TreeItem {
 	}
 
 	constructor(
-		public readonly parent: TreeNodeParent,
+		public parent: TreeNodeParent,
```
alexr00, line 104: I see why you needed to remove the `readonly` to change the `parent` later. To make the code more self-documenting, can we:
- Make `parent` `_parent` and make it private
- Have a `get parent`
- Have a public method for setting the parent. Something like `updateParentDirectory`.

src/view/treeNodes/fileChangeNode.ts
```diff
@@ -161,6 +163,19 @@ export class FileChangeNode extends TreeNode implements vscode.TreeItem {
...
 		this.contextValue = `${Schemes.FileChange}:${GitChangeType[this.changeModel.status]}:${viewed === ViewedState.VIEWED ? 'viewed' : 'unviewed'
 			}`;
+		this.checkboxState = viewed === ViewedState.VIEWED ?
+			vscode.TreeItemCheckboxState.Checked : vscode.TreeItemCheckboxState.Unchecked;
+	}
+
+	updateCheckbox(newState: vscode.TreeItemCheckboxState) {
+		const viewed = newState === vscode.TreeItemCheckboxState.Checked ? ViewedState.VIEWED : ViewedState.UNVIEWED;
+		this.updateViewed(viewed);
+		vscode.commands.executeCommand(
```
alexr00, line 173: Executing our own commands can indicate that we're circumventing some intentional layering and makes debugging and code exploration more difficult (no find references). In this case we can avoid the `executeCommand` by moving part of the commands into `FileChangeNode` as a public methods: 

https://github.com/microsoft/vscode-pull-request-github/blob/4204dd158ad2267548d7cb63db81ff9ddd4fc8fb/src/commands.ts#L908-L912

src/view/fileTypeDecorationProvider.ts
```diff
@@ -156,19 +143,12 @@ export class FileTypeDecorationProvider implements vscode.FileDecorationProvider
...
-		}
-
+	color(status: GitChangeType): vscode.ThemeColor | undefined {
 		let color: string | undefined = vscode.extensions.getExtension('vscode.git') ? this.gitColors(status) : this.remoteReposColors(status);
 		return color ? new vscode.ThemeColor(color) : undefined;
 	}
 
-	letter(status: GitChangeType, viewedState?: ViewedState): string {
-		if (viewedState === ViewedState.VIEWED) {
-			return '✓';
```
alexr00, line 170: Very happy to see this decoration go!

alexr00 (changes requested):

src/view/treeNodes/fileChangeNode.ts
benibenj: I believe this is not possible as we require the parent field due to extending TreeNode.
Because of this, it won't allow us to create accessors (get parent).
We also can't restrict the visibility of parent to private as the super class has it has public.

alexr00 (dismissed):
Approved so we can get this in for selfhosting, but we should still address the file dependency cycle so that we don't accrue further debt.

alexr00 (approved):

PR #3918, opened 2022-09-05
Rendering ignoring line breaks

### LINKED ISSUE #3881: Rendering ignoring line breaks
- Extension version: 0.49.2022082209 
- VSCode Version: 1.70.2
- OS: linux

Steps to Reproduce:
Writing a comment with regular linebreaks (no double space at the end) will render wrong in vscode.

image

in github - the linebreak is rendered properly.

image

The actual content looks like this:

image

While the content is not proper markdown (which would require either an empty line or a double-space to render a linebreak) - i think this extension should align it's rendering to how github is rendering content.

alexr00:
Fixes #3881

### DIFF
--- a/src/github/prComment.ts
+++ b/src/github/prComment.ts
@@ -68,9 +68,9 @@ export class TemporaryComment implements vscode.Comment {
 	public parent: GHPRCommentThread;
 
 	/**
-	 * The text of the comment
+	 * The text of the comment as from GitHub
 	 */
-	public body: string | vscode.MarkdownString;
+	public rawBody: string;
 
 	/**
 	 * If the temporary comment is in place for an edit, the original text value of the comment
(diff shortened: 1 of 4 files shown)


src/github/prComment.ts
```diff
@@ -225,6 +235,16 @@ export class GHPRComment implements vscode.Comment {
 		this.timestamp = new Date(comment.createdAt);
 	}
 
+	get body(): vscode.MarkdownString | string {
+		// VS Code's markdown rendering is more correct and will not render single line breaks as
+		// line breaks in markdown. To make the comment look more like github.com, we replace single line breaks with double.
+		return (this.mode === vscode.CommentMode.Editing) ? this.rawBody : new vscode.MarkdownString(this.rawBody.replace(/(?<!\s)(\r\n|\n)(?!\s)/g, '\n\n'));
```
hediet, line 241: Why is this a different regex than above?
Does it make sense to extract this into a new function?

hediet (approved):

src/github/prComment.ts
alexr00: Thanks for the reminder! I prefer to do refactors separate from other changes: https://github.com/microsoft/vscode-pull-request-github/pull/3921

Issue #3935, opened 2022-09-08
PR diff in web fails to open for added or modified files opened from the All Open node when the PR is the current PR


joyceerhl:
Originally reported by @tyriar 

1. Open https://insiders.vscode.dev/github/joyceerhl/demo-pr-diffs-bug/tree/test. This is a branch which has an associated active PR, and contains a file addition and file deletion
2. Open the All Open view and click on the file diffs for the current PR there
3. :bug: the diff for the deleted file can be opened, but the diff for the added file (and also modified files) cannot be opened.
<img width="1370" alt="image" src="https://user-images.githubusercontent.com/30305945/189188004-41524635-1938-4d88-821b-86c0e702158a.png">

I believe the problem is that we are using `file://` URIs whenever the change is in the current PR and the file change is not a deletion: 

https://github.com/microsoft/vscode-pull-request-github/blob/933b473f22797f7d43ecc4e46159f56fb2fee720/src/view/fileChangeModel.ts#L135-L139

PR #3936, opened 2022-09-08
Use repository root URI scheme for added and changed files

joyceerhl:
This PR fixes https://github.com/microsoft/vscode-pull-request-github/issues/3935

I tested that this fixes opening PR diffs for added and modified files from the All Open view tree node in both RemoteHub and desktop git repositories.

### DIFF
--- a/src/view/fileChangeModel.ts
+++ b/src/view/fileChangeModel.ts
@@ -132,26 +132,20 @@ export class InMemFileChangeModel extends FileChangeModel {
 		super(pullRequest, folderRepositoryManager, change);
 		const headCommit = pullRequest.head!.sha;
 		const parentFileName = change.status === GitChangeType.RENAME ? change.previousFileName! : change.fileName;
-		const filePath = vscode.Uri.file(resolvePath(folderRepositoryManager.repository.rootUri, change.fileName));
-		const parentPath = vscode.Uri.file(resolvePath(folderRepositoryManager.repository.rootUri, parentFileName));
-		this._filePath = isCurrentPR ? toReviewUri(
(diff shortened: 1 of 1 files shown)


src/view/fileChangeModel.ts
```diff
@@ -134,19 +134,24 @@ export class InMemFileChangeModel extends FileChangeModel {
 		const parentFileName = change.status === GitChangeType.RENAME ? change.previousFileName! : change.fileName;
 		const filePath = vscode.Uri.file(resolvePath(folderRepositoryManager.repository.rootUri, change.fileName));
 		const parentPath = vscode.Uri.file(resolvePath(folderRepositoryManager.repository.rootUri, parentFileName));
-		this._filePath = isCurrentPR ? ((change.status === GitChangeType.DELETE)
-			? toReviewUri(filePath, undefined, undefined, '', false, { base: false }, folderRepositoryManager.repository.rootUri)
-			: filePath)
```
alexr00, line 139: We need to use `file` for the active PR because we will end up with duplicate comments (one for the `file` scheme and one for the `review` scheme) if we use the `review` scheme. However, `file` is also clearly not right as the open folder doesn't always have the `file` scheme. Instead, we should use the same scheme as the currently open folder. The way I've done that in the past is to use the repository's root URI.

alexr00 (changes requested):
Thanks for the PR! We still can't use the review scheme here, but you are right that file is also incorrect. If you don't want to make the scheme change let me know and I'll make it.

alexr00 (approved):
Thank you!

PR #3946, opened 2022-09-15
Fix a dead end for GitHub Enterprise auth

alexr00:
When you're already signed in with github.com and you open an github enterprise workspace but don't have the enterprise uri setting set you end up at a dead end where nothing works.


TylerLeonhardt (dismissed):

src/@types/vscode.proposed.treeItemCheckbox.d.ts
```diff
@@ -0,0 +1,45 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+declare module 'vscode' {
+
+	export class TreeItem2 extends TreeItem {
+		/**
+		 * [TreeItemCheckboxState](#TreeItemCheckboxState) of the tree item.
+		 */
+		checkboxState?: TreeItemCheckboxState | { state: TreeItemCheckboxState; tooltip?: string };
+	}
```
benibenj, line 13: You probably don't want to push this vscode proposed API into the repo, right?

benibenj (approved):

src/@types/vscode.proposed.treeItemCheckbox.d.ts
alexr00: All the other API proposals are are in this repo (https://github.com/microsoft/vscode-pull-request-github/tree/main/src/%40types). I can them all to the .gitignore, but for now I'm just going to push it.

joaomoreno (approved):

PR #3979, opened 2022-09-23
Add link commands back into Copy As

alexr00:
Part of microsoft/vscode#157722

### DIFF
--- a/package.json
+++ b/package.json
@@ -1713,6 +1713,23 @@
 					"when": "commentController =~ /^github-(browse|review)/"
 				}
 			],
+			"editor/context/copy": [
+				{
+					"command": "issue.copyGithubPermalink",
+					"when": "gitHubOpenRepositoryCount != 0",
+					"group": "3_githubPullRequests@0"
+				},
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1713,6 +1713,23 @@
...
+					"when": "gitHubOpenRepositoryCount != 0",
+					"group": "3_githubPullRequests@1"
+				},
+				{
+					"command": "issue.copyGithubHeadLink",
+					"when": "gitHubOpenRepositoryCount != 0",
+					"group": "3_githubPullRequests@2"
+				}
+			],
 			"editor/context/share": [
```
benibenj, line 1733: Are you not removing the old ones?

benibenj (approved):

package.json
alexr00: After discussing with the team and with Kai, I decided to have the commands in both places. I don't want to break current user's model for where to find the commands, but we also want to make the more discoverable to users who think to look for "Share" first.

PR #4014, opened 2022-10-03
Enable localization

alexr00:


package.nls.json
```diff
@@ -0,0 +1,205 @@
...
+	"githubPullRequests.commentExpandState.description": "Controls whether comments are expanded when a document with comments is opened.",
+	"githubPullRequests.commentExpandState.expandUnresolved": "All unresolved comments will be expanded.",
+	"githubPullRequests.commentExpandState.collapseAll": "All comments will be collapsed",
+	"githubPullRequests.useReviewMode.description": "Choose which pull request states will use review mode. \"Open\" pull requests will always use review mode.",
+	"githubPullRequests.useReviewMode.merged": "Use review mode for merged pull requests.",
+	"githubPullRequests.useReviewMode.closed": "Use review mode for closed pull requests. Merged pull requests are not considered \"closed\".",
+	"githubPullRequests.assignCreated.description": {
+		"message": "All pull requests created with this extension will be assigned to this user. To assign to yourself, use the '${user}' variable.",
+		"comment": [
+			"{Locked=']${user}'}",
```
TylerLeonhardt, line 33:
```suggestion
			"{Locked='${user}'}",

```

package.nls.json
```diff
@@ -0,0 +1,205 @@
...
+	"githubIssues.userCompletions.enabled.description": "Controls whether completion suggestions are shown for users.",
+	"githubIssues.ignoreCompletionTrigger.description": "Languages that the '#' character should not be used to trigger issue completion suggestions.",
+	"githubIssues.ignoreCompletionTrigger.items": "Language that issue completions should not trigger on '#'.",
+	"githubIssues.ignoreUserCompletionTrigger.description": "Languages that the '@' character should not be used to trigger user completion suggestions.",
+	"githubIssues.ignoreUserCompletionTrigger.items": "Language that user completions should not trigger on '@'.",
+	"githubIssues.issueBranchTitle.markdownDescription": {
+		"message": "Advanced settings for the name of the branch that is created when you start working on an issue. \n- `${user}` will be replace with the currently logged in username \n- `${issueNumber}` will be replaced with the current issue number \n- `${sanitizedIssueTitle}` will be replaced with the issue title, with all spaces and unsupported characters (https://git-scm.com/docs/git-check-ref-format) removed",
+		"comment": [
+			"{Locked=']${...}'}",
```
TylerLeonhardt, line 63:
```suggestion
			"{Locked='${...}'}",

```

package.nls.json
```diff
@@ -0,0 +1,205 @@
...
+			"Do not translate what's inside of the '${..}'. It is an internal syntax for the extension"
+		]
+	},
+	"githubIssues.useBranchForIssues.description": {
+		"message": "Determines whether a branch should be checked out when working on an issue. To configure the name of the branch, set `githubIssues.issueBranchTitle`.",
+		"comment": [
+			"{Locked=']`githubIssues.issueBranchTitle`'}",
```
TylerLeonhardt, line 70:
```suggestion
			"{Locked='`githubIssues.issueBranchTitle`'}",

```

package.nls.json
```diff
@@ -0,0 +1,205 @@
...
+			"Do not translate what's inside of the `...`. It is an setting id."
+		]
+	},
+	"githubIssues.useBranchForIssues.on": "A branch will always be checked out when you start working on an issue. If the branch doesn't exist, it will be created.",
+	"githubIssues.useBranchForIssues.off": "A branch will not be created when you start working on an issue. If you have worked on an issue before and a branch was created for it, that same branch will be checked out.",
+	"githubIssues.useBranchForIssues.prompt": "A prompt will show for setting the name of the branch that will be created and checked out.",
+	"githubIssues.issueCompletionFormatScm.markdownDescription": {
+		"message": "Sets the format of issue completions in the SCM inputbox. \n- `${user}` will be replace with the currently logged in username \n- `${issueNumber}` will be replaced with the current issue number \n- `${issueNumberLabel}` will be replaced with a label formatted as #number or owner/repository#number, depending on whether the issue is in the current repository",
+		"comment": [
+			"Do not translate what's inside of the ${...}. It is an internal syntax for the extension."
```
TylerLeonhardt, line 80: this won't be picked up by pseudo language. Your call if that matters or not.

package.nls.json
```diff
@@ -0,0 +1,205 @@
...
+		]
+	},
+	"githubIssues.workingIssueFormatScm.markdownDescription": {
+		"message": "Sets the format of the commit message that is set in the SCM inputbox when you **Start Working on an Issue**. Defaults to `${issueTitle} \nFixes #${issueNumber}`",
+		"comment": [
+			"Do not translate what's inside of the ${...}. It is an internal syntax for the extension."
```
TylerLeonhardt, line 86: this won't be picked up by pseudo language since it's auto-generated. Your call if that matters or not.

package.nls.json
```diff
@@ -0,0 +1,205 @@
...
+	"command.issue.createIssueFromFile.title": "Create Issue",
+	"command.issue.issueCompletion.title": "Issue Completion Chosen",
+	"command.issue.userCompletion.title": "User Completion Chosen",
+	"command.issue.signinAndRefreshList.title": "Sign in and Refresh",
+	"command.issue.goToLinkedCode.title": "Go to Linked Code",
+	"command.issues.openIssuesWebsite.title": "Open on GitHub",
+	"welcome.github.login.contents": {
+		"message": "You have not yet signed in with GitHub\n[Sign in](command:pr.signin)",
+		"comment" : [
+			"Do not translate what's inside of (...). It is link syntax."
```
TylerLeonhardt, line 187:
```suggestion
			"Do not translate what's inside of (...). It is link syntax.",
			"{Locked='](command:pr.signin)'}"
```
in the past translators have messed with the markdown syntax by adding a space between `]` and `(` so I've historically locked those.

package.nls.json
```diff
@@ -0,0 +1,205 @@
...
+		]
+	},
+	"welcome.github.compareChanges.contents": {
+		"message": "The comparing branch has no upstream remote\n[Publish branch](command:git.publish)",
+		"comment" : [
+			"Do not translate what's inside of (...). It is link syntax."
```
TylerLeonhardt, line 193:
```suggestion
			"Do not translate what's inside of (...). It is link syntax.",
			"{Locked='](command:git.publish)'}"
```

package.nls.json
```diff
@@ -0,0 +1,205 @@
...
+	},
+	"welcome.pr.github.uninitialized.contents": "Loading...",
+	"welcome.pr.github.noFolder.contents": "You have not yet opened a folder.",
+	"welcome.pr.github.noRepo.contents": "No git repositories found",
+	"welcome.issues.github.uninitialized.contents": "Loading...",
+	"welcome.issues.github.noFolder.contents": "You have not yet opened a folder.",
+	"welcome.issues.github.noRepo.contents": "No git repositories found",
+	"welcome.github.activePullRequest.contents": "Loading...",
+	"submenu.github.pullRequests.overflow.label": "More actions...",
+	"submenu.github.issues.overflow.label": "More actions..."
```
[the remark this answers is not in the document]
TylerLeonhardt: @alexr00 I assume they were wrapped in `%` in the package.json when you did that?

src/authentication/configuration.ts
```diff
@@ -15,12 +20,13 @@ export const HostHelper = class {
 				return vscode.Uri.parse(testEnv);
 			}
 
+			const yes = vscode.l10n.t('Yes');
 			const result = await vscode.window.showInformationMessage(
-				`The 'GITHUB_TEST_SERVER' environment variable is set to '${testEnv}'. Use this as the GitHub API endpoint?`,
+				vscode.l10n.t('The \'GITHUB_TEST_SERVER\' environment variable is set to \'{0}\'. Use this as the GitHub API endpoint?', testEnv),
```
TylerLeonhardt, line 25: btw you can use backticks with `vscode.l10n.t(...)` doesn't have to use `''`. I just like using `''` by default and then either `""` or backticks when the string contains `'` but totally your call.

src/common/utils.ts
```diff
@@ -9,6 +9,7 @@ import dayjs from 'dayjs';
 import * as relativeTime from 'dayjs/plugin/relativeTime';
 import * as updateLocale from 'dayjs/plugin/updateLocale';
 import { Disposable, Event, Uri } from 'vscode';
+// TODO: localization for webview needed
```
TylerLeonhardt, line 12: Does this file get used by webviews? If so, how does that `vscode` import work?

src/issues/util.ts
```diff
@@ -113,13 +113,13 @@ export function userMarkdown(origin: PullRequestDefaults, user: User): vscode.Ma
 		markdown.appendMarkdown('  \r\n\r\n---');
 	}
 	if (user.location) {
-		markdown.appendMarkdown(`  \r\n$(location) ${user.location}`);
+		markdown.appendMarkdown(`  \r\n${vscode.l10n.t('{0} {1}', '$(location)', user.location)}`);
```
TylerLeonhardt, line 116: doesn't really need localization does it

package.nls.json
TylerLeonhardt: I mislead you hear I think... the `]` in the example was to prevent translators from putting a space between`]` and `(` in the markdown syntax `[]()`

src/issues/util.ts
alexr00: Does this not need localization for RTL languages so that the icon will be on the correct side?

package.nls.json
alexr00: The comment won't get picked up? Seems fine son long as it gets picked up for actual translation.

package.nls.json
alexr00: Seems fine so long as it gets picked up for real translation.

package.nls.json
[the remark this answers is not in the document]
alexr00: Yep:

image

src/authentication/configuration.ts
alexr00: Good to know, thank you!

src/common/utils.ts
alexr00: It does, but we're only using the types from `vscode`, so it's fine today (added the explicit `type` to the import).

src/issues/util.ts
TylerLeonhardt: We don't support [any RTL languages](https://github.com/microsoft/vscode/issues/11770) but that's a good point. Might as well leave it

package.nls.json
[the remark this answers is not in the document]
TylerLeonhardt: Bizarre... ok you skip those for now. Can you get an issue going?

package.nls.json
TylerLeonhardt: Agreed. This is fine.

To summarize, for future us, if you Lock something with the Locked comment syntax, the pseudolanguage generation will honor it.

If you only add a comment like this one, purely describing what to leave be, pseudo won't honor since it's auto-generated and no human looks at it.

TylerLeonhardt (approved):
So exciting!

package.nls.json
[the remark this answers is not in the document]
alexr00: Done: https://github.com/microsoft/vscode/issues/162671

PR #4032, opened 2022-10-10
Fix break from loc API change

alexr00:
@TylerLeonhardt FYI

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -312,7 +312,7 @@ export function registerCommands(
 			} catch (e) {
 				if (e.gitErrorCode === GitErrorCodes.BranchNotFullyMerged) {
 					const action = await vscode.window.showErrorMessage(
-						vscode.l10n.t('The local branch \'{0}\' is not fully merged. Are you sure you want to delete it?', pullRequestModel.localBranchName),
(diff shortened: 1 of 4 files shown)


src/commands.ts
```diff
@@ -312,7 +312,7 @@ export function registerCommands(
 			} catch (e) {
 				if (e.gitErrorCode === GitErrorCodes.BranchNotFullyMerged) {
 					const action = await vscode.window.showErrorMessage(
-						vscode.l10n.t('The local branch \'{0}\' is not fully merged. Are you sure you want to delete it?', pullRequestModel.localBranchName),
+						vscode.l10n.t('The local branch \'{0}\' is not fully merged. Are you sure you want to delete it?', [pullRequestModel.localBranchName]),
```
TylerLeonhardt, line 315: This isn't right. It should work with rest arguments. If it doesn't, that's a bug in localization

TylerLeonhardt (changes requested):

src/commands.ts
TylerLeonhardt: Is localBranchName a string?

TylerLeonhardt (approved):

Issue #4029, opened 2022-10-10
Polish Extension UI


daviddossett:
Creating this polish issue to track a number of quick wins I think I/we can tackle. Some initial ideas:
- Full pass on all webview UI from typography to alignment
- See how we can bridge the gap from GH to VS Code design languages
- Polish comments input element to feel more like other inputs in VS Code 
- Explore introducing webview UI toolkit (ref #3035)

<img width="934" alt="CleanShot 2022-10-09 at 20 03 13@2x" src="https://user-images.githubusercontent.com/25163139/194793907-1871c922-cf0f-480e-8cf0-c2975f407da6.png">

<img width="315" alt="CleanShot 2022-10-09 at 20 03 28@2x" src="https://user-images.githubusercontent.com/25163139/194793899-2cc42f31-769e-48e5-8c00-8af187b2d26c.png">

<img width="558" alt="CleanShot 2022-10-09 at 20 06 59@2x" src="https://user-images.githubusercontent.com/25163139/194794113-45de0f43-ac70-4ff0-b837-7438ceb6ef65.png">

cc @digitarald

digitarald:
cc @alexr00: any areas where you'd like to see design opportunities explored?

@daviddossett one aspect to pull in is how we represent actions, checks and statuses; happening in https://github.com/microsoft/vscode-pull-request-github/issues/681

daviddossett:
That's a good point @alexr00. I think maintaining the general GH layout structure makes enough sense unless we have good reasons to change that.

As for the specific components and elements I feel like we can bias towards VS Code design language as a gut reaction.

It seems like we should also split this effort into two distinct parts:
1) UI polish (mostly CSS + perhaps swapping out components)
2) UX changes (like #681)

For `2`, I think it makes sense to track those in the specific issues. I'll happily help with that work, but I think the main focus of this issue is to polish what's already there.

daviddossett:
Spent some time looking at the description page today to see how we could refresh the overall visual language and fix some responsive issues. This does not yet go into the weeds of how markdown and code is rendered within the feed. Just looking for at the overall impression to start.

A couple of things I'm trying to do:
- Visually align the header and metadata column (assignees, labels, etc.) with the extension detail page
- Clean up layout of header (button placement, typography, PR status tag, etc.)
- Modernize feed, checks, and metadata visual design
- Use Webview UI Toolkit components (buttons, icon buttons—especially for proper dimensions and feedback on hover/focus)

[...]

PR #4050, opened 2022-10-13
Update header UI

daviddossett:
This PR addresses the following:
- Updates header styles and layout to match the design in  #4029 and aligns with extension detail page in core
- Uses extension detail- like buttons instead of mixed icons/large buttons
- Ensures content reflows on small viewport/editor sizes
- Updates max width of overview content
- Refactors header JSX by extracting out `Title`, `Subtitle`, and `ButtonGroup` into their own components.

## Before

CleanShot 2022-10-13 at 11 16 32@2x

## After

CleanShot 2022-10-13 at 11 05 35@2x

## Demo

https://user-images.githubusercontent.com/25163139/195673197-6470f0fc-3760-4f3d-a472-bf03f202d156.mp4

@alexr00 I didn't have the time but I think we should replace the title editing form with a quick pick like how Settings Profile rename works. Thoughts?

cc @digitarald


webviews/components/header.tsx
```diff
@@ -66,50 +67,66 @@ function Title({ title, number, url, canEdit, isCurrentlyCheckedOut, isIssue, re
...
+function Subtitle({ state, isDraft, isIssue, author, base, head }) {
+	return (
+		<div className="subtitle">
+			<div id="status">{getStatus(state, isDraft)}</div>
+			<div className="author">
+				{!isIssue ? <Avatar for={author} /> : null}
+				{!isIssue ? (
+					<div className="merge-branches">
+						<AuthorLink for={author} /> {getActionText(state)} into{' '}
+						<code className="branch-tag"> {base}</code> from <code className="branch-tag"> {head} </code>
```
rzhao271, line 127: Not sure about the spacing here. Is there a difference when there's a space in the code tag and a space outside? And, does the space after `{head}` add padding?

rzhao271 (approved):

webviews/components/header.tsx
```diff
@@ -66,50 +67,66 @@ function Title({ title, number, url, canEdit, isCurrentlyCheckedOut, isIssue, re
...
+	const editableTitle = inEditMode ? titleForm : displayTitle;
+	return editableTitle;
+}
+
+function ButtonGroup({ isCurrentlyCheckedOut, canEdit, isIssue, repositoryDefaultBranch, setEditMode }) {
+	const { refresh, copyPrLink } = useContext(PullRequestContext);
+
+	return (
+		<div className="button-group">
+			<CheckoutButtons {...{ isCurrentlyCheckedOut, isIssue, repositoryDefaultBranch }} />
```
rzhao271, line 100: I'm a bit confused on how the `{...{a, b, c}}` part works. I'd have expected something like `{...[a, b, c]}` considering how the variables were already extracted in the function parameter.

Thomas1664:
A visual distinction between drafts and open PRs like on the GitHub website would be nice, i.e. by making the `open` label green

webviews/components/header.tsx
```diff
@@ -66,50 +67,66 @@ function Title({ title, number, url, canEdit, isCurrentlyCheckedOut, isIssue, re
...
+				</a>
+			</h2>
+		</div>
+	);
+
+	const editableTitle = inEditMode ? titleForm : displayTitle;
+	return editableTitle;
+}
+
+function ButtonGroup({ isCurrentlyCheckedOut, canEdit, isIssue, repositoryDefaultBranch, setEditMode }) {
```
Thomas1664, line 95: Can you move this to its own file? I want to use it in #4057 as well.

daviddossett: Could you create an issue? This button group isn't really generic (yet) and I'm keeping these PRs as scoped as possible.

webviews/components/header.tsx
daviddossett: This syntax was already used in those components so I'll leave this for a follow up PR.

webviews/components/header.tsx
daviddossett: Seems like it had no effect on render but I cleaned it out anyway.

Thomas1664:
@daviddossett In #4052 I informed you that your PR breaks icon button focus outline, but you didn't take my concern serious: https://github.com/microsoft/vscode-pull-request-github/pull/4052#discussion_r996013349

After merging your changes from main into #4057 I noticed that my previously added support for focus is no longer visible. I hope that you take my concern seriously now and fix the issue that was caused by your PR.

mjbvz (approved):

daviddossett:
@Thomas1664 where are you still seeing the issue? It sounded like it was fixed by this PR removing the icons from the header.

Thomas1664:
> @Thomas1664 where are you still seeing the issue? It sounded like it was fixed by this PR removing the icons from the header.

In the comment action bar. But you have to check out #4057 because the action buttons weren't visible before on focus.

daviddossett:
I already fixed it in #4066

alexr00:
> I didn't have the time but I think we should replace the title editing form with a quick pick like how Settings Profile rename works. Thoughts?

I have a slight preference for leaving it as is. When you edit a comment in the overview, you get the text area. Keeping the title the same way is consistent.

Issue #4046, opened 2022-10-12
Does the GHPRI extension need to be * activated?


joyceerhl:
It looks like this came in via https://github.com/microsoft/vscode-pull-request-github/pull/2522. It used to be activated `onStartupFinished`.

IMO the GHPRI extension should only activate (at least in web) when the user
- Is currently on a PR
- Focuses one of GHPRI's views
- Runs one of GHPRI's commands

To achieve this we also need RemoteHub to not force-activate GHPRI unless the user is on a PR.

PR #4051, opened 2022-10-13
Do not * activate the GHPRI extension

joyceerhl:
Part of https://github.com/microsoft/vscode-pull-request-github/issues/4046

It seems like the original intent of adding the * activation event was to improve the perceived speed of loading the PR diff when the current workspace is a PR, but this has the side effect of activating GHPRI even if the user doesn't have a PR checked out. Additionally, when testing this change with sideloaded GHPRI running out of sources in insiders.vscode.dev, it did not seem like the PR diff took much longer than usual to load. 

This change seems safe to me and worthwhile to test out in insiders because if we can avoid unnecessary extension activation, we can reduce network and web worker extension host contention and hopefully improve overall performance of vscode.dev. Please let me know if I've missed some context!

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
 		"vscode.github-authentication"
 	],
 	"activationEvents": [
-		"*",
 		"onCommand:github.api.preloadPullRequest",
 		"onFileSystem:newIssue",
 		"onFileSystem:pr",


package.json
```diff
@@ -30,7 +30,6 @@
 		"vscode.github-authentication"
 	],
 	"activationEvents": [
-		"*",
```
joyceerhl, line 33: Realized that this PR as it stands isn't enough, because GHPRI sets a lot of context keys on activation which hide commands, and all of those context keys probably need to be revisited to ensure that functionality isn't improperly hidden. @alexr00 maybe we can chat about this further when you are in next week?

PR #4057, opened 2022-10-14
Fixes to comment action bar & use bin as delete icon

Thomas1664:
The current delete icon is a cross (`x`) but I think that we associate a cross more with close. Therefore, I think the bin icon expresses 'delete' better than a cross.

For some reason the hygiene script was not happy with my copyright comment. Because I noticed that other PRs didn't add a copyright when editing icons, I chose to do the same.

I also changed the comment action bar to hiding instead of removing itself if the comment is not hovered. Furthermore, the action bar used to not show on focus. I fixed this bug as well.


Thomas1664:
Depends on #4050 because I want to align the secondary button on comments to the right as well because the user needs less eye movement to see both buttons. 

### Before:

image

### After:

image

This would be a lot easier with `<ButtonGroup>`

webviews/editorWebview/index.css
```diff
@@ -415,6 +415,11 @@ body button .icon {
 	margin-right: 6px;
 }
 
+.review-comment-header .avatar {
+	/* Adds up to 8px together with margin from `.review-comment-header > a` */
+	margin-right: 4px;
+}
+
```
daviddossett, line 422: This adds up to more than 8px. Note that these are now different spacings. It looks correct in main as is.

From your branch:

CleanShot 2022-10-14 at 16 31 06@2x

webviews/components/icon.tsx
```diff
@@ -28,3 +28,4 @@ export const diffIcon = <Icon src={require('../../resources/icons/diff.svg')} />
 export const repoIcon = <Icon src={require('../../resources/icons/repo.svg')} />;
 export const gitCompareIcon = <Icon src={require('../../resources/icons/git_compare.svg')} />;
 export const settingsIcon = <Icon src={require('../../resources/icons/settings.svg')} />;
+export const closeIcon = <Icon src={require('../../resources/icons/close.svg')} />;
```
daviddossett, line 31: This breaks the "Remove" action on the sidebar. Could you add `deleteIcon` as a separate icon since we need both?

CleanShot 2022-10-14 at 16 43 31@2x

webviews/editorWebview/index.css
```diff
@@ -220,8 +220,8 @@ body .comment-container .review-comment-header {
 	gap: 8px;
 }
 
-.form-actions > input[type='submit'] {
-	margin-left: auto;
+.status-check-detail-text {
+	margin-left: 0.7em;
```
daviddossett, line 224: This was intentionally removed in a previous PR.

daviddossett (changes requested):
Thanks for the PR! Couple of issues but should be good to go once resolved.

Thomas1664:
Depends on #4066

alexr00 (changes requested):
I'm seeing a misalignment of the buttons when editing a comment:

image

Otherwise, it looks good!

Thomas1664:
> I'm seeing a misalignment of the buttons when editing a comment:

What do you mean by misalignment?

If you are referring to both buttons now being on the right, this is intentional:

> I want to align the secondary button on comments to the right as well because the user needs less eye movement to see both buttons.

alexr00:
@Thomas1664 I'm referring to the lack of padding around the buttons in the bottom right corner. They are right up against the comment thread border.

Thomas1664:
### After:

image

#### Normal comment:

image

daviddossett:
Here's what I'm seeing. The text area is crammed up against the edges, as are the buttons:

CleanShot 2022-10-17 at 14 57 23@2x

The case where a comment is being edited inline is special in that the form actions need different padding rules to fit nicely within their container. I suggest we give this `form` 16px padding to match the normal content rendered in a comment. As a second step, we should remove the `padding-bottom: 10px` from the `form-actions` here to avoid double padding below the form.

CleanShot 2022-10-17 at 14 58 32@2x

PR #4059, opened 2022-10-14
UI fixes for checks section

Thomas1664:
- Make show/hide an actual button
- Add spacing between check name and `details`

### Before:

image

### After:

image


Thomas1664:
Note that I reported the smaller ✔ on the wrapping check name as #4064

webviews/components/merge.tsx
```diff
@@ -65,10 +65,10 @@ const StatusChecks = ({ pr }: { pr: PullRequest }) => {
 					<StateIcon state={status.state} />
 					<div className="status-item-detail-text">
 						<span>{getSummaryLabel(status.statuses)}</span>
-						<a href="javascript:void(0)" aria-role="button" onClick={toggleDetails}>
-							{showDetails ? 'Hide' : 'Show'}
-						</a>
 					</div>
+					<button className='secondary' onClick={toggleDetails} style={{ padding: '0 5px', marginLeft: 'auto' }}>
```
daviddossett, line 69: Let's avoid inline styles. What would be great here would be to extract the styles out of the header's new button group into something like `.small-button` so we can reuse them across both areas.

daviddossett (changes requested):

alexr00 (approved):
Looks good to me, thank you for the PR! Will merge after @daviddossett has also signed off.

daviddossett (approved):
Looks good!

PR #4072, opened 2022-10-17
Commit & Create PR failed to push new commit

### LINKED ISSUE #4055: Commit & Create PR failed to push new commit
Repro:

1. Create a new branch
2. Create a commit on it
3. Publish the branch to the remote
5. Create another commit on it but this time use the `Commit & Create Pull Request` option from the SCM view

🐛 A PR will be created but only containing the first commit. I expected the second commit to have been pushed before creating the PR.

cc @lszomoru

alexr00:
Fixes #4055

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -363,19 +363,36 @@ export function registerCommands(
 		);
 	}
 
-	function isSourceControl(x: any): x is { rootUri: vscode.Uri } {
+	function isSourceControl(x: any): x is Repository {
 		return !!x?.rootUri;
 	}
 
 	context.subscriptions.push(
 		vscode.commands.registerCommand(
 			'pr.create',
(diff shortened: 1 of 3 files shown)


src/commands.ts
```diff
@@ -381,6 +381,23 @@ export function registerCommands(
...
+		vscode.commands.registerCommand(
+			'pr.pushAndCreate',
+			async (args?: any | Repository) => {
+				if (isSourceControl(args)) {
+					const reviewManager = await chooseReviewManager(args.rootUri.fsPath);
+					if (reviewManager) {
+						if (args.state.HEAD?.upstream) {
+							await args.push();
+						}
+						reviewManager.createPullRequest();
```
rzhao271, line 394: Does this part mean that even if the push fails, the PR might be created?

alexr00: If the push fails there will be an uncaught exception, so the PR will not be created.

connor4312 (approved):

PR #4149, opened 2022-11-03
Show "Loading..." in create PR view

### LINKED ISSUE #4100: Improve `Create PR view` startup UX
Every time I create a Pull Request I get a similar UX to this:

Recording 2022-10-24 at 12 27 31

It appears that the view immediately renders to some previous state. In this example I had just created a PR from `joao/tall-elephant` 5 minutes prior, then checked a different branch `joao/widespread-tarantula` and tried creating a PR for it.

Some thoughts:

- If the view is loading, should it report progress somewhere in the UI?
- If the view is loading, should the UI elements be disabled and render some loading state instead?
- Even better, could we be smarter and render what we already know on first render? The local branch name is already known. I guess the loading is for the title and description mostly.

alexr00:
Fixes #4100

### DIFF
--- a/webviews/common/createContext.ts
+++ b/webviews/common/createContext.ts
@@ -17,14 +17,29 @@ const defaultCreateParams: CreateParams = {
 };
 
 export class CreatePRContext {
+	public createParams: CreateParams;
+
 	constructor(
-		public createParams: CreateParams = { ...defaultCreateParams, ...vscode.getState() },
 		public onchange: ((ctx: CreateParams) => void) | null = null,
 		private _handler: MessageHandler | null = null,
 	) {
+		this.createParams = vscode.getState();
 		if (!_handler) {
 			this._handler = getMessageHandler(this.handleMessage);
 		}
+	}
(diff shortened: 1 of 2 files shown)


webviews/common/createContext.ts
```diff
@@ -17,16 +17,31 @@ const defaultCreateParams: CreateParams = {
...
 		private _handler: MessageHandler | null = null,
 	) {
+		this.createParams = vscode.getState();
 		if (!_handler) {
 			this._handler = getMessageHandler(this.handleMessage);
 		}
 	}
 
+	get initialized(): boolean {
+		if (this.createParams.availableBaseRemotes.length !== 0
```
rzhao271, line 33: Are the `!== 0` checks there because the variables are nullable?

alexr00: They're arrays and if the length is 0 then then `this.createParams` hasn't been initialized yet.

rzhao271 (approved):

PR #4160, opened 2022-11-05
Narrow types in TimelineEvent so that it may be treated as a tagged union

joshuaobrien:
This PR narrows the types in `TimelineEvent` a bit by being explicit about the event types. This means that we can treat `TimelineEvent` as a tagged union, which results in improved type safety and fewer lines of code.


alexr00 (approved):
NIce change, thank you!

aeschli (approved):

PR #4217, opened 2022-11-21
Use multi-select quick picks to manage assignees/reviewers

### LINKED ISSUE #4084: Use multi-select quick picks to manage assignees/reviewers
Today we use the gear icon to open a quick pick to add assignees and/or reviewers. However, to remove them, the "x" remove icon button is placed inline next to the person's name in the sidebar. 

This is potentially confusing for two reasons:
- "X" could be confused for review status (i.e. a blocked review)
- Adding and removing are done in the same menu on GitHub.com. 

I would have expected the quick pick counterpart in VS Code would be the same. I.e. it maintains a state and items can be checked/unchecked.

CleanShot 2022-10-19 at 13 36 56@2x

CleanShot 2022-10-19 at 13 40 29@2x

cc @Tyriar

alexr00:
Fixes #4084

### DIFF
--- a/src/github/createPRViewProvider.ts
+++ b/src/github/createPRViewProvider.ts
@@ -368,7 +368,7 @@ export class CreatePullRequestViewProvider extends WebviewViewBase implements vs
 			return;
 		}
 		try {
-			await pr.updateAssignees([resolved]);
+			await pr.addAssignees([resolved]);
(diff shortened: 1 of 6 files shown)


hediet (approved):

daviddossett:
Awesome, thank you for doing this!

PR #4285, opened 2022-12-03
Fix comment layout & use bin as delete icon

Thomas1664:
Continuation of #4057

> The current delete icon is a cross (x) but I think that we associate a cross more with close. Therefore, I think the bin icon expresses 'delete' better than a cross.

> I also changed the comment action bar to hiding instead of removing itself if the comment is not hovered. Furthermore, the action bar used to not show on focus. I fixed this bug as well.

> I want to align the secondary button on comments to the right as well because the user needs less eye movement to see both buttons.


46rouser (approved):

alexr00 (changes requested):
Thanks for the PR! I'm seeing an x overlayed over the bin icon:

image

Other than that, it looks good.

alexr00 (approved):
Thank you!

andreamah (approved):

PR #4286, opened 2022-12-03
Colorize status badge

Thomas1664:
Indicate the status of a PR with different colors and icons. One can now see immediately weither a PR is closed, merged, open or a draft.

Note that you were against colored status badges in #535. However, I think that they are looking substantially different from buttons now and  also have icons in them:

In particular, the differences to buttons are:

- larger
- rounded corners
- different font color
- no outline on focus

Furthermore, I think that it is clear from the context around them that they are status badges and not buttons.

Another concern was that the previous implementation did not respect theme settings. I added the background colors to the color contribution point.


alexr00:
@daviddossett what do you think of the colored status?

daviddossett (changes requested):
Open to this in principle but there are some theming issues to iron out:

https://user-images.githubusercontent.com/25163139/205709555-11634b60-0c34-4a19-9d4f-a029396a0080.mp4

- I believe the color should not be static for all themes. Rather, it should feel like part of that theme. That's why I used `badge.background` if memory serves. The challenge here is that people have strong associations with the GitHub colors. So we need to balance it feeling like the colors fit with the current theme with that association. Open to ideas here.
- Some themes don't seem to render a background color at all

Thomas1664:
@daviddossett

> Some themes don't seem to render a background color at all

This is only the case for high contrast themes and on purpose. I added a border in my latest commit.

> The color is not static for all themes.

Theme authors can change the color using `pullRequests.[PR status]`. Obviously, it seems now static since those new color settings exist only in this PR. IMO it's up to theme authors to adopt these colors in their extension and therefore not our responsibility. On the other hand, vscode could add more colors to badges that reflect a certain state, e.g. some red-ish color for "closed", "failed", or "error" . 

Furthermore, I fixed some issues with the font color on light themes.

webviews/editorWebview/index.css
```diff
@@ -468,11 +468,41 @@ body button .icon {
 #status {
 	box-sizing: border-box;
 	line-height: 18px;
-	background: var(--vscode-badge-background);
-	color: var(--vscode-badge-foreground);
+	color: var(--vscode-foreground);
```
daviddossett, line 471: This renders the text as gray on top of the color, which doesn't look right. `editorInlayHint-foreground` is probably suitable here since that's what I'm doing to get the brighter text on the branch name labels below.

daviddossett: Scratch that, it doesn't always work unless used with its background color counterpart. `button-foreground` is worth a try.

daviddossett (approved):

package.json
```diff
@@ -1965,6 +1965,46 @@
...
+					"highContrast": "editor.background",
+					"highContrastLight": "editor.background"
+				},
+				"description": "The color used for indicating that a pull request is a draft."
+			},
+			{
+				"id": "pullRequests.open",
+				"defaults": {
+					"dark": "#238636",
+					"light": "#238636",
```
alexr00, line 1992: Let's base this off of the `issues.open` color so we don't have two slightly different open colors.
```suggestion
					"dark": "issues.open",
					"light": "issues.open",
```

package.json
```diff
@@ -1965,6 +1965,46 @@
...
+					"highContrast": "editor.background",
+					"highContrastLight": "editor.background"
+				},
+				"description": "The color used for indicating that a pull request is open."
+			},
+			{
+				"id": "pullRequests.closed",
+				"defaults": {
+					"dark": "#da3633",
+					"light": "#da3633",
```
alexr00, line 2002: Same as for `issues.open`:
```suggestion
					"dark": "issues.closed",
					"light": "issues.closed",
```

alexr00 (changes requested):
Thanks for the PR, this looks very nice. I have just two small comments about reusing colors.

alexr00 (approved):
I accepted my suggestions so that the change can be merged. The colors can still be changed before release if there is disagreement.

Thomas1664:
Thanks for taking this PR! I'm ok with the changed colors.

PR #1350, opened 2019-10-02
Update the "Request Changes" and "Approve" buttons alignment

guilherme-gm:
This Pull Request updates the style of the PR Description page to move the `Request Changes` and `Approve` buttons to the right side, near the `Comment` button. This way, those related functionalities are grouped together and far from the `Close Pull Request` button.

**Related Issues**: #297 

Preview:
image

image

PR #4368, opened 2022-12-24
UI fixes for PR view

Thomas1664:
Fixes:

- PR title and number were not in the same line
- Space between the buttons for 'Add PR comment' form is too large

### DIFF
--- a/webviews/components/header.tsx
+++ b/webviews/components/header.tsx
@@ -83,7 +83,7 @@ function Title({ title, titleHTML, number, url, inEditMode, setEditMode, setCurr
 	const displayTitle = (
 		<div className="overview-title">
 			<h2>
-				<div dangerouslySetInnerHTML={{ __html: titleHTML }} />
+				<span dangerouslySetInnerHTML={{ __html: titleHTML }} />
 				{' '}
 				<a href={url} title={url}>
 					#{number}
--- a/webviews/editorWebview/index.css
+++ b/webviews/editorWebview/index.css
@@ -746,10 +746,6 @@ textarea {
 	margin-bottom: 10px;
 }
 
-body .comment-form .form-actions button {
-	margin-right: 10px;
-}
-
 .details .comment-body {
 	padding: 19px 0;
 }


alexr00:
@daviddossett would you like to take this one? I think you may have recently changed some of these things.

webviews/components/sidebar.tsx
```diff
@@ -166,6 +166,7 @@ function Label(label: ILabel & { canDelete: boolean }) {
 				backgroundColor: labelColor.backgroundColor,
 				color: labelColor.textColor,
 				borderColor: `${labelColor.borderColor}`,
+				paddingRight: canDelete ? '2px' : '8px'
```
daviddossett, line 169: Is this to handle PRs where you can't remove the labels yourself? I.e. they don't feature an icon?

daviddossett: Before/after screenshot would be helpful for me here

webviews/components/comment.tsx
```diff
@@ -353,13 +353,12 @@ export function AddComment({
 						{continueOnGitHub ? 'Approve on github.com' : 'Approve'}
 					</button>
 				) : null}
-				<input
+				<button
 					id="reply"
-					value="Comment"
 					type="submit"
-					className="secondary"
```
daviddossett, line 360: @alexr00 Not sure I have the full context here—we're these styled secondary for some reason?

Before/after screenshots would be helpful here and for all other UI changes @Thomas1664

alexr00: I don't actually see a `primary` class in the css files.

alexr00: Styling it as `secondary` appears to have just been to get the alignment correct: https://github.com/microsoft/vscode-pull-request-github/pull/1350

Thomas1664:
@alexr00 @daviddossett I added screenshots for all UI changes

alexr00:
@Thomas1664 David and I will work on reviewing this, but for future PRs it would really help if you could make small, targeted, PRs that only fix one thing at a time. It's difficult to tell which of your changes affect which elements.

webviews/components/comment.tsx
```diff
@@ -206,7 +206,7 @@ function EditComment({ id, body, onCancel, onSave }: EditCommentProps) {
 				<button className="secondary" onClick={onCancel}>
 					Cancel
 				</button>
-				<input type="submit" name="submitButton" value="Save" />
+				<button type="submit" name="submitButton">Save</button>
```
Thomas1664, line 209: inputs are invisible in light high contrast

webviews/components/header.tsx
```diff
@@ -83,7 +83,7 @@ function Title({ title, titleHTML, number, url, inEditMode, setEditMode, setCurr
 	const displayTitle = (
 		<div className="overview-title">
 			<h2>
-				<div dangerouslySetInnerHTML={{ __html: titleHTML }} />
+				<span dangerouslySetInnerHTML={{ __html: titleHTML }} />
```
Thomas1664, line 86: span as `display: inline-block` by default. Prevents wrapping of PR number.

webviews/components/timeline.tsx
```diff
@@ -122,7 +122,7 @@ const ReviewEventView = (event: ReviewEvent) => {
 	const comments = groupCommentsByPath(event.comments);
 	const reviewIsPending = event.state.toLocaleUpperCase() === 'PENDING';
 	return (
-		<div id={reviewIsPending ? 'pending-review' : null} className="comment-container comment">
+		<div id={reviewIsPending ? 'pending-review' : undefined} className="comment-container comment">
```
Thomas1664, line 125: drive-by fix to make IntelliSense happy because `id` is of type `string | undefined` and not `string | null`

Thomas1664 (commented):
@alexr00 I reverted some changes that might have affected many things I and added comments on what individual diffs actually fix if there isn't already some annotation provided by the context of the diff.

daviddossett (dismissed):

webviews/components/timeline.tsx
```diff
@@ -222,32 +222,38 @@ function AddReviewSummaryComment() {
 	const { isAuthor } = pr;
 	const comment = useRef<HTMLTextAreaElement>();
 	return (
-		<div className="comment-form">
+		<form onSubmit={(event) => {
+			event.preventDefault();
+			submit(comment.current!.value);
+		}}>
```
alexr00, line 228: This is personal preference, but I don't like the inconsistency of having the other buttons have an `onClick` but "Submit" being up here in the form's `onSubmit`. Can you move it back the way it was?

alexr00 (changes requested):
Just two small requests!

alexr00 (approved):

alexr00:
Thank you!

PR #4393, opened 2023-01-12
Remove unavailable badge from readme

sravan1946:
lgtm has shut down and the badge can't function anymore, so removing it


alexr00 (approved):
I don't see an obvious replacement either. Thank you for the PR!

Tyriar (approved):

PR #4401, opened 2023-01-17
Adopt git extension API getRefs

alexr00:


src/@types/git.d.ts
```diff
@@ -196,6 +202,8 @@ export interface Repository {
 	getBranches(query: BranchQuery): Promise<Ref[]>;
 	setBranchUpstream(name: string, upstream: string): Promise<void>;
 
+	getRefs(query: RefQuery, cancellationToken?: CancellationToken): Promise<Ref[]>;
```
lszomoru, line 205: I would copy the whole file `git.d.ts` file as it contains an updated signature for the `getBranches()` method since I added support for cancellation tokens.

src/api/api.d.ts
```diff
@@ -201,6 +207,8 @@ export interface Repository {
 	getBranch(name: string): Promise<Branch>;
 	getBranches(query: BranchQuery): Promise<Ref[]>;
 	setBranchUpstream(name: string, upstream: string): Promise<void>;
+	getRefs(query: RefQuery, cancellationToken?: CancellationToken): Promise<Ref[]>;
```
lszomoru, line 210: Same here. I would copy the whole file from the git extension.

src/github/folderRepositoryManager.ts
```diff
@@ -868,7 +868,7 @@ export class FolderRepositoryManager implements vscode.Disposable {
 			return [];
 		}
 
-		const localBranches = this.repository.state.refs
+		const localBranches = (await this.repository.getRefs({}))
```
lszomoru, line 871: Given that we seem to be doing the filtering, I would suggest that we pass a `pattern` here so that we only get the refs that we are interested in. (ex: `getRefs({ pattern: 'refs/heads/*' }));`

src/api/api.d.ts
alexr00: This file contains the API that GHPRI expects extension that register with it as a git provider to implement. Since this includes extensions other than the built in git extension, just copying over the file from the git extension would be misleading. I've copied over a little more of the file, but it is safer to leave it as is.

src/github/folderRepositoryManager.ts
alexr00: Looks like `refs/heads/` works 👍

PR #4420, opened 2023-01-20
Error signing in to Github. Try Again doesn't try again

### LINKED ISSUE #4148: Error signing in to Github. Try Again doesn't try again
- Extension version: v0.54.0
- VSCode Version: 1.73
- OS: macOS 12.3

I'm unable to log in with the extension.

Capture d’écran 2022-11-03 à 15 04 50

The Try Again button doesn't even attempt to open the browser in order to perform the standard login flow.

This is what I did:

1. Install the extension
2. Log in with an account part of an org that enforces SAML without granting that org.
3. Observe the following error message which is expected:

 > Fetching pull requests failed: Resource protected by organization SAML enforcement. You must grant your OAuth token access to this organization.

4. Look through this repo issues to find out how to fix the issue (feature request: please add a button to fix this problem)
5. Revoke the app in github.com
6. Observe "bad credentials" in the extension
7. Uninstall the extension
8. Install the extension again
9. Observe "Error signing in to GitHub"

How can I perform the regular login flow?
Thank you

alexr00:
Fixes #4148

### DIFF
--- a/src/github/credentials.ts
+++ b/src/github/credentials.ts
@@ -206,10 +206,11 @@ export class CredentialStore implements vscode.Disposable {
 		const errorPrefix = vscode.l10n.t('Error signing in to GitHub{0}', getGitHubSuffix(authProviderId));
 		let retry: boolean = true;
 		let octokit: GitHub | undefined = undefined;
+		const sessionOptions: vscode.AuthenticationGetSessionOptions = { createIfNone: true };
 
 		while (retry) {
 			try {
-				await this.initialize(authProviderId, { createIfNone: true });
+				await this.initialize(authProviderId, sessionOptions);
 			} catch (e) {
 				Logger.appendLine(`${errorPrefix}: ${e}`);
 				if (e instanceof Error && e.stack) {
@@ -221,6 +222,9 @@ export class CredentialStore implements vscode.Disposable {
 				retry = false;
 			} else {
 				retry = (await vscode.window.showErrorMessage(errorPrefix, TRY_AGAIN, CANCEL)) === TRY_AGAIN;
+				if (retry) {
+					sessionOptions.forceNewSession = true;
+				}
 			}
 		}
 


jrieken (approved):

dinofx:
@alexr00 is it possible this PR could have introduced an infinite loop for users that have clicked on "Try again?" when using a GH:E token that doesn't have the right access?  We've had a few users today that have clicked on this button, and we've noticed thousands of requests from a few user's IPs to the graphql endpoint here:

https://github.com/microsoft/vscode-pull-request-github/blob/ad954b54bcfe36967a7d0b1948614de6a2bf7e7e/src/github/credentials.ts#L370-L372

alexr00:
@dinofx would these users be using a PAT and then just pasting the same incorrect PAT into VS Code?

dinofx:
I suspect the users might have provided an access token, but maybe it didn't have the right permissions?  But when you click on "Try again", I don't think it prompts you again to change tokens.

I can't confirm the steps to reproduce this.  Our IT department has asked everyone to disable the extension, as it only took a few users with this problem to bring our GH:E to its knees.

dinofx:
Would this change make sense:

```ts
while (retry) {
  retry = false;
  try {
```

so that the loop can only repeat with user interaction?

alexr00:
@dinofx unless I'm missing something the structure of that `while` loop already means that the retry only occurs if there's user interaction. 

How sure are you that this loop/the "Try again" button is what is causing the requests? I tried to force an error in a few ways but I wasn't able to kick off thousands of requests. Are you able to share some debug logs when these requests occur? Steps for getting logs:
1. Set `"githubPullRequests.logLevel": "debug"`
2. Reproduce the many requests issues. 
3. Share the output from "GitHub Pull Request". If you don't want to share these logs publicly then you can email them to me using the email in my bio.

alexr00:
>  Is there any chance the pre-release version was older than the official release on Feb 3rd?

If a user had installed a specific version of the extension or had turned off automatic extension updating then this could happen. But by default no, it shouldn't happen.

alexr00:
@dinofx this maybe related: https://github.com/microsoft/vscode-pull-request-github/issues/4507

PR #4429, opened 2023-01-23
Render permalinked lines in the comments widget

alexr00:
Part of  #4229

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -995,3 +995,15 @@ export class TernarySearchTree<K, V> {
 		}
 	}
 }
+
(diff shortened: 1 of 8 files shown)


src/common/utils.ts
```diff
@@ -995,3 +995,15 @@ export class TernarySearchTree<K, V> {
 		}
 	}
 }
+
+export async function stringReplaceAsync(str: string, regex: RegExp, asyncFn: (substring: string, ...args: any[]) => Promise<string>): Promise<string> {
+	const promises: Promise<string>[] = [];
+	str.replace(regex, (match, ...args) => {
```
connor4312, line 1001: It'd be a little faster for large input to keep track of the offsets in the match and replace them without running the regex again.

connor4312 (approved):

PR #4473, opened 2023-01-27
Fix video and add thank yous

alexr00:


Tyriar (approved):

levrik:
The video doesn't load for me when viewed in VS Code. Unsure if this is an VS Code issue or an issue with the changelog.
Filed https://github.com/microsoft/vscode/issues/173260.

alexr00:
@levrik thanks for the heads up! I think I've fixed it in all places.

PR #4481, opened 2023-01-31
Slow feedback when updating labels / assignees

### LINKED ISSUE #4271: Slow feedback when updating labels / assignees
Testing #4255

When updating multiple labels with the quick pick, it often takes a few seconds for the labels to show up in the PR description view

I think it would be helpful to either:

- Show a progress indicator while labels are being udpated
- Fake the label update in the UI as soon as the quick pick closes, even if GitHub hasn't confirmed the labels yet

alexr00:
Fixes #4271


src/common/utils.ts
```diff
@@ -216,43 +216,22 @@ export interface PromiseAdapter<T, U> {
...
-			return result;
-		},
-		error => {
-			subscription.dispose();
-			throw error;
-		},
-	);
+// Copied from https://github.com/microsoft/vscode/blob/cfd9d25826b5b5bc3b06677521660b4f1ba6639a/extensions/vscode-api-tests/src/utils.ts#L135-L136
+export async function asPromise<T>(event: Event<T>, timeout = 5000): Promise<T> {
+	const error = new Error('asPromise TIMEOUT reached');
```
hediet, line 221: I would create the Error on demand. I could imagine that this improves the stacktrace and even performance, as you don't need to copy the stacktrace all the time.

alexr00: @rebornix you recently moved the error here (https://github.com/microsoft/vscode/pull/155920), was there some reason you moved it so that it isn't created on demand?

alexr00: Discussed with Henning and I've deleted the timeout entirely as it is incorrect in this use case.

hediet (approved):

Issue #4360, opened 2022-12-22
Expansion slow in PULL REQUEST view


isidorn:
The expansion in the PULL REQUEST view is slow because of the GH back and forth. Have we considered doing some requests to GitHub up front so we make the tree expand experience smoother?
Clicking on the `description` node also takes around a second.

For example, could we always request GH data for all visible nodes in advance to their expansion? So as the tree call render on the node (it becomes visible) we could request its expand state. Not sure how aggressive this would be, just an idea that we "predict" what the user might click and get that data in advance...

IMHO this would greatly help in the direction of making the experience butter smooth 🧈 

![Recording 2022-12-22 at 18 25 21](https://user-images.githubusercontent.com/1926584/209191550-d2ec871e-e543-4ce8-8d14-14b608ac6a1b.gif)

alexr00:
How fast would our target be for expanding a PR? I am very hesitant to pre-fetch all the files and comments for all visible PRs since a PR could have many changes in it. 

I did find one call that we could defer slightly, which may save us 10s of milliseconds.

isidorn:
@alexr00 ok. I will be happy to try it out once we have that improvement. 
As for the hesitation to pre-fetch all files and comments for visible PRs - is the concern that pre-fetching might take a long time or that it ends up being unnecessary?

alexr00:
> As for the hesitation to pre-fetch all files and comments for visible PRs - is the concern that pre-fetching might take a long time or that it ends up being unnecessary?

The concern is that we would pre-fetch a lot of stuff that wouldn't be needed. 

Closing, as there has been a small improvement here. We can always re-open if we want to make further perf improvements.

PR #4483, opened 2023-02-01
Stop awaiting comment threads in PRs view

alexr00:
Part of #4360

### DIFF
--- a/src/view/treeNodes/treeNode.ts
+++ b/src/view/treeNodes/treeNode.ts
@@ -18,7 +18,7 @@ export type TreeNodeParent = TreeNode | BaseTreeNode;
 export const EXPANDED_QUERIES_STATE = 'expandedQueries';
 
 export abstract class TreeNode implements vscode.Disposable {
-	children: TreeNode[];
+	protected children: TreeNode[];
 	childrenDisposables: vscode.Disposable[];
 	parent: TreeNodeParent;
 	label?: string;
--- a/src/view/treeNodes/treeNode.ts
+++ b/src/view/treeNodes/treeNode.ts
@@ -5,6 +5,7 @@
 
 import * as vscode from 'vscode';
 import Logger from '../../common/logger';
+import { dispose } from '../../common/utils';
 
 export interface BaseTreeNode {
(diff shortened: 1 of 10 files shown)


src/view/treeNodes/treeNode.ts
```diff
@@ -17,6 +18,7 @@ export type TreeNodeParent = TreeNode | BaseTreeNode;
 export const EXPANDED_QUERIES_STATE = 'expandedQueries';
 
 export abstract class TreeNode implements vscode.Disposable {
+	children: TreeNode[];
```
Tyriar, line 21: It's a little confusing having `children`, `getChildren()` and `cachedChildren()`. Should `children` be private or protected?

Tyriar (approved):

src/view/treeNodes/treeNode.ts
alexr00: Yes, `children` should be protected.

Tyriar (approved):

src/view/treeNodes/treeNode.ts
```diff
@@ -46,7 +48,18 @@ export abstract class TreeNode implements vscode.Disposable {
...
+	async cachedChildren(): Promise<TreeNode[]> {
+		if (this.children && this.children.length) {
+			return this.children;
+		}
+		return this.getChildren();
+	}
+
 	async getChildren(): Promise<TreeNode[]> {
+		if (this.children && this.children.length) {
+			dispose(this.children);
```
joshuaobrien, line 60: Hey @alexr00, sorry for digging up an old PR, but I've got a quick question.

Whilst debugging something for an upcoming PR I ran into this. Why do we dispose of the children here within `getChildren`?

alexr00: With the introduction of `cachedChildren`, we started caching the child nodes. So, when calling `getChildren` instead of `cachedChildren`, we want to recalculate all the child nodes. We should also clear the cache at this time so that we don't end up with old child nodes sticking around. Since the child nodes are disposable, they should also be disposed when they are cleared.

Issue #4490, opened 2023-02-03
Adopt extension logging API


alexr00:
Now that the extension logging API can persist across reload and restart it's time to deprecate the current log level setting and move over to the extension API.

alexr00:
To verify:
1. Install the pre-release GitHub Pull Requests and Issues
2. Open the Output for GitHub Pull Request
3. Run the command "Developer: Set Log Level" and set the log level of GitHub Pull Request to "Trace"
4. Verify that you start seeing trace messages in the output.

PR #4491, opened 2023-02-03
Adopt extension log level API

alexr00:
Fixes https://github.com/microsoft/vscode-pull-request-github/issues/4490

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,8 @@
 						"off"
 					],
 					"default": "info",
-					"description": "%githubPullRequests.logLevel.description%"
+					"description": "%githubPullRequests.logLevel.description%",
+					"deprecationMessage": "%githubPullRequests.logLevel.deprecationMessage%"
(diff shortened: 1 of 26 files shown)


package.json
```diff
@@ -97,7 +97,8 @@
 						"off"
 					],
 					"default": "info",
-					"description": "%githubPullRequests.logLevel.description%"
+					"description": "%githubPullRequests.logLevel.description%",
+					"deprecationMessage": "%githubPullRequests.logLevel.deprecationMessage%"
```
sandy081, line 101: Probably use `markdownDeprecationMessage` so that the set log level command can be rendered as link?

sandy081 (approved):

package.json
alexr00: Good point. Opened https://github.com/microsoft/vscode-pull-request-github/pull/4494

Issue #2771, opened 2021-06-11
Strange avatar is displayed for GitHub Enterprise


think120:
- Extension version:  v0.27.0
- VSCode Version:  Code 1.57.0 
- OS: Win10 1909
- GitHub Enterprise version:  GitHub Enterprise Server 3.0.4 

The avatar of the Pull Request in GitHub Enterprise is strange in "GitHub" and "GitHub Pull Request" tab, as shown in the picture below:
![image](https://user-images.githubusercontent.com/78618719/121686543-3b71b400-cab9-11eb-8a92-f41de9edb321.png)

Looks like it is because in the following postion,  `this.pullRequestModel.userAvatarUri` is used instead of `new vscode.ThemeIcon('github')`
https://github.com/microsoft/vscode-pull-request-github/blob/565d4e6ca5215dfe4b3201f2cbc9d97bd404bbd0/src/view/treeNodes/pullRequestNode.ts#L266-L269

think120:
@alexr00 , I've tested the nightly build, 
1. "GitHub" tab now can display correct avatar
![image](https://user-images.githubusercontent.com/78618719/121695090-a378c800-cac2-11eb-90f9-62b0d8ff3da3.png)

2. For "**GitHub Pull Request**" tab,  still shows the strange avatar
![image](https://user-images.githubusercontent.com/78618719/121695597-1b46f280-cac3-11eb-9c7a-16fcbaa7295a.png)

3. In the `Pull Request Description` page: 

- For the Pull Request created by **myself**, sill shows the strange avatar
![image](https://user-images.githubusercontent.com/78618719/121695745-416c9280-cac3-11eb-9e01-2adc76384aba.png)

- For For the PR created **by other people**,  correct avatar is displayed.
![image](https://user-images.githubusercontent.com/78618719/121695986-8395d400-cac3-11eb-94a4-c5c7027c5456.png)

alexr00:
Thanks for testing this @think120. I've found the source of the bug for the other avatar locations. There will be a new nightly build on Monday. If you could test again with that upcoming build and comment back here about whether the bug is fixed that would be very helpful!

think120:
@alexr00, I've upgraded the nightly build of this extension to v2021.6.19263, but the issue 2 and 3 still exist.  Does that build include your fix? Many thanks!

alexr00:
Yes, that build has the fix. Thanks for testing this out.

think120:
[...]

PR #4497, opened 2023-02-03
Support GitHub Enterprise avatars via Enterprise REST service and optionally Gravatar

kabel:
In GitHub Enterprise, the avatarUrl endpoints that the APIs return require authentication in standard configuration. There is currently no way to do token authentication to these endpoints, so a fallback is necessary to render an image.

The interface was previously coded to handle fallback to the Gravatar service (just like GitHub Desktop [did](https://github.com/desktop/desktop/pull/3513)), but this behavior was disabled in 13ab0a7653008e8433cbac0136d4d8fb97e1328c and 6769dd4cba8b3b6bfef20c94ef28918103634680 because of end-user preference for making this feature configurable.

Recently an enterprise-only REST API was discovered that allows us to proxy the avatar image. This allows us to fetch the binary and encode it as a data URI to inject into the various views that otherwise render the avatar with standard http requests. If this API fails for some reason, Gravatar remains an optional fallback choice.

A new setting `githubPullRequests.defaultGravatarsStyle` has been added of type `enum` with values for each of the Gravatar supported fallback styles. Setting this value to `none` (the default) disables Gravatar integration.

These settings provide a richer user experience for GitHub Enterprise pull requests, allowing users to opt-in/out on avatar rendering via a third-party service with multiple options for the default fallback image style.

For consistency in multiple interface areas where `<Avatar>` is rendered, the GraphQL queries (and matching typescript interfaces) have been updated to include the necessary `email` property for integration with Gravatar.

REST provided actors are missing email. If missing email and using GitHub Enterprise and it's needed by Gravatar, make another REST API call to the user endpoint to get email. This allows a fallback avatar to be generated.

This fixes GitHub Enterprise avatars not being rendered in the tree views too.

The utility functions have been adjusted to support doing the fallback image URL generation directly after the raw results have been loaded from the normal REST and GraphQL APIs. This should be a cleaner implementation.


kabel:
This is a long overdue PR [requested](https://github.com/microsoft/vscode-pull-request-github/issues/2771#issuecomment-877267773) by @alexr00

kabel:
Example Overview render with default settings:
image

kabel:
Looks like I'm referencing outdated information. GitHub Desktop now appears to be using an undocumented GitHub Enterprise REST API endpoint at `/api/v3/enterprise/avatars/u/e?email=`

image

src/github/graphql.ts
```diff
@@ -117,15 +120,26 @@ export interface Commit {
...
-			avatarUrl: string;
-			name: string;
+			avatarUrl: string
+			name: string
+			email: string
+			user: {
+				login: string;
+				avatarUrl: string;
+				url: string;
+				email: string;
```
alexr00, line 141: Same as above: we're now getting all this info twice: once on the `commiter` property and once on the `commiter.user` property.

src/github/graphql.ts
```diff
@@ -680,7 +703,7 @@ export interface GetChecksResponse {
 				nodes: {
 					commit: {
 						statusCheckRollup?: {
-							state: string;
+							state: 'EXPECTED' | 'ERROR' | 'FAILURE' | 'PENDING' | 'SUCCESS';
```
alexr00, line 706: Once again, I do think this is probably a good change, but to keep what needs to be tested for it scoped down could you revert this?

src/github/graphql.ts
```diff
@@ -612,10 +631,11 @@ export interface ContributionsCollection {
...
 	user: {
 		login: string;
-		avatarUrl?: string;
-		bio?: string;
-		company?: string;
-		location?: string;
+		avatarUrl: string;
+		email: string;
+		bio: string;
+		company: string;
```
alexr00, line 637: The `?` on these is correct. A user doesn't have to have these set.

```suggestion
		avatarUrl?: string;
		email?: string;
		bio?: string;
		company?: string;
```

package.json
```diff
@@ -503,6 +503,36 @@
...
 					"default": "firstDiff",
 					"description": "%githubPullRequests.focusedMode.description%"
+				},
+				"githubPullRequests.githubEnterpriseAvatars": {
+					"type": "boolean",
+					"default": true,
+					"description": "%githubPullRequests.githubEnterpriseAvatars.description%",
+					"scope": "window"
+				},
+				"githubPullRequests.defaultGravatarsStyle": {
```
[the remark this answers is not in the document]
kabel: Good question. I took inspiration from Git Lens that has these settings separated because of multiple implementations. I'd be fine either way.

src/github/graphql.ts
```diff
@@ -117,15 +120,26 @@ export interface Commit {
 	id: string;
 	commit: {
 		author: {
+			avatarUrl: string
+			name: string
+			email: string
```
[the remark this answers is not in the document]
kabel: It's possible for the author and/or committer to not be _GitHub_ users, but rather just standard git users. In those cases the email will come from the commit, rather than the GitHub profile.

src/github/graphql.ts
kabel: A user doesn't have to set them, but that will make them null/empty, not undefined/missing, in the API response. From the GraphQL schema, these fields will always return some value. It's one of those cases where null vs undefined mean very different things. Unless I'm reading the schema wrong or there have been schema changes?

image

image

image

image

Example response:
image

src/github/graphql.ts
kabel: Yes. Again, just trying to match it up to the current GraphQL schema.

src/github/utils.ts
```diff
@@ -766,7 +775,7 @@ export function parseGraphQLTimelineEvents(
 					sha: commitEv.commit.oid,
 					author: commitEv.commit.author.user
 						? parseAuthor(commitEv.commit.author.user, githubRepository)
-						: { login: commitEv.commit.committer.name },
+						: { login: commitEv.commit.author.name, avatarUrl: commitEv.commit.author.avatarUrl, email: commitEv.commit.author.email },
```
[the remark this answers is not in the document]
kabel: Good question. It's difficult to understand the UI purpose of this model, but it doesn't really make sense to change the meaning/display of `author` to `committer` only when the `author` is not a _GitHub_ user. Non-GitHub authors _should_ be treated/displayed just like any other user.

src/github/graphql.ts
kabel: See response above. However, it's worth calling out that the `committer` field appears to be unused entirely in the code base/UI.

kabel:
Thanks for the review/questions. 

Based on the availability of the Enterprise REST API (as now used in GitHub Desktop), I added a bit of scope creep to this request. In c0145681df76ef688929ab21181245974db7725e the fallback logic will first check the REST API for the avatar image binary. If the API returns the image binary, it is converted to a data URI to be rendered by the webviews and treeviews. I updated the CSP meta of the webviews to allow `img-src data:` to make that work. If that's too much for this PR, I can save that commit for a future request.

Other comments that don't require additional feedback should now be addressed.

src/github/utils.ts
```diff
@@ -1110,13 +1109,98 @@ export function hasEnterpriseUri(): boolean {
...
+	}
+
+	return `https://www.gravatar.com/avatar/${gravatarId}?s=${size}&d=${style}`;
 }
 
-export function getAvatarWithEnterpriseFallback(avatarUrl: string, email: string | undefined, authProviderId: AuthProvider): string | undefined {
-	return authProviderId === AuthProvider.github ? avatarUrl : (email ? generateGravatarUrl(
-		crypto.createHash('md5').update(email?.trim()?.toLowerCase()).digest('hex')) : undefined);
+const enterpriseAvatarQueue = new PQueue({concurrency: 3});
+const enterpriseAvatarCache: {[k: string]: Promise<string | void>} = {};
```
kabel, line 1132: This only caches enterprise avatars in memory. This may not be desirable if we want to keep memory usage low. There is an opportunity to improve this static-asset caching by saving these images into the extension's storage URI on the filesystem and then having the webviews serve them from that `vscode-resource:` location. That would avoid having to re-fetch the avatars every time the window reloads.

src/github/utils.ts
```diff
@@ -1110,13 +1109,98 @@ export function hasEnterpriseUri(): boolean {
...
+						const dataUri = `data:${resp.headers['content-type']};base64,${Buffer.from(resp.data).toString(
+							'base64',
+						)}`;
+						return dataUri;
+					},
+					() => {
+						delete enterpriseAvatarCache[cacheKey];
+						return;
+					},
+				),
```
alexr00, line 1174: Why is this needed?

alexr00:
I'm sorry @kabel but this PR has become too big for me to continue reviewing now and it mostly seems to be because of fetching the user email when parsing user results from the GitHub REST API. 

I will try to plan time to review in the next couple months, but ideally the PR would be much smaller.

Thank you again for this change. If I hear from users that gravatars are important I will definitely reference your changes to see how to enable them.

alexr00:
I've opened an issue to try to gauge how much desire there is for this feature: https://github.com/microsoft/vscode-pull-request-github/issues/4517

src/github/activityBarViewProvider.ts
```diff
@@ -382,7 +386,7 @@ export class PullRequestViewProvider extends WebviewViewBase implements vscode.W
 <html lang="en">
 	<head>
 		<meta charset="UTF-8">
-		<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src vscode-resource: https:; script-src 'nonce-${nonce}'; style-src vscode-resource: 'unsafe-inline' http: https: data:;">
+		<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src vscode-resource: https: data:; script-src 'nonce-${nonce}'; style-src vscode-resource: 'unsafe-inline' http: https: data:;">
```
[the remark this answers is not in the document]
kabel: It's part of the scope creep I mentioned at https://github.com/microsoft/vscode-pull-request-github/pull/4497#issuecomment-1423080477.

I modified this PR to include grabbing the avatar binary from

1. the (undocumented) GitHub Enterprise REST API for avatars
2. the Gravatar service

The former is really the preferred method of getting avatars as it uses the true system-of-record for the user's preferred image. The latter is an opinion when all else fails. Currently, I have the former integrated by encoding the avatar API response (binary image) into a data URI.

image

src/github/utils.ts
kabel: See https://github.com/microsoft/vscode-pull-request-github/pull/4497#issuecomment-1423080477

src/github/graphql.ts
kabel: I'll change them to `string | null` as the schema defines them.

src/github/utils.ts
```diff
@@ -219,14 +221,26 @@ export function updateCommentReviewState(thread: GHPRCommentThread, newDraftMode
 	});
 }
 
-export function convertRESTUserToAccount(
+export async function convertRESTUserToAccount(
```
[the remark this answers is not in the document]
kabel: Following your advice, I refactored and force-pushed the implementation to be much cleaner and simpler.

src/github/githubRepository.ts
```diff
@@ -509,21 +514,28 @@ export class GitHubRepository implements vscode.Disposable {
...
+								raw.issues.edges.map(async issue => {
+									const parsedIssue = await parseGraphQLIssue(issue.node, this);
+									githubRepository = this.getRepoForIssue(githubRepository, parsedIssue);
+									return new IssueModel(githubRepository, githubRepository.remote, parsedIssue);
+								}),
+							);
+							return { milestone, issues };
+						}),
+					)
+				).filter(x => x !== null) as { milestone: IMilestone; issues: IssueModel[] }[];
```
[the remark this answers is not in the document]
kabel: This was based on how other filtering loops were done in the file, but has since been refactored.

kabel:
Sounds fair @alexr00. I've got the latest push down to a more manageable size (mostly added lines for the new Enterprise REST integration), but I'll wait for more time to become available. Having a better Enterprise experience continues to be important for me, but I can understand how avatars aren't going to be a huge win feature-wise for folks.

PR #4512, opened 2023-02-10
Add Team Reviewers to Pull Request

### LINKED ISSUE #1126: Add Team Reviewers to Pull Request
image

When using the "Add Reviewers" component of the Description view, the autocomplete list of users shows all users available to request review from, but _not_ any of the GitHub Organization teams that are available from the web view.

image

alexr00:
Fixes #1126


src/github/interface.ts
```diff
@@ -41,6 +41,27 @@ export interface IAccount {
...
+export function reviewerId(reviewer: ITeam | IAccount): string {
+	return (reviewer as ITeam).id || (reviewer as IAccount).login;
+}
+
+export function reviewerLabel(reviewer: ITeam | IAccount): string {
+	return (reviewer as ITeam).name || (reviewer as IAccount).login;
+}
+
+export function isTeam(reviewer: ITeam | IAccount): reviewer is ITeam {
+	return (reviewer as ITeam).id !== undefined;
```
hediet, line 62: you could also check ``return "id" in reviewer;`` - then typescript helps you with the types.

hediet (dismissed):

src/github/folderRepositoryManager.ts
```diff
@@ -744,6 +747,44 @@ export class FolderRepositoryManager implements vscode.Disposable {
...
+		} catch (e) {
+			// file doesn't exit
+		}
+		if (!teamReviewersCacheExists) {
+			return undefined;
+		}
+
+		const cache: { [key: string]: ITeam[] } = {};
+		const hasAllRepos = (await Promise.all(this._githubRepositories.map(async (repo) => {
+			const key = `${repo.remote.owner}/${repo.remote.repositoryName}.json`;
```
kabel, line 766: There could be name collisions with enterprise instances.

alexr00: I think we could only get a collision if there are two unrelated enterprises with the same owner + repo name combination. Is that correct?

kabel: Because the cache key makes no distinction between repo origins, yes, two unrelated enterprises might collide. But more importantly and likely would be a collision between `github.com` repos and any enterprise.

Example:
* `github.com/microsoft/vscode`
* `github.contoso.net/microsoft/vscode`

While it may be unlikely that these examples exist and represent distinct teams, they are indeed two different sources of truth for information when querying/caching on the context `microsoft/vscode`

alexr00: Got it, thanks. I'll update the key in another PR.

PR #4530, opened 2023-02-15
Ignore more files from the vsix

kabel:
`.husky` and `tsfmt.json` are currently leaking into the marketplace vsix.


alexr00 (approved):
Thanks for the change!

bpasero (approved):

PR #4539, opened 2023-02-19
Unify style of re-request review button

joshuaobrien:
Fixes #4502

## Before
https://user-images.githubusercontent.com/7654793/219946203-356deb91-b2ff-4a7e-a1fb-6f2a2bf7ba0d.mp4

## After
https://user-images.githubusercontent.com/7654793/219946205-093c0029-964c-4faa-9762-53aea9aa4245.mp4

PR #4540, opened 2023-02-19
Ensure `re-request-review` command is handled in activityBarViewProvider

joshuaobrien:
Whilst working on #4539 I noticed that this command wasn't handled for this view

### DIFF
--- a/src/github/activityBarViewProvider.ts
+++ b/src/github/activityBarViewProvider.ts
@@ -96,6 +96,8 @@ export class PullRequestViewProvider extends WebviewViewBase implements vscode.W
 				return openPullRequestOnGitHub(this._item, (this._item as any)._telemetry);
 			case 'pr.checkout-default-branch':
 				return this.checkoutDefaultBranch(message);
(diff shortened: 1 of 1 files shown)


src/github/activityBarViewProvider.ts
```diff
@@ -113,6 +115,18 @@ export class PullRequestViewProvider extends WebviewViewBase implements vscode.W
 		}
 	}
 
+	private reRequestReview(message: IRequestMessage<string>): void {
```
joshuaobrien, line 118: Copied from https://github.com/joshuaobrien/vscode-pull-request-github/blob/63ef3f541c6ad96e7eba60fec06da0fb0f52bbae/src/github/pullRequestOverview.ts#L891

alexr00: There's likely a lot of duplication between these two files. I'll take it as a debt item to clean them up.

alexr00 (approved):
Thank you for the fix!

PR #4573, opened 2023-02-24
Missing gap between label and dropdowns (macos)

### LINKED ISSUE #4572: Missing gap between label and dropdowns
image

alexr00:
Fixes #4572


jrieken (approved):

bpasero:
This is a flex layout right, I wonder why a hardcoded `min-width` is needed. Can we do a better fix that does not involve guessing font widths?

PR #4584, opened 2023-03-02
Update GHPRI link sharing contributions

joyceerhl:
For https://github.com/microsoft/vscode/issues/175676

This PR enables GHPRI copy link commands to show up in a new context menu that is anchored to editor line numbers. When a command contributed to the editor line number context menu is invoked, extensions will receive both the URI and line number that the copy link command was invoked for. I've modified the link generation code to observe that line number when it's available, instead of using the current selection as it does when invoked from the editor context menu or the command palette.

https://user-images.githubusercontent.com/30305945/222589980-f1640071-cf40-47e3-b1c0-ba2e66633193.mp4

Additionally this PR
- adds Copy github.dev Link when the user is in github.dev or a codespace
- reworks Copy GitHub __ Link to generate github.com links to avoid redundancy
- adds Copy GitHub __ Link actions to all share menus tracked by https://github.com/microsoft/vscode/issues/176316


src/issues/issueFeatureRegistrar.ts
```diff
@@ -1150,19 +1229,17 @@ ${body ?? ''}\n
...
 	}
 
-	async copyHeadLink(fileUri?: vscode.Uri) {
-		const link = await this.getHeadLinkWithError(fileUri);
+	async copyHeadLink(fileUri?: vscode.Uri, includeRange = true) {
+		const link = await this.getHeadLinkWithError(fileUri, includeRange);
 		if (link.permalink) {
-			return vscode.env.clipboard.writeText(
-				link.originalFile ? (await this.getContextualizedLink(link.originalFile, link.permalink)) : link.permalink);
+			return vscode.env.clipboard.writeText(link.permalink);
```
alexr00, line 1242: Why not contextualize this one now?

alexr00 (commented):
Thank you for the PR, I just have a few questions!

src/issues/issueFeatureRegistrar.ts
joyceerhl: Given that we now have vscode.dev commands registered even when in vscode.dev (and github.dev commands registered in github.dev), I think it would be redundant to also offer contextualized GitHub permalinks, so I'm proposing to convert the GitHub permalink actions to always return github.com links. The only scenario where these links are still contextualized is for the github.dev links that GHPRI will register with this PR.

package.json
```diff
@@ -1975,30 +2035,77 @@
...
+					"group": "0_vscode@0"
 				}
 			],
-			"editor/title/context": [
+			"editor/lineNumber/context": [
 				{
 					"command": "issue.copyGithubPermalink",
 					"when": "github:hasGitHubRemotes",
-					"group": "1_cutcopypaste@10"
+					"group": "1_cutcopypaste@1"
```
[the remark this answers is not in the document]
joyceerhl: I inherited this group name from `editor/title/context` so that we can show the Copy vscode.dev Link action in the same group as the GHPRI actions (so that we do not have a single copy action in its own group in the `editor/lineNumber/context` menu). Thanks for asking--I realize I put the Copy github.dev action into the wrong group (should also be `1_cutcopypaste`). The other ones can be renamed to `1_githubPullRequests`.

package.json
```diff
@@ -1347,7 +1378,7 @@
 				},
 				{
 					"command": "pr.copyVscodeDevPrLink",
-					"when": "github:inReviewMode"
+					"when": "github:inReviewMode && remoteName != codespaces"
```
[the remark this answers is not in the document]
joyceerhl: Yeah, the main thing is ensuring that vscode.dev links don't show up in github.dev, and since Codespaces are served from the github.dev domain I think the same expectation applies for vscode.dev links to Codespaces.

package.json
```diff
@@ -1975,30 +2035,77 @@
...
 				},
 				{
 					"command": "issue.copyGithubHeadLink",
+					"when": "github:hasGitHubRemotes && activeEditor == workbench.editors.files.textFileEditor && config.editor.lineNumbers == on",
+					"group": "1_cutcopypaste@5"
+				},
+				{
+					"command": "issue.copyGithubDevLink",
+					"when": "github:hasGitHubRemotes && remoteName == codespaces || github:hasGitHubRemotes && embedderIdentifier == github.dev",
+					"group": "1_cutcopypaste@0"
```
alexr00, line 2064: nit: since this one is `1_cutcopypaste@0` I would put it first.

package.json
```diff
@@ -1975,30 +2035,77 @@
...
+					"group": "1_cutcopypaste@3"
+				},
+				{
+					"command": "issue.copyMarkdownGithubPermalink",
+					"when": "github:hasGitHubRemotes && activeEditor == workbench.editors.files.textFileEditor && config.editor.lineNumbers == on",
+					"group": "1_cutcopypaste@4"
 				},
 				{
 					"command": "issue.copyGithubHeadLink",
+					"when": "github:hasGitHubRemotes && activeEditor == workbench.editors.files.textFileEditor && config.editor.lineNumbers == on",
```
alexr00, line 2058: Are all the `config.editor.lineNumbers == on` conditions needed? I would expect this to be controlled by VS Code core.

alexr00 (approved):
Thanks for the PR! If the answer to https://github.com/microsoft/vscode-pull-request-github/pull/4584#discussion_r1129299349 is yes then feel free to merge!

package.json
joyceerhl: I'm not sure it would make sense for core to handle hiding these commands, since the editor/lineNumber/context menu could also eventually include other extension-contributed items that we'd want to show regardless of line number enablement.

joyceerhl: This behavior also isn't final since I've already gotten feedback that it's unexpected that these commands would go away based on line numbers being toggled off to reduce visual noise.

PR #4590, opened 2023-03-06
Fix setting name in changelog

alexr00:


bpasero (approved):

SamB:
Might want to get this into a release so that the in-app ChangeLog stops showing the wrong message. I nearly reported this until I realized it was only a bug in the ChangeLog, and that it would probably be easier to submit a PR for it than to finish writing the bug report, but when I went to edit the file on GitHub the error was gone. 😅

alexr00:
While it is annoying, it doesn't meet our bar for a recovery release. The `experimental` setting actually still works so if someone copy+pastes it's ok.

PR #4591, opened 2023-03-07
Support checkout PR in vscode.dev

joyceerhl:
For https://github.com/microsoft/vscode/issues/175676

### DIFF
--- a/package.json
+++ b/package.json
@@ -604,6 +604,12 @@
 				"icon": "$(arrow-right)"
 			},
 			{
+				"command": "pr.pickOnVscodeDev",
+				"title": "%command.pr.pickOnVscodeDev.title%",
+				"category": "%command.pull.request.category%",
+				"icon": "$(globe)"
+			},
+			{
 				"command": "pr.exit",
 				"title": "%command.pr.exit.title%",
 				"category": "%command.pull.request.category%"
@@ -1162,6 +1168,10 @@
 					"when": "false"
 				},
 				{
+					"command": "pr.pickOnVscodeDev",
+					"when": "false"
+				},
+				{
 					"command": "pr.exit",
 					"when": "github:inReviewMode"
 				},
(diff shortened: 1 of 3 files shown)


src/commands.ts
```diff
@@ -467,6 +467,35 @@ export function registerCommands(
...
+
+			const fromDescriptionPage = pr instanceof PullRequestModel;
+			/* __GDPR__
+			"pr.checkout" : {
+				"fromDescriptionPage" : { "classification": "SystemMetaData", "purpose": "FeatureInsight" }
+				"location" : { "classification": "SystemMetaData", "purpose": "FeatureInsight" }
+			}
+		*/
+			telemetry.sendTelemetryEvent('pr.checkout', { fromDescription: fromDescriptionPage.toString(), location: 'vscode.dev' });
+
```
alexr00, line 494: Let's not include the telemetry event here and keep it just for when a PR is checked out locally.

package.json
```diff
@@ -1578,17 +1588,22 @@
...
 					"command": "pr.refreshPullRequest",
 					"when": "view == pr:github && viewItem =~ /pullrequest|description/",
-					"group": "pullrequest@2"
+					"group": "2_pullrequest@2"
 				},
 				{
 					"command": "pr.openPullRequestOnGitHub",
 					"when": "view == pr:github && viewItem =~ /pullrequest|description/",
-					"group": "pullrequest@3"
+					"group": "3_pullrequest@3"
```
alexr00, line 1606: image


4 groups(1_, 2_, 3_, and no prefix) seems like a lot given that some onf them only have 1 item in them. Let's move `pr.openPullRequestOnGitHub` into 1_ and not have a 3_. `pr.refreshPullRequest` can just go in `pullrequest@1`

alexr00 (changes requested):

src/commands.ts
joyceerhl: Should we still have some telemetry event for this?

alexr00: We basically already do since we have telemetry for when commands are run.

alexr00 (approved):

meganrogge (approved):

Issue #4598, opened 2023-03-08
Copy GitHub __ Link actions from `file/share` menu copy links with file paths and line ranges


joyceerhl:
1. Open a GitHub repo with the stable version of GHPRI installed
2. Open the File > Share menu
3. Copy GitHub Permalink, Copy GitHub HEAD Link, or Copy GitHub Permalink as Markdown
4. :bug: the active file path and line are included

We special-cased the file menu's behavior for copying vscode.dev links in the builtin GitHub extension and the Remote Repositories extensions to not copy the file path or line, so GHPRI's contributions to `file/share` are inconsistent (this is not a regression from the changes just checked into main). However @rzhao271 and @amunger had feedback that the `file/share` behavior was not what they expected, I think they expected it to relate to the active editor, in which case we would keep GHPRI's current behavior and change the vscode.dev link behavior.

Whatever we do here we should be consistent across all the copy link commands so I'm opening this issue to discuss the desired behavior.

alexr00:
@joyceerhl have you come to any conclusions on whether to include the file + line number when using commands from the File > Share menu?

joyceerhl:
I think we should not include the file and line number for now. However there was a good suggestion to consolidate the share submenus as a single contribution point in https://github.com/microsoft/vscode/issues/176316, and if we did this then we'd be able to remove a lot of the duplicated commands in favor of a single command which behaves differently depending on the passed arguments.

alexr00:
Looks like https://github.com/microsoft/vscode/issues/176316 is in flux. Moving out of the March milestone.

alexr00:
Closing, we'll keep the file and lines.

PR #4599, opened 2023-03-08
`file/share` contribs should not include path or range

joyceerhl:
For https://github.com/microsoft/vscode-pull-request-github/issues/4598

### DIFF
--- a/package.json
+++ b/package.json
@@ -951,12 +951,22 @@
 				"category": "%command.issues.category%"
 			},
 			{
+				"command": "issue.copyGithubPermalinkFile",
+				"title": "%command.issue.copyGithubPermalink.title%",
+				"category": "%command.issues.category%"
+			},
+			{
 				"command": "issue.copyGithubPermalink",
 				"title": "%command.issue.copyGithubPermalink.title%",
 				"category": "%command.issues.category%"
 			},
 			{
 				"command": "issue.copyGithubHeadLink",
+				"title": "%command.issue.copyGithubHeadLink.title%",
+				"category": "%command.issues.category%"
+			},
+			{
+				"command": "issue.copyGithubHeadLinkFile",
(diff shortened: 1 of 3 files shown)


connor4312 (approved):

package.json
```diff
@@ -1418,14 +1433,26 @@
 					"command": "issue.copyGithubPermalink",
 					"when": "github:hasGitHubRemotes"
 				},
+				{
+					"command": "issue.copyGithubPermalinkFile",
```
alexr00, line 1437: Now we will have two commands that show with the same title in the command palette. Is there ever a case where we don't want the file + line included now? There's a proliferation of copy commands and it would be nice to remove all `*File` ones if the differentiation isn't needed.

alexr00 (changes requested):

package.json
joyceerhl: >  Is there ever a case where we don't want the file + line included now? 

My understanding is we don't want the file and line included if the command is invoked from the `file/share` menu--at least, that's the behavior we have for copying vscode.dev links. I don't know if this is extensively used though and I did get feedback that this behavior is unexpected.

package.json
```diff
@@ -1418,14 +1433,26 @@
...
+					"when": "false"
+				},
 				{
 					"command": "issue.copyGithubHeadLink",
 					"when": "github:hasGitHubRemotes"
 				},
+				{
+					"command": "issue.copyGithubHeadLinkFile",
+					"when": "github:hasGitHubRemotes"
```
alexr00, line 1446: This `when` should also be be false

```suggestion
					"when": "false"
```

alexr00 (changes requested):

alexr00:
@joyceerhl is this ok to merge?

alexr00:
@joyceerhl if this is still needed please feel free to re-open!

PR #4614, opened 2023-03-11
fix: label left margin in webview

yiliang114:
Fix Label right margin in webview

### DIFF
--- a/webviews/createPullRequestView/index.css
+++ b/webviews/createPullRequestView/index.css
@@ -74,7 +74,7 @@ body {
 }
 
 .input-label.combo-box {
-	margin-right: 2px;
+	margin-right: 10px;
 	min-width: 48px;
 }
 


yiliang114:
image

alexr00:
@yiliang114 I'm not able to reproduce the issue shown in the screenshot. Do you have any steps I can use to reproduce it?

yiliang114:
> @yiliang114 I'm not able to reproduce the issue shown in the screenshot. Do you have any steps I can use to reproduce it?

@alexr00  Do you mean that it looks like there is no margin on the right side of the label as in the screenshot?  This gif may be helpful. It seems strange that there is no interval. It should not be handled in this way, is it?

bug12

alexr00:
This is what I see:
image

yiliang114:
> This is what I see: image

@alexr00 Does your screenshot on Firefox ?  

I tried several browsers separately: chrome 110.0.5481.177,  Firefox  109.0.1 and Safari 15.3, It seems that chrome's style is not quite right. I feel that it is caused by the original label.

image

PR #4644, opened 2023-03-20
Gracefully handle errors where the SSH configuration file is corrupt or malformed

eastwood:
Developer notes:

- Bug is around some presumptious FP composition which doesn't consider errors thrown from outside environment.
- Have filtered out issues where the config resolution is erroneous.
- Would like to write tests, but the particular code path is not accessible by tests. The singleton appears to be making it difficult and debugging tests hasn't been entirely easy with the singletons.
- ~~Will have a go at injecting the config file dependency so that we test the scenario~~

### DIFF
--- a/src/common/protocol.ts
+++ b/src/common/protocol.ts
@@ -30,12 +30,11 @@ export class Protocol {
 
 	public readonly url: vscode.Uri;
 	constructor(uriString: string) {
+		if (this.parseSshProtocol(uriString)) {
+			return;
+		}
+
 		try {
-
-			if (this.parseSshProtocol(uriString)) {
-				return;
-			}
-
 			this.url = vscode.Uri.parse(uriString);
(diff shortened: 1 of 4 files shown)


eastwood:

alexr00 (approved):
Thanks for the PR! I will merge after the March release.

eastwood:
You're welcome :) I performed some QA on this locally, but had a bit of difficulty hooking into the tests in `protocol.test.ts` due to the singletons. Would you be able to confirm that with a malformed .ssh/config file such as:

```
Host: 
```

The error logs are what you expect? 

This is my first contribution to a vscode extension, so hoping that I used the idiomatic logging mechanism. Criticism is appreciated :)

alexr00:
For this extension, the `Logger` that you used is the best way to do it

aeschli (approved):

PR #4637, opened 2023-03-16
Allow empty labels array to be pushed to set-labels to remove all of them

Balastrong:
This PR fixes #4634

If there are no checked labels from the quick pick menu, the command `set-labels`is not posted, preventing the possibility to remove all (the last one) selected labels.
I didn’t remove the if because `labelsToAdd` can be undefined if the menu is closed with esc, but the length control is causing the bug hence should be removed.

PR #4648, opened 2023-03-20
Allow empty array to be pushed to remove the last label

Balastrong:
Pretty much the same as #4637 but in the PR description view.

Here is less debilitating though, because labels have the X button anyway (which already works fine)

image


alexr00 (approved):
Thank you for the fix!

sandy081 (approved):

PR #4651, opened 2023-03-21
Clean up some when clauses

alexr00:


package.json
```diff
@@ -1943,12 +1918,7 @@
 				{
 					"command": "pr.openReview",
 					"group": "inline@2",
-					"when": "commentController =~ /^github-browse/ && prInDraft"
-				},
-				{
-					"command": "pr.openReview",
-					"group": "inline@2",
-					"when": "commentController =~ /^github-review/ && reviewInDraftMode"
+					"when": "(commentController =~ /^github-browse/ && prInDraft) || (commentController =~ /^github-review/ && reviewInDraftMode)"
```
ulugbekna, line 1921: Just FYI: || has lower precedence than &&, so in principle, you don't need parens here -- but having visual separation makes sense

ulugbekna (approved):
LGTM

package.json
alexr00: I checked the [docs](https://code.visualstudio.com/api/references/when-clause-contexts), but I didn't see anything about the order of operations.

ulugbekna: You're right :-) I'm working on updating and improving our when-clauses docs

Btw, since the new parser comes with stable 1.77, these changes in when-clauses would require that GHPRI is compatible with >= 1.77. I'm not sure whether that's desirable

alexr00: 👍I'm often trying out proposed API here so we already require version 1.77+. Since this is basically debt work I'm going to wait to merge the PR until after endgame as it doesn't meet the bar for being part of the March release.

PR #4695, opened 2023-03-31
Change file mode for execute husky hook on MacOS

### LINKED ISSUE #4662: Husky hooks are not executed
I'm not sure if it's only me, but when I `git commit` I get these warnings on console and husky hooks are not executed:

```
hint: The '.husky/pre-commit' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
```

According to the [answers here](https://stackoverflow.com/questions/8598639/why-is-my-git-pre-commit-hook-not-executable-by-default) a simple `chmod ug+x .husky/*` fixes the issue. I tested it locally and it worked.

Also, after running `chmod` the husky file is marked as changed so probably creating a PR with that should fix the issue for everyone. I can submit it in a PR if this makes sense :)

Balastrong:
This issue fixes #4662 

I'm on MacOS Ventura 13.2.1 and the husky hook is not executed, probably due to a wrong file mode.

```
$ ls -l .husky/pre-commit
-rw-r--r--  1 leonardo  staff  69 Mar 31 18:53 .husky/pre-commit
$ chmod ug+x .husky/*                         
$ ls -l .husky/pre-commit
-rwxr-xr--  1 leonardo  staff  69 Mar 31 18:53 .husky/pre-commit
```

Now that I have added the `x` permission I see the hook is properly executed.


alexr00:
Thanks for the PR!

jrieken (approved):

PR #4766, opened 2023-04-25
Remove "and Submit"

alexr00:


aiday-mar (approved):

alexr00:
@bpasero there is no good localization story for webviews at the moment, so we do not localize any of the webview strings.

PR #4785, opened 2023-04-28
FEAT: "Copy Pull Request Number" command

### LINKED ISSUE #4353: Copy PR number
I have some automation in GitHub that uses the PR number to generate an environment, and I would like to be able to write a launch config to talk to that environment. I can get the PR number for the current pull request from `${command:pr.copyVscodeDevPrLink}`, but I need the bare number. I see a command exists to do this for issues, but not for PRs. Would it be possible to add a command like `copyPrNumber` for this purpose?

aidoskanapyanov:
Implement "Copy Pull Request Number" command.
Closes #4353.

Here's how it looks on the Extension Development Host:
image
image
Here's successfully copied pr number in the clipboard:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -953,6 +953,11 @@
 				"category": "%command.issues.category%"
 			},
 			{
+				"command": "pr.copyPrNumber",
+				"title": "%command.pr.copyPrNumber.title%",
+				"category": "%command.issues.category%"
+			},
+			{
 				"command": "pr.copyVscodeDevPrLink",
 				"title": "%command.pr.copyVscodeDevPrLink.title%",
 				"category": "%command.issues.category%"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -952,6 +952,11 @@
 				"title": "%command.issue.createIssueFromClipboard.title%",
 				"category": "%command.issues.category%"
 			},
+			{
+				"command": "pr.copyPrNumber",
+				"title": "%command.pr.copyPrNumber.title%",
+				"category": "%command.issues.category%"
```
aidoskanapyanov, line 958: Should it be in `%command.pull.request.category%` instead? If so, same should be for `pr.copyVscodeDevPrLink` I think.

PR #4833, opened 2023-05-22
Nudge users towards sign-in

### LINKED ISSUE #4756: Nudge users towards sign-in
Currently 31% of users that have GH PR extension installed in VS Code are NOT logged in to GitHub. Thus they are missing on all the great GH PR extension functionality. This is a large number and we need to look into product changes on the extension side to better nudge users towards sign-in.

One of the simple first steps is to show a badge in the accounts menu when the user is not signed in after installing the GH PR extension. 

@alexr00 you probably have better ideas :) You also shared that historically we did not do this for the vscode.dev flow so users would not get a lot of badges on the Account menu when they open it for the first time. However since majority of GH PR extension is in VS Code desktop we should gear the experience towards that.

fyi @joaomoreno

alexr00:
Fixes #4756

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -217,9 +217,6 @@ async function init(
 
 	await vscode.commands.executeCommand('setContext', 'github:initialized', true);
 
-	const experimentationService = await createExperimentationService(context, telemetry);
-	await experimentationService.initializePromise;
-	await experimentationService.isCachedFlightEnabled('githubaa');
 	registerPostCommitCommandsProvider(reposManager, git);
 	// Make sure any compare changes tabs, which come from the create flow, are closed.
 	CompareChangesTreeProvider.closeTabs();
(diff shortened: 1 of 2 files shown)


rzhao271 (approved):
What happens when undefined is passed to credentialStore? Does the user receive a notification or sign-in dialog?

alexr00:
@rzhao271 you mean here https://github.com/microsoft/vscode-pull-request-github/pull/4833/files#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626R338 ?

The value of `telemetry` should never be undefined at that point.

rzhao271:
Woops, I forgot to tag the line. I meant https://github.com/microsoft/vscode-pull-request-github/pull/4833/files#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626R344

alexr00:
The argument that `credentialStore` takes is just the `vscode.AuthenticationGetSessionOptions`, which is just passed on to the VS Code `getSession` API. The behavior will be as outlined in https://github.com/microsoft/vscode/blob/38f024adf8e43c53a56c42cdb0d5bcd4697b376e/src/vscode-dts/vscode.d.ts#L15662-L15665. So, `undefined` means that a badge will show on the Accounts icon.

PR #4838, opened 2023-05-23
Adopt share provider proposed API

joyceerhl:
For https://github.com/microsoft/vscode/issues/176316

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
 		"codiconDecoration",
 		"diffCommand",
 		"contribCommentEditorActionsMenu",
+		"shareProvider",
 		"quickDiffProvider"
 	],
 	"version": "0.64.0",
--- a/src/issues/shareProviders.ts
+++ b/src/issues/shareProviders.ts
@@ -31,7 +31,7 @@ export class ShareProviderManager implements vscode.Disposable {
 	}
 }
 
-const supportedSchemes = [Schemes.File, Schemes.Review, Schemes.Pr, Schemes.GithubPr, Schemes.VscodeVfs];
+const supportedSchemes = [Schemes.File, Schemes.Review, Schemes.Pr, Schemes.VscodeVfs];
 
 abstract class AbstractShareProvider implements vscode.Disposable, vscode.ShareProvider {
 	private disposables: vscode.Disposable[] = [];
--- a/src/issues/shareProviders.ts
+++ b/src/issues/shareProviders.ts
@@ -9,7 +9,6 @@ import { GitApiImpl } from '../api/api1';
 import { fromReviewUri, Schemes } from '../common/uri';
 import { FolderRepositoryManager } from '../github/folderRepositoryManager';
 import { RepositoriesManager } from '../github/repositoriesManager';
-import { getRepositoryForFile } from '../github/utils';
 import { encodeURIComponentExceptSlashes, getBestPossibleUpstream, getOwnerAndRepo, getSimpleUpstream, getUpstreamOrigin, rangeString } from './util';
 
 export class ShareProviderManager implements vscode.Disposable {
@@ -32,7 +31,7 @@ export class ShareProviderManager implements vscode.Disposable {
 	}
 }
 
-const supportedSchemes = [Schemes.File, Schemes.Review, Schemes.Pr, Schemes.PRNode, Schemes.FileChange, Schemes.GithubPr, Schemes.VscodeVfs, Schemes.Comment];
+const supportedSchemes = [Schemes.File, Schemes.Review, Schemes.Pr, Schemes.GithubPr, Schemes.VscodeVfs];
 
 abstract class AbstractShareProvider implements vscode.Disposable, vscode.ShareProvider {
 	private disposables: vscode.Disposable[] = [];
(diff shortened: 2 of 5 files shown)


src/issues/shareProviders.ts
```diff
@@ -0,0 +1,242 @@
...
+			new GitHubHeadLinkShareProvider(repositoryManager, gitAPI)
+		);
+	}
+
+	dispose() {
+		this.disposables.forEach((d) => d.dispose());
+	}
+}
+
+const supportedSchemes = [Schemes.File, Schemes.Review, Schemes.Pr, Schemes.PRNode, Schemes.FileChange, Schemes.GithubPr, Schemes.VscodeVfs, Schemes.Comment];
```
alexr00, line 35: I don't think we need all these schemes. Specifically, I don't think we need the following:
- `Comment` is from the comment widget. 
- `PRNode` is for a tree item that isn't related to a file. 
- `GitHubPr` is for the diffs you can see while you're creating a PR.
- `FileChange` is only used in a tree item `context` value so I don't actually think we will ever encouter it.

src/issues/shareProviders.ts
```diff
@@ -0,0 +1,242 @@
...
+	private async initialize() {
+		if ((await this.hasGitHubRepositories()) && this.shouldRegister()) {
+			this.register();
+		} else {
+			this.disposables.push(this.gitAPI.onDidOpenRepository(async e => {
+				await e.status();
+				if ((await this.hasGitHubRepositories()) && this.shouldRegister()) {
+					this.register();
+				}
+			}));
```
alexr00, line 66: You can use the folder repo manager events for this: 

https://github.com/microsoft/vscode-pull-request-github/blob/7ea3684f3b991402df8995a505eaca828101976d/src/github/folderRepositoryManager.ts#L150-L154

I think you should be able to get rid of the `status()` call here and in `hasGitHubRepositories` if you use the folder repo manager events.

src/issues/shareProviders.ts
```diff
@@ -0,0 +1,242 @@
...
+	protected abstract shouldRegister(): boolean;
+	protected abstract getBlob(folderManager: FolderRepositoryManager, uri: vscode.Uri): Promise<string>;
+	protected abstract getUpstream(repository: Repository, commit: string): Promise<Remote | undefined>;
+
+	public async provideShare(item: vscode.ShareableItem): Promise<vscode.Uri | undefined> {
+		// Get the repository that this URI is associated with
+		const repository = getRepositoryForFile(this.gitAPI, item.resourceUri);
+		if (!repository) {
+			throw new Error(vscode.l10n.t('Current file does not belong to an open repository.'));
+		}
```
alexr00, line 107: You can get the `repository` with `folderManager.repository`.

src/issues/shareProviders.ts
```diff
@@ -0,0 +1,242 @@
...
+			getOwnerAndRepo(this.repositoryManager, repository, { ...remote, fetchUrl: remote.fetchUrl }),
+			'/blob/',
+			blob,
+			path,
+			range
+		].join(''));
+	}
+}
+
+export class GitHubDevShareProvider extends AbstractShareProvider implements vscode.ShareProvider {
```
alexr00, line 138: Should this handle both github.dev and vscode.dev?

I think we might need a concept of "entry point" for this API. We have a "Copy vscode.dev Link" command that exists in several places. In some of those places (editor context) it includes a path and range. In other places (file share) it does not include the path and range.

alexr00 (approved):
Looks good! I have a few comments but nothing major I think given that we are still using the menu contribution points for now.

src/issues/shareProviders.ts
joyceerhl: Thanks! I think we'd want to support sharing links from inside a diff so I will leave `Schemes.GithubPr` in. I would also like to support share in tree items via a new `shareableItem` optional property in tree items, but that depends on us having support for asynchronous context menu items from Electron, so I will take that out until we have a better story figured out there.

alexr00: PR diffs are `Schemes.Pr` and `Schemes.Review`. `Schemes.GitHubPr` is just the diff you can see when _creating_ a PR.

src/issues/shareProviders.ts
joyceerhl: I was hoping to abstract that away from the providers by having what each provider returns be solely dependent on the `ShareableItem` context that is passed without also needing to account for where actions are being positioned in the UI. For example, in the file/share menu we would pass a workspace folder root URI and in the editor context menu we would pass the document's full URI. It's possible that we might still need to surface a way to exclude providers from certain contributions, but I'd like to avoid that until we have a really compelling reason to add it.

package.json
```diff
@@ -20,6 +20,7 @@
 		"codiconDecoration",
 		"diffCommand",
 		"contribCommentEditorActionsMenu",
+		"shareProvider",
```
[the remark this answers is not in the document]
joyceerhl: 💯 so far all the providers are pretty much guaranteed to be activated by the time that share is invoked.

src/issues/shareProviders.ts
joyceerhl: `onDidLoad` is exactly what I needed! I think I need a specific URI in order to know which folder repo manager to use, so I'm using the equivalent `onDidLoadAnyRepository` event on the parent `RepositoriesManager` class.

src/issues/shareProviders.ts
joyceerhl: > Should this handle both github.dev and vscode.dev?

Right now RemoteHub contributes the vscode.dev links in web and the builtin GitHub extension contributes the vscode.dev links in desktop, which is a bit of a pain from an implementation standpoint to maintain different providers.  If we were to move vscode.dev links into GHPRI then users would only get those links after having both RemoteHub and GHPRI installed on VS Code desktop, so we should be careful about that if we do consolidate.

amunger (approved):

PR #4882, opened 2023-06-02
Cannot assign a specific reviewer

### LINKED ISSUE #4881: Cannot assign a specific reviewer
1. Create a PR, try to assign a specific reviewer
2. :bug: get an error from GHPRI

image

2023-06-01 11:43:27.355 [info] Trying to use globalState for team reviewers.
2023-06-01 11:43:27.466 [error] Unexpected non-whitespace character after JSON at position 2

This doesn't repro in stable vscode.dev so maybe I just have some bad state in globalState for GHPRI in insiders.vscode.dev. Can GHPRI clear out bad globalState so that the extension can recover (I don't want to nuke all my global state in insiders.vscode.dev)?

alexr00:
Fixes #4881


bpasero (approved):
I think I would probably just always log the error when `JSON.parse` fails, not probe on the error message.

Issue #4663, opened 2023-03-21
Share permalink as markdown on tab should use filename for link text


amunger:
Testing microsoft/vscode#177486

open a file and place your cursor somewhere in that file
right click the file tab and 'share github permalink as markdown`
:bug: the generated link will use the text at the cursor location, but seems like it should just be the filename
if the cursor isn't in that file, then the text seems to be picked up from its last location, possibly from another window

amunger:
the command will use the line as text when the editor is focused, and the filename if the editor is not focused.

PR #4886, opened 2023-06-02
Implement Copy Permalink as Markdown share provider

joyceerhl:
For https://github.com/microsoft/vscode/issues/176316

### DIFF
--- a/src/issues/shareProviders.ts
+++ b/src/issues/shareProviders.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as pathLib from 'path';
 import * as vscode from 'vscode';
 import { Commit, Remote, Repository } from '../api/api';
 import { GitApiImpl } from '../api/api1';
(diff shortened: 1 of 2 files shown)


src/issues/shareProviders.ts
```diff
@@ -194,9 +202,49 @@ export class GitHubPermalinkShareProvider extends AbstractShareProvider implemen
...
+				return selectedText;
+			}
+
+			const wordRange = document.getWordRangeAtPosition(item.selection.start);
+			if (wordRange) {
+				return document.getText(wordRange);
+			}
+		}
+
+		return fileName;
```
joyceerhl, line 241: This implementation should address https://github.com/microsoft/vscode-pull-request-github/issues/4663 once we inline Share (which requires async context menus in Electron)

joyceerhl:
@alexr00 it's hidden by default but surfaces in the command palette when you enable `"workbench.experimental.share.enabled": true` and also next to the command center when you enable `"window.commandCenter": true`.

alexr00:
@joyceerhl the 0.66.0 release has happened. It is ok to merge this when you're ready!

joyceerhl:
Thank you @alexr00!

PR #4892, opened 2023-06-05
Fix null when no user name available 

### LINKED ISSUE #4891: User hover shows null when writing the @username
When writing the username of someone that doesn't have a name, just a username, the user hover will show null in place of the name.
image

I happen to look at the code and notice that in the `userHoverProvider` it doesn't show the hover if the user doesn't have a name, maybe it could still show the hover just replace the null with a space
https://github.com/microsoft/vscode-pull-request-github/blob/8396110f2812b3fa040c977d0b227a4f4b3c8fc4/src/issues/userHoverProvider.ts#L62

SKPG-Tech:
This fixes #4891

I'm not sure whether or not it is better to fix this or consider the other solution where it shows users without names on hover, but that's up for decision 😉. For now, it only fixes the null issue 😄.


alexr00 (approved):
Thank you for the PR!

hediet (approved):

PR #4902, opened 2023-06-09
Order of workspaces in multi-root workspace is not what I expect

### LINKED ISSUE #4837: Order of workspaces in multi-root workspace is not what I expect
When I say "create a pull request" I get this picker

image

and vscode is always last even though it is first alphabetically and first in my workspace config file. Other places, like the other git commands, vscode always comes first, and this just throws me off.

precha1989:
Fixes #4837


precha1989 (commented):
@alexr00

PR #4960, opened 2023-06-26
pr: Add back labels feature

hbons:


webviews/common/label.tsx
```diff
@@ -14,18 +14,8 @@ export interface LabelProps {
...
-			style={{
-				backgroundColor: labelColor.backgroundColor,
-				color: labelColor.textColor,
-				borderColor: `${labelColor.borderColor}`,
-				paddingRight: canDelete ? '2px' : '8px'
-			}}
-		>
-			{name}{label.children}
-		</div>
+		<li>{name}{label.children}</li>
```
alexr00, line 19: I just remembered that this `Label` component is also used in the PR description (the full editor that shows all the details of an existing PR). We shouldn't make changes here that will also impact that view.

For now, can you just duplicate the `Label` component in this file and call it `LabelCreate`? Then we can safely modify `LabelCreate`.

alexr00 (changes requested):

webviews/common/label.tsx
hbons: Fixed.

alexr00 (dismissed):
Looks good, feel free to merge when you're ready!

alexr00 (dismissed):

hbons:
Sorry, had to rebase again. :)

alexr00 (approved):

alexr00 (approved):

Issue #5035, opened 2023-07-11
Jump to last diff does not go to the previous page


mgyucht:
- Extension version: v0.68.0
- VSCode Version: 1.80.0-insider
- OS: macOS 13.4.1

Steps to Reproduce:

1. Open a diff from the extension when there are multiple files in the diff. Make sure you're not in the first file of the pull request.
2. Run the "Go to previous diff in pull request" command when focused on the first diff in a file.

![Screenshot 2023-07-11 at 09 41 04](https://github.com/microsoft/vscode-pull-request-github/assets/1850319/33847e0e-4dae-4a65-88d3-b57cffd27477)

I believe this has to do with this line: https://github.com/microsoft/vscode-pull-request-github/blob/main/src/commands.ts#L1371. When going to the previous request, the extension should use `reviewManager.reviewModel.localFileChanges.length - 1 - i` instead of `reviewManager.reviewModel.localFileChanges.length - 1`.

PR #5036, opened 2023-07-11
Correctly iterate backwards through diffs across files

mgyucht:
Fixes https://github.com/microsoft/vscode-pull-request-github/issues/5035.

I'm not exactly sure how to test this change out so would appreciate steps.


alexr00 (approved):
We don't have tests for this, so no further changes are needed. Thank you for the fix!

lszomoru (approved):

PR #5065, opened 2023-07-14
a11y: Shorten separator label

hbons:
VoiceOver was reading out all the dashes. This looks ok until we investigate using a real menu.

image


alexr00:
If the only option is to have the single dash, then does it look better without a separator at all?

meganrogge (commented):
Thanks for making noticing this and making the PR - I am wondering the same thing as Alex

hbons:
Ok, let's remove it for now. We're trying to make this use a proper menu next instead of a `<select>` and we can ask for support for separators there.

PR #5085, opened 2023-07-19
Use the context menu for Create menu

alexr00:


package.nls.json
```diff
@@ -194,6 +194,11 @@
 	"command.issue.createIssueFromSelection.title": "Create Issue From Selection",
 	"command.issue.createIssueFromClipboard.title": "Create Issue From Clipboard",
 	"command.pr.copyVscodeDevPrLink.title": "Copy vscode.dev Pull Request Link",
+	"command.pr.createPrMenuCreate.title": "Create",
+	"command.pr.createPrMenuDraft.title": "Create Draft",
+	"command.pr.createPrMenuSquash.title": "Create & Auto-Squash",
+	"command.pr.createPrMenuMerge.title": "Create & Auto-Merge",
+	"command.pr.createPrMenuRebase.title": "Create & Auto-Rebase",
```
hbons, line 201: I really prefer the "+" for these, it's much easier to scan. :)

roblourens (approved):

package.nls.json
alexr00: Will change to the +. I wasn't paying attention and didn't realize that's what we had before.

PR #5101, opened 2023-07-24
A11y avatar fixes

hbons:
Fix a few things:

- Remove the links around the avatars, which clashes with the click event for the whole list (when trying to open QuickPick).
- Hide elements to VoiceOver that are decoration or not useful.
- Add user's full name on hover over users

PR #5103, opened 2023-07-25
Fix branch button styles

hbons:
This fixes https://github.com/microsoft/vscode-pull-request-github/issues/5104.

The focus border in the Light HC is almost indistinguishable from the normal state, but that is an issue in that theme.

Last row shows focused states:

Group 134

cc @daviddossett


alexr00 (commented):
Looks like this has commits from https://github.com/microsoft/vscode-pull-request-github/pull/5101 in it.

Issue #5104, opened 2023-07-24
Create PR view branch background color is quite dark on light theme


bpasero:
![image](https://github.com/microsoft/vscode/assets/900690/61a999fe-7e46-4365-bfc7-80ca6649da22)

Is the intent here to match with the secondary button color? I wonder if that color should be a bit lighter on light theme?

bpasero:
This is so much better 👏 

<img width="408" alt="image" src="https://github.com/microsoft/vscode-pull-request-github/assets/900690/b23baa81-4418-4140-8ff1-7876711b038d">

PR #5105, opened 2023-07-25
Theme fixes

hbons:
This fixes https://github.com/microsoft/vscode-pull-request-github/issues/5104.

The focus border in the Light HC is almost indistinguishable from the normal state, but that is an issue in that theme.

Last row shows focused states:

Group 134

cc @daviddossett


hbons:
made a new PR as I wasn't able to resolve the conflict.

Issue #5129, opened 2023-07-26
Cannot use keyboard in create PR view to open list of assignees


mjbvz:
From #5115

1. Open create PR view
2. Add an assignee
3. Click into view contents
4. Tab through the PR view sections until the assignee area is highlighted
5. Try pressing space or enter to open the list of assignees

**bug**
Nothing happens

![Screenshot 2023-07-26 at 2 30 50 PM](https://github.com/microsoft/vscode-pull-request-github/assets/12821956/fd5c3538-4fb4-4d3f-b648-8838eea0a285)

PR #5132, opened 2023-07-27
a11y: Allow Enter/Space to open quickpicks

hbons:
Fixes https://github.com/microsoft/vscode-pull-request-github/issues/5129

### DIFF
--- a/webviews/createPullRequestViewNew/app.tsx
+++ b/webviews/createPullRequestViewNew/app.tsx
@@ -151,6 +151,17 @@ export function main() {
 					create();
 				}
 
+				function activateCommand(event: MouseEvent | KeyboardEvent, command: string): void {
+					if (event instanceof KeyboardEvent) {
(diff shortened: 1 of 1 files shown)


webviews/createPullRequestViewNew/app.tsx
```diff
@@ -151,6 +151,17 @@ export function main() {
 					create();
 				}
 
+				function activateCommand(event: MouseEvent | KeyboardEvent, command: string): void {
+					if (event instanceof KeyboardEvent) {
+						if (event.key === "Enter" || event.key === " ") {
```
alexr00, line 156: We should use single quotes here:
```suggestion
						if (event.key === 'Enter' || event.key === ' ') {
```

alexr00 (changes requested):
Thanks for fixing, just one comment!

alexr00 (approved):

PR #5229, opened 2023-09-08
Add a refresh button in the header of comment thread

### LINKED ISSUE #5161: Add an icon button in Comment Thread widget to manually refresh the current thread
**Motivation** I regularly ran into such a problem -- somebody leaves a new comment in a thread from Github web interface, but that comment won't be timely revealed in VSCode until a few minutes later. I know that this extension is polling new updates in the background every several minutes, but it would be nicer to provide a choice for eager people to trigger an update right away.

Thus I propose to add a "refresh" icon button in the comment thread widget (for example, next to the "collapse all comments" icon), on which clicked, latest contents of the current thread are pulled eagerly at the moment. This idea is similar to the one in #4387 .

I am also interested in file a PR for this feature. But before that, I was wondering to know the team's attitude towards this feature. Are there any caveats on this we should care about?

hsfzxjy:
fix #5161

### DIFF
--- a/package.json
+++ b/package.json
@@ -2245,8 +2245,13 @@
 			],
 			"comments/commentThread/title": [
 				{
+					"command": "pr.refreshComments",
+					"group": "0_refresh@0",
+					"when": "commentController =~ /^github-(browse|review)/"
+				},
+				{
 					"command": "pr.collapseAllComments",
-					"group": "collapse@0",
+					"group": "1_collapse@0",
 					"when": "commentController =~ /^github-(browse|review)/"
 				}
 			],


package.json
```diff
@@ -2248,6 +2248,11 @@
 					"command": "pr.collapseAllComments",
 					"group": "collapse@0",
 					"when": "commentController =~ /^github-(browse|review)/"
+				},
+				{
+					"command": "pr.refreshComments",
+					"group": "refresh@0",
```
alexr00, line 2254: The order here seems wrong. I would keep this before the collapse action. 

image

alexr00 (changes requested):
Thanks for the PR! I have one comment.

package.json
alexr00: The order is determined by the `group`, not but the position in the `package.json`.

alexr00 (changes requested):

package.json
hsfzxjy: Sorry for the delay. I've fixed it now.

alexr00 (approved):
Looks good, thanks!

aeschli (approved):

PR #5429, opened 2023-11-02
Fix deprecation messages for createDraft and setAutoMerge

flpcury:
The suggestion in the deprecation messages is to use `githubPullRequests.defaultCreateMethod`, but it should be `githubPullRequests.defaultCreateOption` instead.


alexr00 (approved):
Thanks for the change!

lszomoru (approved):

PR #5435, opened 2023-11-03
add setting to never offer ignoring default branch pr

kurowski:
In my organization's workflow, we _always_ have a PR associated with the default branch. And because we're heavy users of Dev Containers and Codespaces, our vscode workspaces are fairly ephemeral so we're always creating new ones.

Which means that in new workspaces we always see "there's a pull request associated with the default branch, do you want to ignore" notification and have to click the "Don't show" button. And since we create new environments frequently we see this notification frequently.

This pull request adds a setting to disable the offer to ignore PRs on the default branch, which improves our team's developer experience.

### DIFF
--- a/package.json
+++ b/package.json
@@ -329,6 +329,10 @@
 					},
 					"description": "%githubPullRequests.ignoredPullRequestBranches.description%"
 				},
+				"githubPullRequests.neverIgnoreDefaultBranch": {
+					"type": "boolean",
+					"description": "%githubPullRequests.neverIgnoreDefaultBranch.description%"
+				},
 				"githubPullRequests.overrideDefaultBranch": {
 					"type": "string",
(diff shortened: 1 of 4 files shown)


kurowski:
@microsoft-github-policy-service agree company="University of California Education Abroad Program"

alexr00:
@kurowski is there a reason you can't use the existing setting `githubPullRequests.ignoredPullRequestBranches`?

kurowski:
@alexr00 I believe that would work if we _always_ wanted to ignore the pull request on the default branch, but the problem is that we _never_ want to ignore the pull request. I can't figure out a way to get that behavior from the available settings.

alexr00:
I misunderstood, thanks for clarifying!

benibenj (approved):

PR #5437, opened 2023-11-05
Iterate the diffs in each active PR in order

ThomsonTan:
For both `pr.goToNextDiffInPr` and `pr.goToPreviousDiffInPr`, it tries to scroll the next invisible diff section into the editor view, this makes it not iterating each diff one by one. For example, if there are 3 diff sections in the beginning and all visible, call `pr.goToNextDiffInPr` scroll the 4th diff into editor view, this makes it a little hard to traverse each diffs.

This PR sets the cursor to each diff while traverses each diff section one by one, without skipping any diff. The new behavior is also consistent with behavior of commands `workbench.action.compareEditor.nextChange` and `workbench.action.compareEditor.previousChange`.


alexr00 (approved):
Looks good, thanks for the change!

sandy081 (approved):

PR #5542, opened 2023-12-10
Ensure only directories are added to FilesCategoryNode's internal directories array

joshuaobrien:
I noticed that in certain circumstances `GitFileChangeNode`s on the root were getting disposed of and never completely reinitialised. I think this was because they were sometimes being treated as though they were `DirectoryTreeNode`s

image


alexr00:
This change breaks the file tree for PRs where the file that has been changed is at the root: 

image

How it should look:

image

Do you have an example of the issue you were seeing?

CH503822 (approved):

CH503822 (commented):
棒

joshuaobrien:
Ah, it's been a while. I'll open this back up if/when I get a chance to look at the original issue again 👍

PR #5624, opened 2024-01-16
Fixes #5622

### LINKED ISSUE #5622: PR Status References Wrong Foreground Color
- Extension version: 0.78.1
- VSCode Version: 1.85.1
- OS: Irrelevant
- Repository Clone Configuration (single repository/fork of an upstream repository): Irrelevant
- Github Product (Github.com/Github Enterprise version x.x.x): Irrelevant

The extension references the wrong foreground color for the PR Status badges:

I'm noticing this because my theme has colors that are not supposed to be used for the foreground of these badges:

Screenshot 1402-10-26 at 3 21 04 PM

Since the badge colors are customizable, the foreground colors must be as well. I'm going to make a PR that addresses this issue.

pouyakary:
This PR adds foregrounds colors for the PR badges which fixes #5622

### DIFF
--- a/package.json
+++ b/package.json
@@ -2625,6 +2625,16 @@
 				"description": "The color used for indicating that an issue is open."
 			},
 			{
+				"id": "issues.open.foreground",
+				"defaults": {
+					"dark": "#E1FFD9",
+					"light": "#E1FFD9",
+					"highContrast": "editor.background",
+					"highContrastLight": "editor.background"
+				},
+				"description": "The foreground color of badges used for indicating that an issue is open."
+			},
+			{
 				"id": "issues.closed",
 				"defaults": {
 					"dark": "#cb2431",
@@ -2633,6 +2643,16 @@
 					"highContrastLight": "editor.foreground"
 				},
 				"description": "The color used for indicating that an issue is closed."
+			},
+			{
+				"id": "issues.closed.foreground",
+				"defaults": {
+					"dark": "#FFD9D9",
+					"light": "#FFD9D9",
+					"highContrast": "editor.background",
+					"highContrastLight": "editor.background"
+				},
+				"description": "The foreground color of badges used for indicating that an issue is closed."
 			},
 			{
 				"id": "pullRequests.merged",
@@ -2645,6 +2665,16 @@
 				"description": "The color used for indicating that a pull request is merged."
 			},
 			{
+				"id": "pullRequests.merged.foreground",
+				"defaults": {
+					"dark": "#FFD9FE",
+					"light": "#FFD9FE",
+					"highContrast": "editor.foreground",
+					"highContrastLight": "editor.foreground"
+				},
+				"description": "The badge foreground color used for indicating that a pull request is merged."
+			},
+			{
 				"id": "pullRequests.draft",
 				"defaults": {
 					"dark": "#6e7681",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2644,6 +2644,16 @@
 				},
 				"description": "The color used for indicating that a pull request is merged."
 			},
+			{
+				"id": "pullRequests.merged.foreground",
+				"defaults": {
+					"dark": "#FFD9FE",
+					"light": "#FFD9FE",
+					"highContrast": "editor.foreground",
+					"highContrastLight": "editor.foreground"
```
alexr00, line 2653: I'm happy for these to be customizable, but in general it's good to make the colors derive from an existing color. In this case, I think all 4 options here (and for the other new colors) should derive from the current color: `button.foreground`. This will create the least amount of disruption for users, and give an opportunity to for the color to be automatically configured to something that makes sense with their theme.

alexr00: If it's better in your theme, I would also be fine with `button.secondaryForeground`.

alexr00 (changes requested):

package.json
pouyakary: Thanks for the review, and great point! I'll make them as such!

pouyakary: I was thinking, and what about using the original GitHub colors to look exactly like the website? Background colors refer to them, so why not the foregrounds?

alexr00: If you're referring to the open, merged, and draft background colors then matching the website is fine if we don't already do it. I would have to double check and make sure that the colors don't look odd in the default VS Code dark and light themes.

For the foreground, I think it's just black or white on github.com, isn't it?

pouyakary: Yeah and I thought we could default to white/black instead of button foreground since the background color is fixed, it would be dangerous to let the foreground be what the theme says (it will break everywhere)

pouyakary: Can we please have the option to customize all of them (because it would make at least my theme much more pretty) and have all of them default to this color?

alexr00: Sure, that's fine for me. Let's have `pullRequests.statusOpenForeground`, then have the other `pullRequests.status*Foreground` values default to `pullRequests.statusOpenForeground`.

webviews/editorWebview/index.css
```diff
@@ -488,18 +488,22 @@ body button .icon {
 
 .status-badge-merged {
 	background-color: var(--vscode-pullRequests-merged);
+	color: var(--vscode-pullRequests-merged-foreground)
 }
 
 .status-badge-open {
 	background-color: var(--vscode-pullRequests-open);
```
0214ggyy, line 495: wwwwwwwww

package.json
pouyakary: Sure thing, I'll apply it now

pouyakary:
@alexr00 I have applied the changes you requested, I hope you find them acceptable.

alexr00 (changes requested):
@pouyakary the light/dark colors were inverted, so I've swapped them. I also saw that the color wasn't being applied and fixed that too. Looks like the color is still missing from the icon in the badge.

pouyakary:
Oh, sorry :( Let me see if I can fix it

0214ggyy:
在欢乐时，朋友们会认识我们；在患难时，我们会认识朋友。谢谢，来信！！！

PR #5794, opened 2024-02-28
Disable copy commands in embedded editors

mjbvz:
This disables these commands in the interactive playground and chat blocks

In the future, it would be better to instead check that the file belongs to the github repo. That way we'd disable this command in files from virtual file systems. However that is a larger change and would require a custom context key


alexr00 (approved):
Thanks for the PR!

Issue #5931, opened 2024-04-16
Mark files as reviewed from Multi File Diff Editor (Open Changes)


IllusionMH:
When I checkout new PR it automatically opens multi file editor with all changed files. It would be extremely helpful to "weed out" simple/clear etc. changes and mark them as reviewed so I can proceed to review of more important changes quickly.
Ideally it would also mimic GH behavior (collapse reviewed files) when marked as reviewed (let me know if I need to create another issue for this one).

Currently when I pull PR I usually close "Changes in Pull Request #123" tab and switch to browser for this initial check and mark reviewed run on GH directly.

alexr00:
It would be great to be able to programmatically collapse the files. @hediet in case this is something you're considering adding.

alexr00:
@hediet I think this is still not possible for me to add. Is there a way to tell the URI of an open multi diff editor? I'm trying to figure out how to know which multi-diff editors are from GHPR.

ljuba-roofr:
This would be very nice to have

I have `"explorer.autoReveal": false,` because of this issue https://github.com/microsoft/vscode-pull-request-github/issues/4521 . Having that set to `false` makes it really hard to keep track between file I'm viewing in the `Open Changes` and finding that file in pull request sidebar to mark it as viewed

ljuba-roofr:
Is there any update on this? This is one of the missing features that keeps me using the browser

PR #5986, opened 2024-05-02
Add "mark as viewed/unviewed" in multi-diff editor

alexr00:
Part of #5931

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
 		"contribCommentThreadAdditionalMenu",
 		"contribCommentsViewThreadMenus",
 		"contribEditorContentMenu",
+		"contribMultiDiffEditorMenus",
 		"contribShareMenu",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -2389,6 +2390,18 @@
 					"when": "(resourceScheme == pr) || (resourcePath in github:viewedFiles) || (resourcePath in github:unviewedFiles)"
 				}
 			],
+			"multiDiffEditor/resource/title": [
+				{
+					"command": "pr.markFileAsViewed",
+					"group": "navigation",
+					"when": "resourceScheme != pr && resourceScheme != review && resourceScheme != filechange && resourcePath in github:unviewedFiles"
```
alexr00, line 2397: `resourcePath` isn't set here. This is an upstream issue that we need to wait for a fix for

PR #6014, opened 2024-05-23
Fix suggested changes border in comment view not being themeable

robertrossmann:
Adds theme colour reference to a "Suggested change" border to make it less jarring for themes that use different colour palette than the built-in VS Code themes. 🎨 

### Before:
Screenshot 2024-05-23 at 22 43 51

### After:
Screenshot 2024-05-23 at 22 42 59

Theme: [Remedy](https://github.com/robertrossmann/vscode-remedy)

### DIFF
--- a/webviews/editorWebview/index.css
+++ b/webviews/editorWebview/index.css
@@ -93,7 +93,7 @@ body hr {
 	display: block;
 	height: 1px;
 	border: 0;
-	border-top: 1px solid #555;
+	border-top: 1px solid var(--vscode-editorHoverWidget-border);
 	margin: 0 !important;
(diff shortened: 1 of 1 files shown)


alexr00:
To change the color of something in the comments widget you'll need to make a PR in the VS Code repo, not in this repo.

alexr00:
That screenshot doesn't look like any webview I'm aware of! That looks like the VS Code comment widget. The PR description webview shows suggestions quite differently:

image

I think the change in VS Code would be in https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/comments/browser/media/review.css.

PR #6020, opened 2024-05-28
Feature : Add open pr on github from pr description node

### LINKED ISSUE #6000: Allow to open PR on web from PR view when showing multiple
Here I really miss a way to click on the 🌐 icon to open the PR externally:

image

Santhoshmani1:
Closes  #6000 

- Pull requests can now opened on github from pr description nodes

## Preview
Screenshot 2024-05-27 210617


package.json
```diff
@@ -2180,6 +2180,11 @@
 					"group": "inline@2",
 					"when": "view =~ /(pr|prStatus):github/ && viewItem =~ /description/"
 				},
+				{
+					"command": "pr.openPullRequestOnGitHub",
+					"group": "inline@3",
+					"when": "view =~ /(pr|prStatus):github/  && viewItem =~ /description/"
```
alexr00, line 2186: Could we also limit this action to show only when there's more than one active PR? The original issue is that when there's more than one active PR then there's no easy way to open github.com.

alexr00: See here for an example of how to check the count of active PRs:

https://github.com/Santhoshmani1/vscode-pull-request-github/blob/1d12381dd4266025b54ad2942ccea1d4bf890e9b/package.json#L718-L719

alexr00 (changes requested):
Thanks for the PR! You are correct that this was a good place to start. I have one comment:

package.json
Santhoshmani1: I have modified the openPrOnGithub Icon visibility to show only when multiple active pull requests are present.

package.json
```diff
@@ -2180,6 +2180,11 @@
 					"group": "inline@2",
 					"when": "view =~ /(pr|prStatus):github/ && viewItem =~ /description/"
 				},
+				{
+					"command": "pr.openPullRequestOnGitHub",
+					"group": "inline@3",
+					"when": "view =~ /(pr|prStatus):github/  && viewItem =~ /description/ && github:activePRCount >= 2"
```
alexr00, line 2186: One more thing I didn't notice before: I think this should only show in the prStatus view:

```suggestion
					"when": "view == prStatus:github && viewItem =~ /description/ && github:activePRCount >= 2"
```

alexr00 (changes requested):

alexr00 (approved):
Looks good, thank you! I will merge this after the 0.92.0 release.

Tyriar (approved):

PR #6103, opened 2024-07-22
Polish revert PR feature with feedback

alexr00:
and properly clear the active context

### DIFF
--- a/src/github/revertPRViewProvider.ts
+++ b/src/github/revertPRViewProvider.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as vscode from 'vscode';
-import { CreatePullRequestNew } from '../../common/views';
+import { CreateParamsNew, CreatePullRequestNew } from '../../common/views';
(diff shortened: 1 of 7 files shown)


src/github/revertPRViewProvider.ts
```diff
@@ -50,6 +48,33 @@ export class RevertPullRequestViewProvider extends BaseCreatePullRequestViewProv
...
+			return;
+		}
+
+		switch (message.command) {
+			case 'pr.openAssociatedPullRequest':
+				return this.openAssociatedPullRequest();
+			default:
+				// Log error
+				vscode.window.showErrorMessage('Unsupported webview message');
+		}
```
Tyriar, line 75: I think we added a lint rule to require break/return to the vscode repo recently, could help here to prevent accidents

Tyriar (approved):

src/github/revertPRViewProvider.ts
alexr00: It does sound like a good one, but I haven't been able this rule.

PR #6153, opened 2024-08-21
Create suggestions from working tree changes

### LINKED ISSUE #5256: Create suggestions from working tree changes
Hi, thanks for the extension, we love using it.

Wondering what we think of this as a feature. Currently, making a suggestion on a PR is a multi-step process. You highlight the line numbers you want to suggest, click the left "+" on the bottom of the lines, then make your edits without syntax highlighting or Intellisense support.

It would be cool if instead, we could make changes directly to the working tree (creating unstaged changes), and then have a way to "Add suggestions from working tree diff" (or something similar), where it goes through each of the diffs you've made and allows you to transform it into a suggestion with a comment.

This would particularly be helpful for making suggestions on larger PRs, and for making them much faster (just make the change and import rather than click-click-click-type).

Thanks again for creating a great tool :)

alexr00:
Fixes #5256

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -191,6 +191,8 @@
 	"command.review.requestChanges.title": "Request Changes",
 	"command.review.approveOnDotCom.title": "Approve on github.com",
 	"command.review.requestChangesOnDotCom.title": "Request changes on github.com",
+	"command.review.createSuggestionsFromChanges.title": "Create Pull Request Suggestions",
+	"command.review.createSuggestionsFromChange.title": "Convert to Pull Request Suggestion",
 	"command.pr.refreshList.title": "Refresh Pull Requests List",
 	"command.pr.setFileListLayoutAsTree.title": "View as Tree",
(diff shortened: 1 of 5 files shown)


package.nls.json
```diff
@@ -191,6 +191,8 @@
 	"command.review.requestChanges.title": "Request Changes",
 	"command.review.approveOnDotCom.title": "Approve on github.com",
 	"command.review.requestChangesOnDotCom.title": "Request changes on github.com",
+	"command.review.createSuggestionsFromChanges.title": "Create Pull Request Suggestions",
+	"command.review.createSuggestionsFromChange.title": "Convert to Pull Request Suggestion",
```
rzhao271, line 195: Duplicate line?

joyceerhl (approved):

package.nls.json
alexr00: Accidentally included. The next change for the feature needs that line.

PR #6202, opened 2024-08-30
Don't show "make suggestion" when there's already a suggestion in the editor

### LINKED ISSUE #6195: Disallow `Make a suggestion` button press if already have a suggestion in the comment
I often hit this situation

1. Start a PR comment
2. Press `Make a suggestion`
3. Type the suggestion
4. Get confused and think that the `Make a suggestion` button will submit the suggestion. Press it. Retract in shame.

We should disable this button once a suggestion is in the comment.

https://github.com/user-attachments/assets/94223707-45d7-48a1-a60a-8cc121d9259a

alexr00:
Fixes #6195

### DIFF
--- a/package.json
+++ b/package.json
@@ -2507,7 +2507,7 @@
 				{
 					"command": "pr.makeSuggestion",
 					"group": "inline@3",
-					"when": "commentController =~ /^github-(browse|review)/"
+					"when": "commentController =~ /^github-(browse|review)/ && !github:activeCommentHasSuggestion"
 				}
 			],
 			"comments/commentThread/additionalActions": [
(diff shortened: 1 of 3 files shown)


rzhao271 (approved):

mfulton26:
GitHub web UI allows making multiple suggestions in a single comment. With this change will that still be possible? Maybe "Make a Suggestion" should only be disabled when the cursor is inside the suggestion code block/fence.

alexr00:
With this change, the button is disabled, so you can't easily add another suggestion. Do you find yourself often adding multiple suggestions to one comment? What's your use case for this?

mfulton26:
I don't think it is often, maybe every 1 in 30 comments. The main use case is when something needs cleaning-up but there are multiple ways to do it, so I'll leave two (sometimes even three) different suggestions to choose from. I think not limiting functionality in VSCode compared to GitHub UI is important in and of itself but as it isn't a critical thing (plus I can write ` ```suggestion ` out myself manually if needed) then I don't think it is critical to fix but it would be nice IMO (to have parity with GitHub UI, etc.).

Issue #3059, opened 2021-10-22
quote reply doesn't handle nested replies


diminutivesloop:
Issue Type: <b>Bug</b>

Steps to Reproduce:
1. Click the "Quote Reply" button for a PR comment that quote-replies to another comment.
2. Click the "Comment" button.

Result:
The original comment is nested inside two quote boxes, while the previous quote-reply text is unquoted.
<img width="703" alt="Screen Shot 2021-10-22 at 1 11 41 PM" src="https://user-images.githubusercontent.com/667716/138496108-23834cb1-6f8e-4d4c-a212-1750cdecfca1.png">

Extension version: 0.31.1
VS Code version: Code 1.61.2 (Universal) (6cba118ac49a1b88332f312a8f67186f7f3c1643, 2021-10-19T15:49:28.381Z)
OS version: Darwin x64 20.6.0
Restricted Mode: No

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz (12 x 2600)|
|GPU Status|2d_canvas: enabled<br>gpu_compositing: enabled<br>metal: disabled_off<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|3, 3, 3|
|Memory (System)|32.00GB (0.16GB free)|
|Process Argv|--crash-reporter-id ed50f89e-497f-43cd-86f3-a73119f730ed|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368cf:30146710
vsreu685:30147344
python383cf:30185419
pythonvspyt602:30300191
vspor879:30202332
vspor708:30202333
vspor363:30204092
pythonvspyt639:30300192
pythontb:30283811
pythonptprofiler:30281270
vshan820:30294714
vstes263cf:30335440
vscoreces:30384385
pythondataviewer:30285071
pythonvsuse255:30340121
vscod805cf:30301675
pythonvspyt200:30340761
binariesv615:30325510
vsccppwt:30382697
pythonvssor306:30344512
bridge0708:30335490
pygetstartedt3:30385195
dockerwalkthru:30377721
bridge0723:30353136
[...]

PR #6230, opened 2024-09-17
fix quote reply

ixzhao:
Related to issue #3059 .
There is PR #3148. However, that doesn't seem to work for me.


alexr00 (approved):
Thank you!

Tyriar (approved):

PR #6258, opened 2024-10-02
Adding a new Review doesn't update the Tree Control Node.

### LINKED ISSUE #6251: Adding a new Review doesn't update the Tree Control Node.
1. Create a new Review item.
2. The code shows a comment icon, but the tree node doesn't update.

To show the comment in the tree control, the control must be reloaded.

Conversely, deleting the last comment in a file doesn't remove the comment icon from the tree node either.

alexr00:
Fixes #6251


src/view/commentDecorationProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+				const pr = repo.pullRequestModels.get(query.prNumber);
+				if (pr && pr.reviewThreadsCache.some(c => c.path === query.fileName)) {
+					return {
+						propagate: false,
+						tooltip: 'Commented',
+						// allow-any-unicode-next-line
+						badge: '💬',
+					};
+				}
+			}
```
hediet, line 48: The return does not seem to depend on `repo` or `pr`, thus I'd do it like this:

```suggestion
                        const isCommented = folderManager.gitHubRepositories.some(repo => {
                            const pr = repo.pullRequestModels.get(query.prNumber);
                            return pr && pr.reviewThreadsCache.some(c => c.path === query.fileName);
                        })
                        if (isCommented) {
			    return {
				    propagate: false,
				    tooltip: 'Commented',
				    // allow-any-unicode-next-line
				    badge: '💬',
			    };
                        }
```

This reduces nesting a bit and makes the control flow and data flow easier to understand (control flow, because mentally for a `for`, we have to iterate through the loop to understand how the `return` works, but for `some`, we don't care about execution order; data flow because we see immediately that the return does not depend on previous for iterations/repo or pr).

hediet (approved):

Tyriar (approved):

PR #6275, opened 2024-10-11
Use `CommentThreadCollapsibleState` instead of calling hide

aiday-mar:
Use `CommentThreadCollapsibleState` instead of calling hide


alexr00 (approved):
Thank you!

PR #6283, opened 2024-10-16
Extending summarization tool to pull requests also

aiday-mar:
Extending summarization tool to pull requests also

This is a PR which touches upon the Fetch tool as well as the summarization tool. It will be broken down into two separate PRs.

### DIFF
--- a/src/lm/tools/summarizeTool.ts
+++ b/src/lm/tools/summarizeTool.ts
@@ -1,0 +1,105 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+'use strict';
+
+import * as vscode from 'vscode';
+import { InMemFileChange } from '../../common/file';
+import { FolderRepositoryManager } from '../../github/folderRepositoryManager';
+import { IssueModel } from '../../github/issueModel';
+import { PullRequestModel } from '../../github/pullRequestModel';
+import { RepositoriesManager } from '../../github/repositoriesManager';
+import { concatAsyncIterable } from './toolsUtils';
+
+interface SummarizationToolParameters {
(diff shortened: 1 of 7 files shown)


src/lm/tools/summarizeTool.ts
```diff
@@ -0,0 +1,105 @@
...
+import * as vscode from 'vscode';
+import { InMemFileChange } from '../../common/file';
+import { FolderRepositoryManager } from '../../github/folderRepositoryManager';
+import { IssueModel } from '../../github/issueModel';
+import { PullRequestModel } from '../../github/pullRequestModel';
+import { RepositoriesManager } from '../../github/repositoriesManager';
+import { concatAsyncIterable } from './toolsUtils';
+
+interface SummarizationToolParameters {
+	issueNumber: number;
```
alexr00, line 16: The issue/pr should have already been fetched with the fetch tool. I would expect the parameter here to be the `FetchResult` (title, body, comments), not an issue number.

src/lm/tools/summarizeTool.ts
```diff
@@ -0,0 +1,105 @@
...
+		if (!folderManager || !owner || !name) {
+			throw new Error(`No folder manager found for ${owner}/${name}. Make sure to have the repository open.`);
+		}
+		let issueOrPullRequest: IssueModel | PullRequestModel | undefined = await folderManager.resolveIssue(owner, name, options.parameters.issueNumber, true);
+		if (!issueOrPullRequest) {
+			issueOrPullRequest = await folderManager.resolvePullRequest(owner, name, options.parameters.issueNumber);
+		}
+		if (!issueOrPullRequest) {
+			throw new Error(`No issue or PR found for ${owner}/${name}/${options.parameters.issueNumber}. Make sure the issue or PR exists.`);
+		}
```
alexr00, line 49: Since the input to this tool should be an issue, we shouldn't need to fetch the issue again.

alexr00 (changes requested):

PR #6284, opened 2024-10-16
Extending the fetching of issues to PRs in the Fetch tool

aiday-mar:
Extending the fetching of issues to PRs in the Fetch tool

### DIFF
--- a/src/lm/tools/fetchTool.ts
+++ b/src/lm/tools/fetchTool.ts
@@ -6,9 +6,11 @@
 
 import * as vscode from 'vscode';
 import { FolderRepositoryManager } from '../../github/folderRepositoryManager';
+import { IssueModel } from '../../github/issueModel';
+import { PullRequestModel } from '../../github/pullRequestModel';
 import { RepositoriesManager } from '../../github/repositoriesManager';
 
-interface IssueToolParameters {
+interface FetchToolParameters {
 	issueNumber: number;
 	repo?: {
 		owner: string;
@@ -16,7 +18,7 @@ interface IssueToolParameters {
 	};
 }
 
-interface IssueResult {
+interface FetchResult {
 	title: string;
 	body: string;
 	comments: {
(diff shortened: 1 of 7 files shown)


src/lm/tools/fetchTool.ts
```diff
@@ -0,0 +1,70 @@
...
+			body: issueOrPullRequest.body,
+			comments: issueOrPullRequest.item.comments?.map(c => ({ body: c.body })) ?? []
+		};
+		return {
+			'text/plain': JSON.stringify(result)
+		};
+	}
+}
+
+export async function fetchIssueOrPR(options: vscode.LanguageModelToolInvocationOptions<FetchToolParameters>, repositoriesManager: RepositoriesManager): Promise<PullRequestModel | IssueModel> {
```
alexr00, line 45: This seems specific to the fetch tool. Could it be a private method in the `FetchTool` class?

aiday-mar:
Has merge conflicts, need to resolve them

alexr00 (commented):
There are some merge conflicts from my PR feedback changes.

src/lm/tools/fetchTool.ts
aiday-mar: Yes initially I extracted it because I wanted to reuse it for the summarization tool. My understanding however now is that we'd like to make the summarization use as input the output of the fetch tool? I am wondering if doing this could give wrong outputs sometimes as the LLM will not make the link between fetching before summarizing.

alexr00: You are correct, that can happen. In that case, we look through the previous chat messages to find relevant information. You can see an example of that here:

https://github.com/microsoft/vscode-pull-request-github/blob/abbceb87c038156493d4be3959a94e984740525a/src/lm/tools/displayIssuesTool.ts#L109-L128

To be clear, I don't know if looking through the history is best thing to do, but it's effective. I plan to see if I can tweak prompts to try to get the outputs to get passed to inputs better.

alexr00 (approved):

PR #6291, opened 2024-10-16
Adding summarization tool which is based on fetch tool

aiday-mar:
Adding summarization tool which is based on fetch tool


lszomoru (approved):

alexr00 (commented):
👏

PR #6298, opened 2024-10-17
Adding mark as read action in the notifications tree view

aiday-mar:
Adding mark as read action in the notifications tree view

### DIFF
--- a/package.json
+++ b/package.json
@@ -2330,17 +2330,17 @@
 				{
 					"command": "notification.chatSummarizeNotification",
 					"group": "inline@1",
-					"when": "view == notifications:github && (viewItem == 'Issue' || viewItem == 'PullRequest')"
+					"when": "view == notifications:github && (viewItem == 'Issue' || viewItem == 'PullRequest') && config.githubPullRequests.experimental.notificationsView"
 				},
 				{
 					"command": "notification.openOnGitHub",
 					"group": "inline@2",
-					"when": "view == notifications:github && (viewItem == 'Issue' || viewItem == 'PullRequest')"
+					"when": "view == notifications:github && (viewItem == 'Issue' || viewItem == 'PullRequest') && config.githubPullRequests.experimental.notificationsView"
 				},
 				{
 					"command": "notification.markAsRead",
 					"group": "inline@3",
-					"when": "view == notifications:github && (viewItem == 'Issue' || viewItem == 'PullRequest')"
+					"when": "view == notifications:github && (viewItem == 'Issue' || viewItem == 'PullRequest') && config.githubPullRequests.experimental.notificationsView"
 				},
 				{
 					"command": "pr.openDescriptionToTheSide",
--- a/package.json
+++ b/package.json
@@ -1516,6 +1516,12 @@
 				"title": "%command.notification.chatSummarizeNotification.title%",
 				"category": "%command.notifications.category%",
 				"icon": "$(copilot)"
+			},
+			{
+				"command": "notification.markAsRead",
+				"title": "%command.notifications.markAsRead.title%",
+				"category": "%command.notifications.category%",
+				"icon": "$(checkmark)"
 			}
 		],
 		"viewsWelcome": [
@@ -2165,6 +2171,10 @@
 					"when": "false"
 				},
 				{
+					"command": "notification.markAsRead",
+					"when": "false"
+				},
+				{
 					"command": "notification.chatSummarizeNotification",
 					"when": "false"
 				}
@@ -2326,6 +2336,11 @@
 					"command": "notification.openOnGitHub",
 					"group": "inline@2",
 					"when": "view == notifications:github && (viewItem == 'Issue' || viewItem == 'PullRequest')"
+				},
+				{
+					"command": "notification.markAsRead",
+					"group": "inline@3",
+					"when": "view == notifications:github"
 				},
 				{
 					"command": "pr.openDescriptionToTheSide",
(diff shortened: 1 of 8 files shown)


src/notifications/notificationsProvider.ts
```diff
@@ -61,6 +61,8 @@ export class NotificationsProvider implements vscode.Disposable {
 		this._notificationsManager.clear();
 	}
 
+	public markAsRead(notification: INotificationItem): void { }
```
alexr00, line 64: Does the action do anything now?

package.json
```diff
@@ -2327,6 +2337,11 @@
 					"group": "inline@2",
 					"when": "view == notifications:github && (viewItem == 'Issue' || viewItem == 'PullRequest')"
 				},
+				{
+					"command": "notification.markAsRead",
+					"group": "inline@3",
+					"when": "view == notifications:github"
```
alexr00, line 2343: I guess we don't need a "is unread" in this `when` because we don't show read notifications?

package.json
```diff
@@ -2327,6 +2337,11 @@
 					"group": "inline@2",
 					"when": "view == notifications:github && (viewItem == 'Issue' || viewItem == 'PullRequest')"
 				},
+				{
+					"command": "notification.markAsRead",
+					"group": "inline@3",
```
alexr00, line 2342: In general, we try to make sure any action that's added `inline` is also added to the right click context menu.

alexr00 (approved):

aiday-mar:
Hi thanks for the feedback. Yes apologies I forgot to mark the PR as draft so I was still working on it.

package.json
aiday-mar: That's right, currently we only show unread notifications.

aiday-mar: Perhaps it would make sense to show also read notifications at some point 🤔

src/notifications/notificationsProvider.ts
aiday-mar: Yep going to fill that in.

package.json
aiday-mar: Hi I am not quite sure I understand. Do you mean I should add the command `notification.markAsRead` elsewhere too?

alexr00: We have the inline actions (circled), and the right click actions (squared). If an action is in the inline actions, we usually also put it in the right click actions.

image

src/notifications/notificationsProvider.ts
```diff
@@ -61,12 +63,31 @@ export class NotificationsProvider implements vscode.Disposable {
...
+		}
+		await gitHub.octokit.call(gitHub.octokit.api.activity.markThreadAsRead, {
+			thread_id: Number(notification.notification.id)
+		});
+		if (this._notifications === undefined) {
+			return;
+		}
+		const notificationIndexToRemove = this._notifications.findIndex(n => n === notification);
+		if (notificationIndexToRemove !== -1) {
+			this._notifications.splice(notificationIndexToRemove, 1);
```
alexr00, line 79: Having a `_notificaitons` member will cause use to be out of sync with the `_notificationsManager`. We should access and manage notifications exclusively through `this._notificationsManager`. I two options:
- Get the notification from the `this._notificationsManager` and set it's `unread` property to `false`. Then, when we [get all notifications in the tree view](https://github.com/microsoft/vscode-pull-request-github/blob/89fe5e32b2cdbe91885c44ba007711806e8ecc5d/src/notifications/notificationsView.ts#L86-L87), we can filter out the onse that have been read.
- Add a `removeNotification` to `NotificationManager`.

alexr00 (changes requested):
We shouldn't be maintaining the collection of notifications in two different places. I left a comment outlining some options!

src/notifications/notificationsProvider.ts
aiday-mar: Right I didn't think about that. I'll make the corresponding change

package.json
aiday-mar: I see better thanks. I added the actions into the group `issues_0@<number>` which placed them into the right-click context menu

src/notifications/notificationsManager.ts
```diff
@@ -77,6 +77,10 @@ export class NotificationsManager {
 		return this._notifications.get(key);
 	}
 
+	public removeNotification(key: string): void {
+		this._notifications.delete(key);
```
alexr00, line 81: I don't think it will have any negative impact now, but it would be good to fire the change event if the removed notification existed in the notification map.

package.json
```diff
@@ -2320,12 +2330,32 @@
 				{
 					"command": "notification.chatSummarizeNotification",
 					"group": "inline@1",
-					"when": "view == notifications:github && (viewItem == 'Issue' || viewItem == 'PullRequest')"
+					"when": "view == notifications:github && (viewItem == 'Issue' || viewItem == 'PullRequest') && config.githubPullRequests.experimental.notificationsView"
+				},
+				{
+					"command": "notification.chatSummarizeNotification",
+					"group": "issues_0@0",
```
alexr00, line 2337: nit: the name of this group could be `notifications_0` instead of `issues_0`

alexr00 (approved):
Thanks for the changes! I think we should only pass the notification manager into the notifications view. I overlooked this in my refactor yesterday. I will address during debt week!

package.json
aiday-mar: Ah right okay, I thought the word 'issues' is important in defining the right click context menu. Thanks for the feedback!

src/notifications/notificationsManager.ts
aiday-mar: Right thanks for the feedback!

PR #6318, opened 2024-10-20
Create deno.yml

Pandupra89:


alexr00:
We will not be supporting deno right now.

PR #6319, opened 2024-10-21
Notifications - Manager -> Provider cleanup

lszomoru:


src/notifications/notificationsManager.ts
```diff
@@ -73,38 +53,99 @@ export class NotificationsManager {
...
+			const notifications = Array.from(this._notifications.values());
 
-	public removeNotification(key: string): void {
-		this._notifications.delete(key);
-	}
+			return {
+				notifications: this._sortNotifications(notifications),
+				hasNextPage: this._hasNextPage
+			};
+		}
```
alexr00, line 64: Will `getNotifications` get called often with `compute === false`? The Javascript `Map` is sorted by insertion order, so we could re-make the map ever time the sort order changes so we don't have to sort here.

src/notifications/notificationsManager.ts
```diff
@@ -73,38 +53,99 @@ export class NotificationsManager {
...
-			const notification = this._notifications.get(update.key);
-			if (notification) {
-				notification.priority = update.priority;
-				notification.priorityReasoning = update.priorityReasoning;
-				updated.push(notification);
-			}
+		const notification = this._notifications.get(notificationIdentifier.notificationKey);
+		if (notification) {
+			this._onDidChangeNotifications.fire([notification]);
+			this._notifications.delete(notificationIdentifier.notificationKey);
```
alexr00, line 138: Might be better to do the `delete` before the `fire` to ensure that the state that is maintained in the `NotificationsManager` is fully up-to-date before event listeners are called.

src/notifications/notificationsView.ts
```diff
@@ -9,35 +9,35 @@ import { dispose } from '../common/utils';
...
+	private _pageCount: number = 1;
 	private _computeNotifications: boolean = false;
 
-	constructor(private readonly _notificationsProvider: NotificationsProvider, private readonly _notificationsManager: NotificationsManager) {
+	constructor(private readonly _notificationsManager: NotificationsManager) {
 		this._disposables.push(this._onDidChangeTreeData);
-		this._disposables.push(this._notificationsManager.onDidChangeNotifications(updates => {
-			this._onDidChangeTreeData.fire(updates);
+		this._disposables.push(this._notificationsManager.onDidChangeNotifications(() => {
+			this._onDidChangeTreeData.fire();
```
alexr00, line 26: nit: now that we have `this.refresh(false)` it would be nice to call that instead of firing the event emitter directly.

alexr00 (approved):
👏

Issue #6358, opened 2024-10-22
Use a Agent name that is different to Github


aeschli:
Testing #6321

the agent name it 'Github' which conflicts with the other 'Github' agent 
![Image](https://github.com/user-attachments/assets/1692493a-b9ac-4dea-b453-c42807b67bd8)

PR #6380, opened 2024-10-23
Changing name of the participant to `Github Pull Requests and Issues`

aiday-mar:
Changing name of the participant to `Github Pull Requests and Issues` from `GitHub`

fixes https://github.com/microsoft/vscode-pull-request-github/issues/6358

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
 			{
 				"id": "githubpr",
 				"name": "githubpr",
-				"fullName": "GitHub",
+				"fullName": "GitHub Pull Requests and Issues",
 				"description": "Chat participant for GitHub Pull Requests extension",
 				"when": "config.githubPullRequests.experimental.chat"
 			}
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
 			{
 				"id": "githubpr",
 				"name": "githubpr",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
 			{
 				"id": "githubpr",
 				"name": "githubpr",
-				"fullName": "GitHub",
+				"fullName": "GitHub Pull Requests and Issues",
```
alexr00, line 69: The extension's name is actually "GitHub Pull Requests". What do you think of aligning with the extension name?

alexr00: It used to be "GitHub Pull Requests", then 5 years ago we changed it to "GitHub Pull Requests and Issues" because we added some issues stuff. Then, 1 year ago we changed it back to "GitHub Pull Requests" because the issues support was minimal.

aiday-mar: Hi ah yes then I will change it to GitHub Pull Requests

aiday-mar:
I removed the 'Issues' in the participant name

alexr00 (approved):

package.json
alexr00: Seems good. Though I do wonder if adding "and Issues" back into the name is in the not-so-distant future.

Issue #6347, opened 2024-10-22
Issue summarization: Reference to other issue not linkified in response


chrmarti:
Testing #6321

The response mentions another issue from the same repo which could be linkified:
![Image](https://github.com/user-attachments/assets/895318ce-5cea-4906-97c1-50cbf8882e17)

Issue #6338, opened 2024-10-22
Issue/PR numbers should be links


TylerLeonhardt:
Testing #6329

![Image](https://github.com/user-attachments/assets/9d4c2d96-7e0e-485a-8bdb-8a4d221ede12)

Can the "Fixed in #...." be a link? Like I'd like to be able to click on the `#12341` and open the PR in the browser

PR #6396, opened 2024-10-23
Linkifying the output links in the summary for summarization tools

aiday-mar:
fixes https://github.com/microsoft/vscode-pull-request-github/issues/6347
fixes https://github.com/microsoft/vscode-pull-request-github/issues/6338

### DIFF
--- a/package.json
+++ b/package.json
@@ -3213,6 +3213,14 @@
 							"type": "string",
 							"description": "The body of the issue/PR"
 						},
+						"owner": {
+							"type": "string",
+							"description": "The owner of the repo in which the issue/PR is located"
+						},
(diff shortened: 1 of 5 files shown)


alexr00 (approved):
👏 cool! I was thinking of using regex like in https://github.com/microsoft/vscode-pull-request-github/blob/6aab2cbd39173cc442974cba30cb73c9c9de7eda/src/issues/issueLinkProvider.ts, but this is probably faster.

aiday-mar:
It could maybe be better to use regex, since it would be a deterministic operation. Since here I am using the LLM to linkify I expect there will sometimes be errors.

PR #6407, opened 2024-10-24
Make Copilot settings real settings

alexr00:


package.json
```diff
@@ -461,6 +461,21 @@
...
+				},
+				"githubPullRequests.experimental.notificationsView": {
+					"type": "boolean",
+					"markdownDescription": "%githubPullRequests.experimental.notificationsView.description%",
+					"default": false
+				},
+				"githubPullRequests.experimental.useQuickChat": {
+					"type": "boolean",
+					"markdownDescription": "%githubPullRequests.experimental.useQuickChat.description%",
+					"default": true
```
lszomoru, line 477: Let's flip this one to false since the Chat panel is moving to the auxiliary sidebar

package.nls.json
```diff
@@ -86,6 +86,9 @@
 	"githubPullRequests.createDefaultBaseBranch.repositoryDefault": "The default branch of the repository",
 	"githubPullRequests.createDefaultBaseBranch.createdFromBranch": "The branch that the current branch was created from, if known",
 	"githubPullRequests.createDefaultBaseBranch.auto": "When the current repository is a fork, this will work like \"repositoryDefault\". Otherwise, it will work like \"createdFromBranch\".",
+	"githubPullRequests.experimental.chat.description": "Enables the `@githubpr` Copilot chat provider in the chat view. `@githubpr` can help search for issues and pull requests, suggest fixes for issues, and summarize issues, pull requests, and notifications.",
```
lszomoru, line 89: Is it chat provider or chat participant?

package.json
aiday-mar: There are also some rendering issues when running the commands in the quick chat. The response is streamed but it disappears sometimes from the quick chat. Does not seem to happen in panel chat.

package.nls.json
alexr00: Thanks for catching!

PR #6421, opened 2024-10-25
Handling undefined values in summarize notifications tool

aiday-mar:
- Checking whether the parameters are defined before using them
- Renaming unreadComments to comments because the LLM behind seems to send the unreadComments as a comments array

### DIFF
--- a/src/lm/tools/fetchNotificationTool.ts
+++ b/src/lm/tools/fetchNotificationTool.ts
@@ -25,17 +25,17 @@ export interface FetchNotificationResult {
 	unread: boolean;
 	title: string;
 	body: string;
-	comments: {
+	comments?: {
 		author: string;
 		body: string;
 	}[];
 	owner: string;
 	repo: string;
-	itemNumber: string;
-	itemType: 'issue' | 'pr';
+	itemNumber?: string;
+	itemType?: 'issue' | 'pr';
 	fileChanges?: FileChange[];
-	threadId: number,
-	notificationKey: string
+	threadId?: number,
+	notificationKey?: string
 }
 
 export class FetchNotificationTool extends RepoToolBase<FetchNotificationToolParameters> {
--- a/src/lm/tools/fetchNotificationTool.ts
+++ b/src/lm/tools/fetchNotificationTool.ts
@@ -25,7 +25,7 @@ export interface FetchNotificationResult {
 	unread: boolean;
 	title: string;
 	body: string;
-	unreadComments: {
+	comments: {
 		author: string;
 		body: string;
 	}[];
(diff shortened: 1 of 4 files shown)


src/lm/tools/fetchNotificationTool.ts
```diff
@@ -25,7 +25,7 @@ export interface FetchNotificationResult {
 	unread: boolean;
 	title: string;
 	body: string;
-	unreadComments: {
+	comments: {
 		author: string;
 		body: string;
 	}[];
```
alexr00, line 31: Since this is also used as the input to `NotificationSummarizationTool`, where it isn't guaranteed that `itemType` and `itemNumber` are defined, it would be good to mark them as optional here for better type checking.

alexr00: And `notificationKey`

aiday-mar: It appears this issue can affect all of the fields of the return result of the notification fetch tool. Would it make sense to mark all of them as optional?

src/lm/tools/summarizeNotificationsTool.ts
```diff
@@ -44,32 +49,34 @@ Patch: ${fileChange.patch}
...
-			arguments: [{
-				threadId: options.parameters.threadId,
-				notificationKey: options.parameters.notificationKey
-			}]
-		};
+		let markAsReadCommand: vscode.Command | undefined;
+		const threadId = options.parameters.threadId;
+		const notificationKey = options.parameters.notificationKey;
+		if (threadId && notificationKey) {
+			markAsReadCommand = {
```
alexr00, line 74: If `markAsReadCommand` does not end up getting defined here, then below we shouldn't add it into the `LanguageModelToolResult` that we return.

src/lm/tools/fetchNotificationTool.ts
alexr00: For the candidate, I think it's fair to just mark the common ones. We should make a pass during debt week and do this for all of our tools

src/lm/tools/summarizeNotificationsTool.ts
aiday-mar: That's right! Thanks for picking up on that. I'll change that

aiday-mar:
Thanks for the review, I made the requested changes

alexr00 (approved):
Looks good!

src/lm/tools/summarizeNotificationsTool.ts
```diff
@@ -44,46 +49,47 @@ Patch: ${fileChange.patch}
...
-			new vscode.LanguageModelTextPart(JSON.stringify(markAsReadCommand)),
-			new vscode.LanguageModelTextPart(responseText)]);
+			content.push(new vscode.LanguageModelTextPart(responseText));
+			return new vscode.LanguageModelToolResult(content);
 		} else {
-			return new vscode.LanguageModelToolResult([new vscode.LanguageModelTextPart(TOOL_COMMAND_RESULT),
-			new vscode.LanguageModelTextPart(JSON.stringify(markAsReadCommand)),
-			new vscode.LanguageModelTextPart(notificationInfo)]);
+			content.push(new vscode.LanguageModelTextPart(notificationInfo));
+			return new vscode.LanguageModelToolResult(content);
```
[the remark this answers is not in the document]
aiday-mar: Right thanks!

alexr00 (approved):

PR #6481, opened 2024-11-18
Revert "GraphQL error: Fragment on Bot can't be spread inside RequestedReviewer (#6461)"

alexr00:
This reverts commit b298cdbd9fa39385ce3c90e674034943228b50ad.

### DIFF
--- a/src/github/queries.gql
+++ b/src/github/queries.gql
@@ -125,51 +125,6 @@ fragment PullRequestFragment on PullRequest {
 	}
 }
 
-fragment ReviewComment on PullRequestReviewComment {
-	id
-	databaseId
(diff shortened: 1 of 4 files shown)


alexr00:
My fix for the GHPR recovery 1 revealed an issue with how I'm merging GraphQL schemas, which is how we handle different schema versions with older versions of GitHub Enterprise. The recovery 1 fix was for a really old version of GHE, so for recovery 2 we'll just revert the recovery 1 change. I'll work on a fix for the merging of schemas for the next regular release. The old GHE version will simply not work for now.

lszomoru (approved):

PR #6517, opened 2024-12-09
Diffing `OUTDATED` comments with HEAD doesn't work in `github.dev`

### LINKED ISSUE #6500: Diffing `OUTDATED` comments with HEAD doesn't work in `github.dev`
Looks like this is specific to GitHub.dev

- Extension version: v0.100.3
- VSCode Version: 1.95.3
- OS: N/A (Web version github.dev)
- Repository Clone Configuration (single repository/fork of an upstream repository): single repo
- Github Product (Github.com/Github Enterprise version x.x.x): Github.com org

Steps to Reproduce:

1. Open Github.dev for a pull request in an org
2. Login
3. Open the comments view
4. Try and diff an outdated comment with HEAD

Shows this:
Image

Doubt this is an error here, I don't see any logs that are pertinent to this but just wanted some feedback as we would love to use this mechanism in github.dev.

alexr00:
Fixes #6500

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -1021,7 +1021,8 @@ export function registerCommands(
 
 	const localUriFromReviewUri = (reviewUri: vscode.Uri) => {
 		const { path, rootPath } = fromReviewUri(reviewUri.query);
-		return vscode.Uri.joinPath(vscode.Uri.file(rootPath), path);
+		const workspaceFolder = vscode.workspace.workspaceFolders![0];
+		return vscode.Uri.joinPath(vscode.Uri.file(rootPath), path).with({ scheme: workspaceFolder.uri.scheme, authority: workspaceFolder.uri.authority });
 	};
 
 	context.subscriptions.push(


src/commands.ts
```diff
@@ -1021,7 +1021,8 @@ export function registerCommands(
 
 	const localUriFromReviewUri = (reviewUri: vscode.Uri) => {
 		const { path, rootPath } = fromReviewUri(reviewUri.query);
-		return vscode.Uri.joinPath(vscode.Uri.file(rootPath), path);
+		const workspaceFolder = vscode.workspace.workspaceFolders![0];
```
Tyriar, line 1024: Could this ever be length=0 / could this throw/

Tyriar (approved):

src/commands.ts
alexr00: This will only be called when there's a PR open, and a PR can only be open if there's at least one workspace folder, so that should never happen.

Issue #6263, opened 2024-10-04
Warning dialog for converting changes to suggestions can be incorrect and vague


alexr00:
1. Check out a PR
2. Edit a line within a commenting range and a line not in a commenting range
3. From the SCM view, use the gift icon to convert your changes to PR suggestion comments
4. You'll see a dialog. It should indicate the number of converted suggestions that succeeded. 

// cc @benibenj

amunger:
I'm not able to find the referenced icon
![Image](https://github.com/user-attachments/assets/364cfbc5-f8c4-4ace-9ced-974ece6d7158)

TylerLeonhardt:
I'm pretty sure it's now this icon?
![Image](https://github.com/user-attachments/assets/696ff145-bf24-4876-a6a7-7dd50d59bdf2)

I tried that and I got this:
![Image](https://github.com/user-attachments/assets/4aa2d8a3-17de-4490-b56f-9d3e89dd8e05)

Even though my change was in the comment range as you can see.

alexr00:
@TylerLeonhardt I'm not able to reproduce that

![Image](https://github.com/user-attachments/assets/54b16e3e-86b5-4c70-997d-78cdab981ccf)

If you still see it, can you share the exact PR that you see this with?

TylerLeonhardt:
I still repro:
![Image](https://github.com/user-attachments/assets/d08c0e81-e6f1-41b3-9427-482b6c28ece4)

Steps:
* Open this PR: https://github.com/microsoft/vscode/pull/232058
* Open the inlineEditsView.ts
* Add `// A very important comment` in the range and then right outside of the range like so:
    ![Image](https://github.com/user-attachments/assets/9dd07926-7ff1-4add-9c35-3f4c9b4fe4e6)
* Do the create pull request suggestions

alexr00:
Thank you! I can repro with that example!

roblourens:
Getting this

![Image](https://github.com/user-attachments/assets/fdce4f96-8988-4abc-9c5c-1fda6655c831)

I believe this should be "1 of 2"

```
2024-12-06 15:00:41.254 [info] Review+1> Switch to Pull Request #232058 - start
2024-12-06 15:00:41.366 [info] PullRequestGitHelper> Branch origin/hediet/b/only-lark doesn't exist on local disk yet.
[...]

PR #6519, opened 2024-12-09
Fix diff count when converting to suggestions

alexr00:
Part of #6263


alexr00 (commented):
The problem was that the n in the "x of n" message was 0 because no diffs were ever pushed to the array that was checked for that message. The fix is to maintain a count of the number of diffs seen.

sandy081 (approved):

lszomoru (approved):

Issue #6422, opened 2024-10-25
Debt: mark all tool result return type fields as optional


aiday-mar:
With the current VS Code tool API, it is not guaranteed that the all the values from one return type are passed as parameters to the input of another tool. We should account for this with a compile time check by marking all the return types as optional.

PR #6521, opened 2024-12-09
Making tool parameters optional

aiday-mar:
fixes https://github.com/microsoft/vscode-pull-request-github/issues/6422

### DIFF
--- a/src/lm/tools/displayIssuesTool.ts
+++ b/src/lm/tools/displayIssuesTool.ts
@@ -102,14 +102,16 @@ export class DisplayIssuesTool extends ToolBase<DisplayIssuesParameters> {
 				case 'number':
 					return `[${issue[column]}](${issue.url})`;
 				case 'labels':
-					return issue[column].map((label) => makeLabel(label)).join(', ');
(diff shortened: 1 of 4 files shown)


src/lm/tools/displayIssuesTool.ts
```diff
@@ -102,14 +102,16 @@ export class DisplayIssuesTool extends ToolBase<DisplayIssuesParameters> {
 				case 'number':
 					return `[${issue[column]}](${issue.url})`;
 				case 'labels':
-					return issue[column].map((label) => makeLabel(label)).join(', ');
+					return issue[column]?.map((label) => label && label.name && label.color ? makeLabel({ name: label.name, color: label.color }) : '').join(', ');
```
alexr00, line 105: You should be able to leave out the `label` check like this.
```suggestion
					return issue[column]?.map((label) => label?.name && label.color ? makeLabel({ name: label.name, color: label.color }) : '').join(', ');
```

alexr00 (commented):
Thanks for taking this!

alexr00 (approved):

src/lm/tools/displayIssuesTool.ts
aiday-mar: PR https://github.com/microsoft/vscode-pull-request-github/pull/6523

Issue #6469, opened 2024-11-10
Default setting for the `python` language, and others which use the `#` and `@` symbols in their syntax and markup


NellyWhads:
The default settings explicitly disable all feature for the python language:  Python is excluded from issue and user autocompletion.
I understand the syntax collision between python decorators and code comments, however, [python is not the only language which uses these symbols in its syntax and markup](https://chatgpt.com/share/6730d5e5-a568-8011-9c87-e8c3dd29fa9a).

Suggested fixes:
1. Disable the default python language exceptions, at least for the `@` symbol to be consistent with all other languages
2. Allow users to configure the extension to work in comments only, or make this the default behavior
    - The extension will behave consistently across all languages
    - Each language will need a (hopefully default) configuration for parsing comment/markup symbols
    - This should automatically be handled by the language support extensions, but I am not a VSCode expert

- Extension version: v0.98.0
- VSCode Version: [1.93.1 (Cursor)](https://changelog.cursor.sh/#042---composer-history-lint-errors-vs-code-1931)
- OS: MacOS Sonoma 14.5 (23F79)
- Repository Clone Configuration (single repository/fork of an upstream repository): single repository, no submodules
- Github Product (Github.com/Github Enterprise version x.x.x): github.com

Steps to Reproduce:

1. Install extension and log in
2. Make python file
3. Attempt to use the extension's # and @ autocomplete features in the python file.

NellyWhads:
Are contributors encouraged to make PRs to this repo?

alexr00:
Yes, PRs are absolutely welcome! This week and next week in particular are a good time for community PRs too as we're we've started our [annual clean up](https://github.com/microsoft/vscode/issues/235576), which means I'll look at community PRs more quickly. Let me know if you want any code pointers.

NellyWhads:
@alexr00 PR is linked and pushed - how might I implement any additional tests, if necessary?

alexr00:
To verify:
- Install the latest pre-release version of GitHub Pull Requests
- Open your local clone of a GitHub repo in VS Code.
- Open or create a python file
- In a python comment, try to use the `@` user completion.

PR #6525, opened 2024-12-11
Remove the python language user mention exception

NellyWhads:
Motivation: #6469 

Implementation:
* Remove the default `python` language exception for user tags (`@`)

Tests:
* Attempt to use python decorators without interjection from the extension


alexr00 (approved):
Thanks for the PR! If users start complaining about this again I will add the default back in.

aiday-mar (approved):

PR #6531, opened 2024-12-12
Don't show hover cards for @-mentioned users in JSDocs in jsx and tsx files

Ronny-zzl:
The fix in #3122 does not work for jsx and tsx files


alexr00 (approved):
Thanks for the PR!

aiday-mar (approved):

PR #6539, opened 2024-12-13
Follow builtin VSCode behavior by respecting `git.showInlineOpenFileAction`

### LINKED ISSUE #6515: Follow builtin VSCode behavior by respecting `git.showInlineOpenFileAction`
The default "Source Control" view uses `git.showInlineOpenFileAction` to decide whether to display an extra button to quickly open a changed file (vs the diff), and also swaps the behavior of the button based on `git.openDiffOnClick` (does the opposite of clicking the filename itself).

By contrast, this extension always shows the inline "Open File" button, and its behavior is always to open the file, which is redundant when clicking the filename will open the file (instead of the diff).

Examples of the Source Control view's behavior:

Clicking the button here does the "open file" action:
```jsonc
{
    "git.openDiffOnClick": true,
    "git.showInlineOpenFileAction": true,
}
```
Image

----

Clicking the button here opens the diff instead, since clicking the filename opens the file:
```jsonc
{
    "git.openDiffOnClick": false,
    "git.showInlineOpenFileAction": true,
}
```
Image

----

The button is completely hidden in this case:
```jsonc
{
    "git.showInlineOpenFileAction": false,
}
```
Image

alexr00:
Fixes #6515

### DIFF
--- a/package.json
+++ b/package.json
@@ -2435,7 +2435,7 @@
 				{
 					"command": "review.openFile",
 					"group": "inline@0",
-					"when": "openDiffOnClick && view == prStatus:github && viewItem =~ /filechange(?!:DELETE)/"
+					"when": "openDiffOnClick && view == prStatus:github && viewItem =~ /filechange(?!:DELETE)/ && config.git.showInlineOpenFileAction"
 				},
 				{
 					"command": "pr.openDiffView",


Tyriar (approved):

package.json
```diff
@@ -2435,7 +2435,7 @@
 				{
 					"command": "review.openFile",
 					"group": "inline@0",
-					"when": "openDiffOnClick && view == prStatus:github && viewItem =~ /filechange(?!:DELETE)/"
+					"when": "openDiffOnClick && view == prStatus:github && viewItem =~ /filechange(?!:DELETE)/ && config.git.showInlineOpenFileAction"
```
ian-h-chamberlain, line 2438: Thanks for this change! Should this also be added to `pr.openDiffView` to hide that button when configured to `false`?

alexr00: Good reminder, thank you. https://github.com/microsoft/vscode-pull-request-github/pull/6552

PR #6540, opened 2024-12-13
Add command and shortcut to copy PR link

### LINKED ISSUE #6473: Add command and shortcut to copy PR link
In the PR page it's possible to copy the link to the PR (something like `https://github.com/org/repo/pull/14`). But as far as I can see there is no command to do that from the command palette and it's not possible to assign it a shortcut. This would be very useful when writing change logs. Ideally it would also be possible to copy the link in Markdown, for example `[#14](https://github.com/orgrepo/pull/14)`.

alexr00:
Fixes #6473

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
 		"tokenInformation",
 		"treeViewMarkdownMessage"
 	],
-	"version": "0.100.1",
+	"version": "0.102.0",
 	"publisher": "GitHub",
 	"engines": {
 		"vscode": "^1.96.0"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
 		"tokenInformation",
 		"treeViewMarkdownMessage"
 	],
-	"version": "0.102.0",
+	"version": "0.100.1",
```
Tyriar, line 35: Was it intentional to go backwards?

alexr00: Thanks for catching!

Tyriar (approved):

PR #6573, opened 2024-12-20
Comments don't show on non-checked out PR when closing and re-opening the file from the PRs veiw

### LINKED ISSUE #6571: Comments don't show on non-checked out PR when closing and re-opening the file from the PRs veiw
When working on stacked PRs it's common to want to address comments on a lower PR in the stack while having the branch from a higher PR checked out.

At the moment it seems that GithubPullRequests extension will not allow this *even when* a PR file or diff is opened in preview mode from the Pull Requests sidebar.

IMO the comments should still appear in these editor windows at the very least, and ideally in any file with comments.

alexr00:
Fixes #6571


catskul:
That was quick! 
Thanks! : )

Issue #6668, opened 2025-02-23
Copy GitHub Permalink doesn't work with custom SSH


christianvuerings:
- Extension version: `0.104.0`
- VSCode Version: `1.97.2`
- OS: `Mac OS 15.2 (24C101)`
- Repository Clone Configuration (single repository/fork of an upstream repository): SSH `org-4@git.musta.ch:organization/repo.git`
- `github-enterprise.uri` set to `https://git.musta.ch`
- Github Product (Github.com/Github Enterprise version x.x.x): `GitHub Enterprise Server 3.13.7`

Steps to Reproduce:

1. Click `Copy As` => `Copy GitHub Permalink`

**Issue**:

Copies
```
https://github.com/organization/repo/blob/b120817b885436984b948dd4ea86d399e35846e9/package.json#L19
```

**Expected**:
  
Copies
```
https://git.musta.ch/organization/repo/blob/b120817b885436984b948dd4ea86d399e35846e9/package.json#L19
```

PR #6669, opened 2025-02-23
Fix Copy GitHub Permalink with custom SSH

christianvuerings:
Fixes https://github.com/microsoft/vscode-pull-request-github/issues/6668

### DIFF
--- a/src/issues/util.ts
+++ b/src/issues/util.ts
@@ -584,7 +584,7 @@ export function getUpstreamOrigin(upstream: Remote, resultHost: string = 'github
 			fetchUrl = fetchUrl.substr('ssh://'.length);
 		}
 		// upstream's origin by ssh
-		if (fetchUrl.startsWith('git@') && !fetchUrl.startsWith('git@github.com')) {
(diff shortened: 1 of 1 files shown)


src/issues/util.ts
```diff
@@ -584,7 +584,7 @@ export function getUpstreamOrigin(upstream: Remote, resultHost: string = 'github
 			fetchUrl = fetchUrl.substr('ssh://'.length);
 		}
 		// upstream's origin by ssh
-		if (fetchUrl.startsWith('git@') && !fetchUrl.startsWith('git@github.com')) {
+		if ((fetchUrl.startsWith('git@') || fetchUrl.includes('@git')) && !fetchUrl.startsWith('git@github.com')) {
```
christianvuerings, line 587: Issue was that we would only match `git@git.musta.ch:organization/repo.git`, not `org-4@git.musta.ch:organization/repo.git`

Open to other suggestions as well - e.g. matching `.endsWith('.git')` instead

christianvuerings:

alexr00 (approved):
Thanks for the fix!

joaomoreno (approved):

Issue #6696, opened 2025-03-06
GraphQL error invalid email address when merging PRs


szschaler:
This might be a regression of #5842.

- Extension version: 0.106.0
- VSCode Version: Version: 1.98.0 (user setup)
Commit: 6609ac3d66f4eade5cf376d1cb76f13985724bcb
Date: 2025-03-04T21:06:18.612Z
Electron: 34.2.0
ElectronBuildId: 11044223
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
- OS: Windows_NT x64 10.0.19045
- Repository Clone Configuration (single repository/fork of an upstream repository): single repository
- Github Product (Github.com/Github Enterprise version x.x.x): github.com

Steps to Reproduce:

1. Create a branch and make some commits (all via VS Code)
2. Publish Branch and elect to create a PR
3. Wait for tests etc to be completed
4. Open the PR in VS Code and click on "Create Merge Commit" -- note that a valid email address that is registered as primary email address against GH account is shown just above that button.
5. Get the error message "GraphQL error invalid email address" and no merged PR
6. Change the selected email address to an alternative address also registered with GH and try again -- see it work

minnogit:
Same for me also on pre-release version

tblackstone0816:
Same for me also

dalpii:
Same for me, no change on repo/organization settings:
<img width="861" alt="Image" src="https://github.com/user-attachments/assets/211a8340-b25e-4c04-9166-9e0d64f375f6" />


Relevant debug log:
`2025-03-11 15:05:32.558 [debug] [FolderRepositoryManager+0] Merging PR: 3 method: merge for user: "" - enter`

scelikcapa:
same for me and I use the workaround solution here.  [workaround post](https://github.com/microsoft/vscode-pull-request-github/issues/5842#issuecomment-2038163493 )

use the button in the GitHub Pull Request Bar on the left side of VSCode

alexr00:
Folks in this thread, do you still see this issue with the latest VS Code Insiders + the latest pre-release version of the extension?

TylerLeonhardt:
@minnogit @tblackstone0816 @dalpii @scelikcapa Can you confirm this is fixed in [VS Code Insiders](http://code.visualstudio.com/insiders/) & the pre-release version of GitHub Pull Request?

TylerLeonhardt:
@alexr00 

[...]

Issue #6593, opened 2025-01-10
Email address for merge commit is not remembered (and cannot be set)


frossie:
Following from #5326 which added the ability to choose the e-mail address associated with a merge commit (thank you!)  

Problem: if you want a different value (for the e-mail to be associated with your merge commit,) than the one the extension defaults to,  you have to change it every time. 

- Ideally, the e-mail selected would default to the one configured in the local git repository

As is, if you have made commits from eg. your work computer and you forget to switch out the email associated with the merge commit (that seems to default to the on your personal Github profile) you end up with the merge commit being with a different e-mail than the rest of your commits. The best option is for the extension to default to the email that matches the git config user.email (provided it matches one of the email addresses returned by the Github API of course)

- Alternatively, the extension could default to the last value chosen in the extension

This is the behavior of the Github UI and probably deals with the vast majority of annoyances. 

- Or, it could be settable as an extension property

This is less than ideal because of people who do open source and employer commits under separate e-mails from the same environment, but it would be an improvement on the current situation.

PR #6797, opened 2025-04-17
Fix merge email confirmation when git config fails

kabel:
Centralizes the preferred email logic for a `PullRequestModel` in the `FolderRepositoryManager`.

`PullRequestGitHelper.getEmail` can return an empty string when it is unable to find `user.email` in the local or global config. This can occur when the `user.email` config is set in a different scope (system, worktree) or its in an include file of the requested (local, global) scope (`git config` is not run with the `--includes` argument). Ideally there would be an API method to get config from _any_ scope, sadly this is not exposed on the `Repository` interface from the core git extension. If we are unable to get the email from git config, continue to default to the primary GitHub email and allow users to manually select the correct address to use.

Use this same default email logic in the Activity Bar View so that the merge buttons behave in a similar manner with regard to what email is used for merge commits.

See #6593, #6696

### DIFF
--- a/src/extensionState.ts
+++ b/src/extensionState.ts
@@ -12,6 +12,7 @@ export const NEVER_SHOW_PULL_NOTIFICATION = 'github.pullRequest.pullNotification
 // Not synced keys
 export const REPO_KEYS = 'github.pullRequest.repos';
 export const PREVIOUS_CREATE_METHOD = 'github.pullRequest.previousCreateMethod';
+export const LAST_USED_EMAIL = 'github.pullRequest.lastUsedEmail';
 
 export interface RepoState {
 	mentionableUsers?: IAccount[];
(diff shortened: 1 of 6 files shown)


kabel:
Additional log context on why the email is blank:

```
[info] > git config --get --local user.email [4ms]
[warning] [Git][config] git config failed: Failed to execute git
[info] > git config --get --global user.email [5ms]
[warning] [Git][config] git config failed: Failed to execute git
```

So email got evaluated to `'' && undefined ?? 'user@example.com'`  which is `''`

alexr00 (commented):
Thanks for the PR

alexr00 (approved):

DonJayamanne (approved):

DonJayamanne (approved):

PR #6900, opened 2025-05-05
Filter private emails out

alexr00:


aeschli (approved):

src/github/githubRepository.ts
```diff
@@ -897,7 +897,7 @@ export class GitHubRepository extends Disposable {
 			const { data } = await octokit.call(octokit.api.users.listEmailsForAuthenticatedUser, {});
```
glektarssza, line 897: Wouldn't it be better to call [`listPublicEmailsForAuthenticatedUser`](https://docs.github.com/en/rest/users/emails?apiVersion=2022-11-28#list-public-email-addresses-for-the-authenticated-user) instead of filtering down below? Given it's causing bugs like #6916?

alexr00: When I tried `listPublicEmailsForAuthenticatedUser` I found that it didn't give the no-reply email, which is necessary.

glektarssza: Fair point!

Issue #6916, opened 2025-05-07
[bug] Secondary emails are not showing up as options when merging PRs


glektarssza:
### System Information ###

* Extension version: 0.109.2025050710
* VSCode Version: 1.100.0-insider
* OS: Windows 11 Pro 24H2 build 26100.1742
* Repository Clone Configuration (single repository/fork of an upstream repository): clone of single repository over SSH
* Github Product (Github.com/Github Enterprise version x.x.x): GitHub.com

### Steps to Reproduce ###

1. Open a PR to be merged.
2. Click the merge button.
3. Observe available email options.

### Expected Results ###

* All emails listed in GitHub profile as available for use as committer emails are listed.

### Actual Results ###

* Only main account email and GitHub's masking email are listed.

### Images ###

Inside VSCode (using [this PR](https://github.com/glektarssza/minecraft-player-handling-customizer-gtnh/pull/6) as an example):

![Image](https://github.com/user-attachments/assets/4fc7b62a-df7c-4efe-9ca4-31459d7d3543)

GitHub web interface (same PR as above):

![Image](https://github.com/user-attachments/assets/c536fe03-ec35-47c1-b40c-7b55e26156a5)

My email account settings showing I have three emails attached to my account set to be usable as committer emails:

![Image](https://github.com/user-attachments/assets/a2d23c76-9933-4ac9-9347-0dd65b9ed88c)

glektarssza:
In regards to the changes made in #6900, here's the output of calling `listEmailsForAuthenticatedUser`:
```json
[
  {
    "email": "<private>@<well-known-provider>",
    "primary": true,
    "verified": true,
    "visibility": "public"
  },
  {
    "email": "94402525+glektarssza@users.noreply.github.com",
    "primary": false,
    "verified": true,
    "visibility": null
  },
  {
    "email": "glek@glektarssza.com",
    "primary": false,
    "verified": true,
    "visibility": null
  },
  {
    "email": "<private>@<personal-domain>",
    "primary": false,
    "verified": true,
    "visibility": null
  }
]
```

I don't know why `visibility` is coming back as `null`...

kabel:
Same. All my non-primary email addresses have `"visibility": null` even though the GitHub web interface shows them as "Visible in emails". If I had to venture a guess, that field is only applicable to how the web UI handles certain aspects of email visibility for the primary address.

alexr00:
[...]

PR #6921, opened 2025-05-07
Allow verified GitHub emails when none are private

kabel:
See  #6916.

If users haven't opted-in to use private email, still allow all secondary email addresses to be used. The REST response shows these email records with `"visibility": null`.

### DIFF
--- a/src/github/githubRepository.ts
+++ b/src/github/githubRepository.ts
@@ -897,7 +897,10 @@ export class GitHubRepository extends Disposable {
 			const { data } = await octokit.call(octokit.api.users.listEmailsForAuthenticatedUser, {});
 			Logger.debug(`Fetch authenticated user emails - done`, this.id);
 			// sort the primary email to the first index
-			return data.filter(email => email.visibility === 'public' || email.email.toLowerCase().endsWith('@users.noreply.github.com')).sort((a, b) => +b.primary - +a.primary).map(email => email.email);
+			const hasPrivate = data.some(email => email.visibility === 'private');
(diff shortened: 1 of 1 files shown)


kabel (commented):
The `noreply` address suffix may not work as expected on GitHub Enterprise Server. I no longer have access to a enterprise server to verify what the `noreply` address is there (if it's customized for the on-premise instance).

src/github/githubRepository.ts
```diff
@@ -897,7 +897,10 @@ export class GitHubRepository extends Disposable {
 			const { data } = await octokit.call(octokit.api.users.listEmailsForAuthenticatedUser, {});
 			Logger.debug(`Fetch authenticated user emails - done`, this.id);
 			// sort the primary email to the first index
-			return data.filter(email => email.visibility === 'public' || email.email.toLowerCase().endsWith('@users.noreply.github.com')).sort((a, b) => +b.primary - +a.primary).map(email => email.email);
+			const hasPrivate = data.some(email => email.visibility === 'private');
+			return data.filter(email => hasPrivate ? email.email.endsWith('@users.noreply.github.com') : email.verified)
```
alexr00, line 901: The GitHub API is unclear, but I think this fix makes sense.

alexr00 (approved):
@kabel, thanks for this PR. I'm likely going to have to release with bug https://github.com/microsoft/vscode-pull-request-github/issues/6916, today, but I'll make a recovery release with this fix tomorrow.

Issue #6792, opened 2025-04-13
Fetching issues fails in Copilot Agent mode


jooyoungseo:
Type: <b>Bug</b>

I am having an issue with fetching GitHub issue in Copilot Chat.

The following is my current setup:

1. Private repo
2. Agent mode (ask mode does not have any issue)

## Chat Question

> `@githubpr Summarize issue xability/maidr-ts#11`

## Copilot Chat Response

> I'll help summarize the issue from the GitHub repository. Let me search for details about issue #11 in the maidr-ts repository.
> 
> Searched text for issue #11, no results
> I apologize, but I couldn't find direct references to issue #11 in the codebase. Let me try a different approach to search for the issue content by looking
> for more general references to issues or pull requests.

there is no issue with GitHub authentication because I can use all the GitHub Pull Requests extension features, including making PR, listing issues/PR, etc in VS Code.

Furthermore, like what I said, I can fetch the issue with the same chat query in Ask mode. Only Agent mode has this issue.

I double-checked whether my GitHub cli was authenticated in terminal. Yes, it is:

```bash
gh issue view 11 --repo xability/maidr-ts
```

```text
docs: add README xability/maidr-ts#11                                                                                   
Open • jooyoungseo opened about 6 months ago • 0 comments                                                               
Assignees: jooyoungseo                                                                                                  
Labels: backlog                                                                                                         
                                                                                                                        
                                                                                                                        
  Need to add README later once this repo is ready for public use.                                                      
                                                                                                                        
                                                                                                                        
                                                                                                                        
View this issue on GitHub: https://github.com/xability/maidr-ts/issues/11                                                
```



Extension version: 0.109.2025041114
VS Code version: Code - Insiders 1.100.0-insider (043f699013796cbeae7c8fe9abac9b9eb26b3c51, 2025-04-11T05:23:04.452Z)
OS version: Windows_NT x64 10.0.26100
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|11th Gen Intel(R) Core(TM) i5-1145G7 @ 2.60GHz (8 x 2611)|
[...]

PR #6956, opened 2025-05-14
Enable all LLM tools in prompts (agent mode)

bendrucker:
Enables [`canBeReferencedInPrompt`](https://code.visualstudio.com/api/extension-guides/tools) for all LLM tools.

> Set to `true` if the tool can be used in [agent mode](https://code.visualstudio.com/docs/copilot/chat/chat-agent-mode.md) or referenced in chat.

Previously, these tools could only be used via the `@githubpr` chat participant. Only the `activePullRequest` tool was available directly. Now, they can all be called with `#` syntax or invoked by agent mode.

Perhaps this was disabled intentionally and there is additional polish to do. But this works and https://github.com/microsoft/vscode-pull-request-github/issues/6792#issuecomment-2862104548 suggested that there was some confusion on whether this is live today (it's not).

## Before

Screenshot 2025-05-14 at 11 48 59

Screenshot 2025-05-14 at 11 53 36

Using the web fetch tool works for public repositories, albeit with a lot of HTML noise as compared to feeding the model a JSON representation of the issue.

Agent mode is unable to fetch details about any issues in private repositories.

## After

Screenshot 2025-05-14 at 11 44 13
Screenshot 2025-05-14 at 11 45 07

## Related

* https://github.com/microsoft/vscode-pull-request-github/issues/6792

### DIFF
--- a/package.json
+++ b/package.json
@@ -3724,7 +3724,7 @@
 				"displayName": "%languageModelTools.github-pull-request_notification_fetch.displayName%",
 				"modelDescription": "Get a GitHub notification's details as a JSON object.",
 				"icon": "$(info)",
-				"canBeReferencedInPrompt": true,
+				"canBeReferencedInPrompt": false,
 				"inputSchema": {
 					"type": "object",
 					"properties": {
@@ -3750,7 +3750,7 @@
 				"displayName": "%languageModelTools.github-pull-request_issue_summarize.displayName%",
 				"modelDescription": "Summarizes a GitHub issue or pull request. A summary is a great way to describe an issue or pull request.",
 				"icon": "$(info)",
-				"canBeReferencedInPrompt": true,
+				"canBeReferencedInPrompt": false,
 				"inputSchema": {
 					"type": "object",
 					"properties": {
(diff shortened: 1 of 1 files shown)


bendrucker:

alexr00:
This was disabled intentionally. `canBeReferencedInPrompt` was just for referencing a tool in a prompt with `#` or using the attach context button. I'm confirming that `canBeReferencedInPrompt` also impacts tool use in agent mode.

bendrucker:
Some other references:

https://code.visualstudio.com/updates/v1_99#_vs-code-extension-tools-in-agent-mode

> Any tool contributed to this API which sets toolReferenceName and canBeReferencedInPrompt in its configuration is automatically available in agent mode.

https://github.com/microsoft/vscode/issues/245129
https://github.com/microsoft/vscode/issues/231927

While that language isn't 100% definitive, my testing has shown that to be required for an agent to call the tool.

The screenshot I included with the checklists is triggered by the `Select Tools` button:

Screenshot 2025-05-15 at 13 18 47

And it seems reasonable to assume that list is the authoritative set of tools available to the agent.

PR #6981, opened 2025-05-22
ux - adopt better casing for buttons

bpasero:
@alexr00 I might have missed some, but generally using uppercase for buttons (unless common stop words), to align with VS Code core.


alexr00 (commented):
Thanks for the change! However, I thought that we only did uppercase for each word when the action is a command.

bpasero:
@alexr00 what got me into this was seeing a "Ready for review" button in the GH PR view on a PR I was working on. My understanding is that buttons use capital casing, except for very common short words, such as `for`. 

@ntrogh defined these rules a while ago.

ntrogh:
Yes! And they are here: https://github.com/microsoft/vscode/wiki/Coding-Guidelines#ui-labels

I have a backlog item to make these content guidelines more extensive.

PR #7002, opened 2025-06-02
fix: Allow Github.com auth when `github-enterprise.uri` is set

dyhagho:
## Problem
Since version `0.110.0`, setting the following configuration:

```json
"github-enterprise.uri": "https://someghe.com"
```

causes the GitHub.com authentication flow to be skipped entirely. This results in:
- No prompt to authenticate with `github.com`
- Pull Request extension for `Github.com` repos breaking silently

For many developers (including myself), contributing to both `github.com` and a company-hosted GitHub Enterprise is common. 

Even though the config can be set **per workspace**, it can be tedious to maintain across multiple repositories. 
And, in practice, setting `github-enterprise.uri` in user settings makes more sense. But this intentionally disables `github.com` functionality with the current logic.

## What does change?
This PR updates `doCreate` so that GitHub.com is initialized even though the enterprise URI is configured.
This mirrors pre-`0.110.0` behavior.

### Before

before

### After

after

### DIFF
--- a/src/github/credentials.ts
+++ b/src/github/credentials.ts
@@ -221,7 +221,6 @@ export class CredentialStore extends Disposable {
 	}
 
 	private async doCreate(options: vscode.AuthenticationGetSessionOptions, additionalScopes: boolean = false): Promise<AuthResult> {
-		const github = await this.initialize(AuthProvider.github, options, additionalScopes ? SCOPES_WITH_ADDITIONAL : undefined, additionalScopes);
 		let enterprise: AuthResult | undefined;
 		const initializeEnterprise = async () => {
 			enterprise = await this.initialize(AuthProvider.githubEnterprise, options, additionalScopes ? SCOPES_WITH_ADDITIONAL : undefined, additionalScopes);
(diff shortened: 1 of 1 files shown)


dyhagho:
@alexr00 could I get a review? :)

src/github/credentials.ts
```diff
@@ -221,6 +221,7 @@ export class CredentialStore extends Disposable {
 	}
 
 	private async doCreate(options: vscode.AuthenticationGetSessionOptions, additionalScopes: boolean = false): Promise<AuthResult> {
+		const github = await this.initialize(AuthProvider.github, options, additionalScopes ? SCOPES_WITH_ADDITIONAL : undefined, additionalScopes);
```
alexr00: This will cause a badge to be shown on the Accounts menu, prompting for github.com auth, even when enterprise auth is successful. If enterprise auth has succeeded, the `options` should include `silent`.

alexr00 (changes requested):
Thanks for the PR! I understand the motivation, but I think we need one change.

src/github/credentials.ts
dyhagho: Thanks for the feedback!

What do you think of this approach?

```ts
const githubOptions = { ...options };
if (enterprise && !enterprise.canceled && this.isAuthenticated(AuthProvider.githubEnterprise)) {
  githubOptions.silent = true;
}
const github = await this.initialize(AuthProvider.github, githubOptions, additionalScopes ? SCOPES_WITH_ADDITIONAL : undefined, additionalScopes);
```

alexr00: I don't know that the `isAuthenticated` check is needed, but otherwise it looks good.

PR #7044, opened 2025-06-19
Active PR tool to include coding agent logs

osortega:
Mainly introducing two changes:
1. The active PR tool now fetches and attaches the coding agent session logs so that the agent can understand the process padawan did.
2. When checking out a PR created by Copilot, we automatically open the chat and attach the active PR tool in case the user wants to keep iterating over it.


src/lm/tools/activePullRequestTool.ts
```diff
@@ -21,18 +29,171 @@ export class ActivePullRequestTool implements vscode.LanguageModelTool<FetchIssu
...
+		cancellationToken: vscode.CancellationToken
+	) {
+		const runs = await github.octokit.api.actions.listWorkflowRunsForRepo(
+			{
+				owner: pullRequest.githubRepository.remote.owner,
+				repo: pullRequest.githubRepository.remote.repositoryName,
+				event: 'dynamic'
+			}
+		);
+		const padawanRuns: any[] = runs.data.workflow_runs
```
alexr00: Can we add a new type for this to avoid the `any`?

alexr00: You should be able to put this in src/github/common.ts and use it here

```
	export type ListWorkflowRunsForRepo = Endpoints['GET /repos/{owner}/{repo}/actions/runs']['response']['data']['workflow_runs'];
```

src/lm/tools/activePullRequestTool.ts
```diff
@@ -21,18 +29,171 @@ export class ActivePullRequestTool implements vscode.LanguageModelTool<FetchIssu
...
+			const sessionsResponse = await fetch(`https://api.githubcopilot.com/agents/sessions/resource/pull/${pullRequest.id}`, {
+				headers: {
+					Authorization: `Bearer ${token}`,
+					Accept: 'application/json',
+				},
+			});
+			if (!sessionsResponse.ok) {
+				throw new Error(`Failed to fetch sessions: ${sessionsResponse.statusText}`);
+			}
+			const sessions = await sessionsResponse.json();
```
alexr00: It would be great to have a type for this too!

src/lm/tools/activePullRequestTool.ts
```diff
@@ -21,18 +29,171 @@ export class ActivePullRequestTool implements vscode.LanguageModelTool<FetchIssu
...
+			const completedSessions = sessions.sessions.filter((s: any) => s.state === 'completed');
+			const mostRecentSession = completedSessions.reduce((latest: any, session: any) => {
+				return !latest || new Date(session.last_updated_at) > new Date(latest.last_updated_at)
+					? session
+					: latest;
```
alexr00: There's very similar "get most recent session" logic above too. Might be good to have a utility for this.

src/lm/tools/activePullRequestTool.ts
osortega: Great idea, I added the type

src/lm/tools/activePullRequestTool.ts
osortega: Added the return object type

src/lm/tools/activePullRequestTool.ts
osortega: Moved it to a utility

alexr00 (approved):

Issue #7053, opened 2025-06-23
Render past tense when agent tool completes


rebornix:
<img width="442" alt="Image" src="https://github.com/user-attachments/assets/0fc87a49-8d6c-40b5-81e4-6feabf7910a5" />

After it finishes, we might want to render a [past tense message](https://github.com/microsoft/vscode/blob/b5ad5e36b2fa5d8753889a19bec3a9626ee31795/src/vscode-dts/vscode.proposed.chatParticipantPrivate.d.ts#L193-L196) which can include markdown link to the PR/Session.

PR #7063, opened 2025-06-23
tool invocation polish

joshspicer:
closes https://github.com/microsoft/vscode-pull-request-github/issues/7053

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
 	"enabledApiProposals": [
 		"activeComment",
 		"chatParticipantAdditions",
+		"chatParticipantPrivate",
 		"codiconDecoration",
 		"codeActionRanges",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -35,7 +36,7 @@
 	"version": "0.112.0",
 	"publisher": "GitHub",
 	"engines": {
-		"vscode": "^1.100.0"
+		"vscode": "^1.102.0"
```
joshspicer, line 39: @alexr00  FYI bumped engine version due to new api proposal (+ @osortega already added `chatParticipantAdditions`).  This is necessary, right?

alexr00: Yep!

alexr00 (approved):

PR #7066, opened 2025-06-24
API for repository description

osortega:
Adding an API for fetching repository description

### DIFF
--- a/src/api/api.d.ts
+++ b/src/api/api.d.ts
@@ -265,6 +265,8 @@ export interface RepositoryDescription {
 	pullRequest?: {
 		title: string;
 		url: string;
+		number: number;
+		id: number;
 	};
 }
 
--- a/src/api/api.d.ts
+++ b/src/api/api.d.ts
@@ -294,8 +294,8 @@ export interface API {
 	registerReviewerCommentsProvider(title: string, provider: ReviewerCommentsProvider): Disposable;
 
 	/**
-	 * Get the repository description for a given URI.
-	 * This includes the owner, repository name, default branch, current branch (if applicable),
+	 * Get the repository description for a given URI, where the URI is a subpath of one of the workspace folders.
+	 * This includes the owner, repository name, default branch,
 	 * and pull request information (if applicable).
 	 *
 	 * @returns A promise that resolves to a `RepositoryDescription` object or `undefined` if no repository is found.
--- a/src/api/api.d.ts
+++ b/src/api/api.d.ts
@@ -261,7 +261,6 @@ export interface RepositoryDescription {
 	owner: string;
 	repositoryName: string;
 	defaultBranch: string;
-	currentBranch?: string;
 	pullRequest?: {
 		title: string;
 		url: string;
--- a/src/api/api.d.ts
+++ b/src/api/api.d.ts
@@ -295,11 +295,11 @@ export interface API {
 	registerReviewerCommentsProvider(title: string, provider: ReviewerCommentsProvider): Disposable;
 
 	/**
-	 * Get the repository description for the current workspace.
+	 * Get the repository description for a given URI.
 	 * This includes the owner, repository name, default branch, current branch (if applicable),
 	 * and pull request information (if applicable).
 	 *
 	 * @returns A promise that resolves to a `RepositoryDescription` object or `undefined` if no repository is found.
 	 */
-	getRepositoryDescription(): Promise<RepositoryDescription | undefined>;
+	getRepositoryDescription(uri: vscode.Uri): Promise<RepositoryDescription | undefined>;
 }
--- a/src/api/api.d.ts
+++ b/src/api/api.d.ts
@@ -257,6 +257,17 @@ export interface ReviewerCommentsProvider {
 	provideReviewerComments(context: { repositoryRoot: string, commitMessages: string[], patches: { patch: string, fileUri: string, previousFileUri?: string }[] }, token: CancellationToken): Promise<ReviewerComments>;
 }
 
+export interface RepositoryDescription {
+	owner: string;
+	repositoryName: string;
+	defaultBranch: string;
+	currentBranch?: string;
+	pullRequest?: {
+		title: string;
+		url: string;
+	};
+}
+
 export interface API {
 	/**
 	 * Register a [git provider](#IGit)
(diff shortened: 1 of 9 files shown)


rebornix (approved):

DonJayamanne (approved):

src/api/api.d.ts
```diff
@@ -257,6 +257,17 @@ export interface ReviewerCommentsProvider {
...
 }
 
+export interface RepositoryDescription {
+	owner: string;
+	repositoryName: string;
+	defaultBranch: string;
+	currentBranch?: string;
+	pullRequest?: {
+		title: string;
+		url: string;
```
alexr00, line 266: Can we include `number` here too? That's usually a useful thing to have.

src/extension.ts
```diff
@@ -405,19 +402,20 @@ async function deferredActivate(context: vscode.ExtensionContext, apiImpl: GitAp
 	const showBadge = (vscode.env.appHost === 'desktop');
 	await credentialStore.create(showBadge ? undefined : { silent: true });
 
+	Logger.debug('Creating tree view.', 'Activation');
```
alexr00: This log line is now not right before the tree view is created.

alexr00: Actually, are there any functional changes in this file?

src/api/api1.ts
```diff
@@ -220,4 +227,39 @@ export class GitApiImpl extends Disposable implements API, IGit {
 	getReviewerCommentsProvider(): { title: string, provider: ReviewerCommentsProvider } | undefined {
 		return this._reviewerCommentsProviders.size > 0 ? this._reviewerCommentsProviders.values().next().value : undefined;
 	}
+
+	async getRepositoryDescription() {
```
alexr00: Instead of just returning info about the first repo, I would suggest passing in a URI and returning info about the repo that the URI belongs to.

src/extension.ts
osortega: No functional changes, this is just because we need to create the `reposManager` before the `apiImp` now because we need to pass it as an argument.
I deleted the log line though since I duplicated it by mistake

src/api/api1.ts
osortega: In this case the API will be used in chat to be able to gather and include the context of the repository in which the user is currently at, since there is not a specific file to fetch the PR from.
Is there a better way to determine the current repository you have open than this? I ended up just comparing the `fsPath`

src/api/api.d.ts
osortega: Good idea, I added number and id in case we needed them

DonJayamanne (dismissed):

src/api/api1.ts
osortega: Changed it to use the URI thanks for the recommendation

src/api/api.d.ts
```diff
@@ -257,6 +257,19 @@ export interface ReviewerCommentsProvider {
 	provideReviewerComments(context: { repositoryRoot: string, commitMessages: string[], patches: { patch: string, fileUri: string, previousFileUri?: string }[] }, token: CancellationToken): Promise<ReviewerComments>;
 }
 
+export interface RepositoryDescription {
+	owner: string;
+	repositoryName: string;
+	defaultBranch: string;
+	currentBranch?: string;
```
alexr00: The current branch can be gotten from the git extension, so I don't think we should expose it.

src/api/api1.ts
```diff
@@ -220,4 +227,33 @@ export class GitApiImpl extends Disposable implements API, IGit {
 	getReviewerCommentsProvider(): { title: string, provider: ReviewerCommentsProvider } | undefined {
 		return this._reviewerCommentsProviders.size > 0 ? this._reviewerCommentsProviders.values().next().value : undefined;
 	}
+
+	async getRepositoryDescription(uri: vscode.Uri) {
+		const repo = this.getGitProvider(uri)?.repositories.find(r => r.rootUri.fsPath === uri.fsPath);
```
alexr00: Instead of checking for equality with the rootUri, we should find the git repo where the uri is a subpath of the git repo's uri. This could be multiple repos.

alexr00: We have a method that does this already for folder managers, I think you can just use that.

https://github.com/microsoft/vscode-pull-request-github/blob/373102408f8e50bca6c7fb888a432d9478b348c4/src/github/repositoriesManager.ts#L131-L132

src/api/api1.ts
```diff
@@ -220,4 +227,33 @@ export class GitApiImpl extends Disposable implements API, IGit {
...
+		if (!repo) {
+			Logger.appendLine(`No repository found for uri: ${uri.fsPath}`, GitApiImpl.ID);
+			return undefined;
+		}
+		const folderManagerForRepo = this.repositoriesManager.folderManagers.find((manager) => manager.gitHubRepositories.some(r => r.rootUri.fsPath === repo?.rootUri.fsPath));
+		const folderManagerForPR = this.repositoriesManager.folderManagers.find((manager) => manager.activePullRequest);
+
+		if (folderManagerForRepo && folderManagerForRepo.gitHubRepositories.length > 0) {
+			const repositoryMetadata = await folderManagerForRepo.gitHubRepositories[0].getMetadata();
+			const pullRequest = folderManagerForPR?.activePullRequest ?? PullRequestOverviewPanel.currentPanel?.getCurrentItem();
```
alexr00: Let's keep this just to the currently checked out PR, not the PR in the webview. When users have asked for this feature in the past they really just want to know about the currently checked out PR.

alexr00 (commented):
Just to emphasize that the Copilot extension should not know about the PR extension: I want to delete the PR title and description stuff from the Copilot extension and move it into the PR extension and will do so in a future debt week.

src/api/api.d.ts
osortega: Removed it

src/api/api1.ts
osortega: Ahhh thanks for this pointer! Makes it so much easier

src/api/api1.ts
osortega: Good idea, removed the logic that gets it through the panel

alexr00 (dismissed):
I'm about to make a pre-release, and I know you've been waiting for this to get in. I pushed some very minor chnages in the interest of getting this into the pre-release sooner. Thank you for addressing my comments!

aeschli (approved):

PR #7075, opened 2025-06-25
Polishing session view

mjbvz:
- Adds back button
- Adds open on web command
- Use codicons
- Fix overflow
- Add state restore

### DIFF
--- a/src/view/sessionLogView.ts
+++ b/src/view/sessionLogView.ts
@@ -7,21 +7,35 @@ import * as vscode from 'vscode';
 import type * as messages from '../../webviews/sessionLogView/messages';
 import { AuthProvider } from '../common/authentication';
 import { Disposable } from '../common/lifecycle';
-import { CopilotApi, SessionInfo } from '../github/copilotApi';
+import { ITelemetry } from '../common/telemetry';
+import { SessionPullInfo } from '../common/timelineEvent';
+import { CopilotApi } from '../github/copilotApi';
 import { CredentialStore } from '../github/credentials';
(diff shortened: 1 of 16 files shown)


DonJayamanne (approved):

src/view/sessionLogView.ts
```diff
@@ -87,46 +150,53 @@ export class SessionLogViewManager extends Disposable {
...
+		});
 
-		new SessionLogView(info, logs, webviewPanel);
+		webviewPanel.onDidChangeViewState(() => {
+			if (webviewPanel.active) {
+				this._activePanel = logView;
+			} else if (this._activePanel === logView && !webviewPanel.active) {
+				this._activePanel = undefined;
+			}
+		});
```
alexr00, line 180: Looks like this listener can leak

webviews/sessionLogView/sessionView.tsx
```diff
@@ -8,38 +8,52 @@ import type monacoType from 'monaco-editor';
...
 	const durationSec = Math.round(durationMs / 1000);
 
 	return (
 		<header className="session-header">
+			{pullInfo && (
+				<button
+					className="session-pull-button"
+					onClick={() => {
+						vscode.postMessage({ type: 'openPullRequestView' });
+					}}>
```
alexr00, line 51: You could also just use a link and use the PR URI format. 

https://github.com/microsoft/vscode-pull-request-github/blob/1e2ed4ef127421ed04ec8a0b50decd26d0385184/src/common/uri.ts#L543-L546

PR #7078, opened 2025-06-25
Implement remoteCodingAgents proposal to integrate in chat

joshspicer:
ref https://github.com/microsoft/vscode-copilot/issues/18918

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
 		"contribShareMenu",
 		"diffCommand",
 		"quickDiffProvider",
+		"remoteCodingAgents",
 		"shareProvider",
 		"tokenInformation",
 		"treeViewMarkdownMessage"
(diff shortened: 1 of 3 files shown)


src/github/copilotRemoteAgent.ts
```diff
@@ -108,7 +108,37 @@ export class CopilotRemoteAgentManager extends Disposable {
 		return continueWithCopilot;
 	}
 
-	async commandImpl() {
+	async commandImpl(args?: any) {
```
alexr00, line 111: Can this `any` be changed to a useful type?

package.json
```diff
@@ -814,6 +823,11 @@
 			]
 		},
 		"commands": [
+			{
+				"command": "githubpr.remoteAgent",
+				"title": "Remote agent integration",
```
alexr00, line 828: Title should be localized too!

src/github/copilotRemoteAgent.ts
joshspicer: changed in another PR 👍

package.json
joshspicer: ah yes!  FWIW I don't intend for it to show up in UIs, instead it should be called programatically)

PR #7113, opened 2025-06-27
Allow viewing multiple session logs per PR

mjbvz:
- Adds `View session` links to all sessions in a PR
- Show session name in the view
- Styling polish

cc @osortega I made a few tweaks to the API. Let me know if this causes problem for other parts of the code

### DIFF
--- a/src/github/copilotRemoteAgent.ts
+++ b/src/github/copilotRemoteAgent.ts
@@ -376,35 +376,34 @@ export class CopilotRemoteAgentManager extends Disposable {
 		return await capi.getLogsFromZipUrl(lastRun.logs_url);
 	}
 
-	async getMostRecentSessionLogsFromPullRequest(pullRequestId: number, completedOnly = true): Promise<IAPISessionLogs | undefined> {
+	async getSessionLogFromPullRequest(pullRequestId: number, sessionIndex = 0, completedOnly = true): Promise<IAPISessionLogs | undefined> {
 		const capi = await this.copilotApi;
 		if (!capi) {
 			return undefined;
 		}
 
 		const sessions = await capi.getAllSessions(pullRequestId);
-		const mostRecentSession = sessions.filter(s => !completedOnly || s.state === 'completed').at(0);
-		if (!mostRecentSession) {
+		const session = sessions.filter(s => !completedOnly || s.state === 'completed').at(sessionIndex);
+		if (!session) {
 			return undefined;
 		}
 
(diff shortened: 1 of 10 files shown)


TylerLeonhardt (approved):

src/github/pullRequestOverview.ts
```diff
@@ -458,9 +458,9 @@ export class PullRequestOverviewPanel extends IssueOverviewPanel<PullRequestMode
 		}
 	}
 
-	private async openSessionLog(_message: IRequestMessage<{ link: SessionPullInfo }>): Promise<void> {
+	private async openSessionLog(_message: IRequestMessage<{ link: SessionLinkInfo }>): Promise<void> {
```
alexr00, line 461: nit: can remove the `_` on `_message` as it's now used.

src/github/copilotRemoteAgent.ts
```diff
@@ -376,35 +376,34 @@ export class CopilotRemoteAgentManager extends Disposable {
...
+	async getSessionLogFromPullRequest(pullRequestId: number, sessionIndex = 0, completedOnly = true): Promise<IAPISessionLogs | undefined> {
 		const capi = await this.copilotApi;
 		if (!capi) {
 			return undefined;
 		}
 
 		const sessions = await capi.getAllSessions(pullRequestId);
-		const mostRecentSession = sessions.filter(s => !completedOnly || s.state === 'completed').at(0);
-		if (!mostRecentSession) {
+		const session = sessions.filter(s => !completedOnly || s.state === 'completed').at(sessionIndex);
```
alexr00, line 386: Is this filter still correct now that we're getting the session at a particular index?

src/github/copilotRemoteAgent.ts
```diff
@@ -376,35 +376,34 @@ export class CopilotRemoteAgentManager extends Disposable {
...
+	async getSessionUrlFromPullRequest(pullRequestId: number, sessionIndex = 0, completedOnly = true): Promise<string | undefined> {
 		const capi = await this.copilotApi;
 		if (!capi) {
 			return undefined;
 		}
 
 		const sessions = await capi.getAllSessions(pullRequestId);
-		const completedSessions = sessions.filter(s => s.state === 'completed');
-		if (completedSessions.length === 0) {
+		const session = sessions.filter(s => !completedOnly || s.state === 'completed').at(sessionIndex);
```
alexr00, line 402: Same as above, I don't think we should do the filter and the at.

alexr00 (commented):
👏

PR #7114, opened 2025-06-27
Handle jobs api errors

joshspicer:
reference https://github.com/github/sweagentd/blob/371ea6db280b9aecf790ccc20660e39a7ecb8d1c/internal/api/jobapi/handler.go#L110-L120

### DIFF
--- a/src/github/copilotApi.ts
+++ b/src/github/copilotApi.ts
@@ -56,12 +56,34 @@ export class CopilotApi {
 			body: JSON.stringify(payload)
 		});
 		if (!response.ok) {
-			const text = await response.text();
-			throw new Error(`Coding agent API error: ${response.status} ${text}`);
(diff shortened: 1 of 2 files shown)


src/github/copilotApi.ts
```diff
@@ -56,14 +56,36 @@ export class CopilotApi {
...
+			case 403:
+				return 'GitHub Coding Agent is not enabled for this repository';
+			case 404:
+				return 'Repository not found';
+			case 409:
+				return 'A Coding Agent pull request already exists';
+			case 500:
+				return 'Server error';
+			default:
+				return `Error: ${status}`;
```
TylerLeonhardt, line 85: Localized?

joshspicer: ah of course - thank you :)

PR #7312, opened 2025-07-16
Move the timeline methods back into the models

alexr00:
This is part 3 of adding a refresh event to the models


lszomoru (dismissed):

lszomoru:
@alexr00, PR build is 🔴

PR #7319, opened 2025-07-17
Change markFileAsViewed to keep files open by default

### LINKED ISSUE #3513: Mark as viewed close file option
Hi All,  Thanks for this fantastic extension!

I am in the process of reviewing a change to over 200 files. (yes, I know...)

I run VSCode with `"workbench.editor.enablePreview": false` (yes, I know.. I am a tab binger)

While reviewing the code I would like the file to close after each review is complete. Which does not happen if `"workbench.editor.enablePreview": false`.

Pardon me if this feature exists, I could not find it.

What I think would be valid options:
1. Close the window when I select Mark as Viewed &/or close the file under review when I select another file for review.
2. New options setting `githubpullrequest.editor.enablePreview` (Default: `true`)
3. Put the editor into `enablePreview` mode during any github pull request review with no user option to change.

For a user like me #1 seems optimal.  As I may have other windows open other than the review window.

But I think any of the solutions would be a big improvement for my scenario.

### Workaround
Just review a batch of files then close all in the window. (but this gets rid of some other windows I have open, no big deal really)

Thanks!

### LINKED ISSUE #5092: Option to *not* close file marked as viewed via keybinding
When you mark a file as viewed through the active editor menu bar or through a keybinding, the file will close. When you mark it as viewed through the tree view then it will not close. If a setting is needed I'll add one, but my hope is that this behavior is better for almost everyone.

_Originally posted by @alexr00 in https://github.com/microsoft/vscode-pull-request-github/issues/3513#issuecomment-1284469843_
           
Per this comment, it seems the default behavior for keybindings (or editor menu bar) calling `pr.markFileAsViewed` is to close the file, but for my workflow I'd prefer it didn't. Is it possible to add either a config setting, new command, or allow `"args"` in the keybinding to disable closing the active editor when this command is executed?

If it would be accepted, I could also attempt a PR for whichever of those options seems preferable. Thanks!

laharigandrapu11:
# Overview
When marking a file as viewed through the editor menu bar or keybinding, the file would automatically close. This behavior differed from using the tree view checkbox, which kept files open. This PR makes "keep open" the default for consistency, while providing a setting for users who prefer the old behavior.

# Changes
- Added new setting `githubPullRequests.closeFileOnMarkFileAsViewed` (default: false)
- Modified the markFileAsViewed command to respect this setting
- Default behavior now matches tree view (keeps files open)
- Users can opt-in to the old behavior via settings

# Demo
## New Default Behavior

image

Default behavior: The File remains open after marking as viewed

## Optional Behavior (Setting Enabled)
The screenshot below shows that the readme.md file closed when the checkmark was clicked
image
Optional behavior: File closes when setting is enabled

## Configuration
image
Users can control this behavior through VS Code settings

# Testing
Verified:
1. Default behavior (false) - files stay open when marked as viewed
2. Optional behavior (true) - files close when marked as viewed
3. Setting can be changed through VS Code settings UI
4. Tree view behavior remains unchanged

Fixes #5092


laharigandrapu11:

mogelbrod:
I'm not a maintainer - but I'm wondering if the additional setting and change in default behavior is necessary. Given that we can [pre-define arguments passed to keybinding commands](https://code.visualstudio.com/docs/configure/keybindings#_command-arguments) it should be sufficient to only add the optional argument (changing the default to close the file to keep existing behavior intact) to the command itself while achieving the same end result.

alexr00 (changes requested):
Thank you for the PR! I agree with @mogelbrod. The feedback in #5092 is that there should be a way to control whether the file is closed when using a keybinding. For this, we don't need a setting, just another argument for `options` added to the command implementation here:

https://github.com/microsoft/vscode-pull-request-github/blob/fd5f8d8f643c16d733cb775768312b9cdaaac236/src/commands.ts#L1521-L1522

For now, the `options` just needs to be something like this: 

```ts
{
    dontCloseFile: boolean;
}
```

PR #7354, opened 2025-07-19
Adding chat sessions provider

osortega:
Creating a chat sessions provider to show the coding agents

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -202,6 +202,11 @@ export async function closeAllPrAndReviewEditors() {
 	}
 }
 
+function isChatSessionWithPR(value: any): value is ChatSessionWithPR {
+	const asChatSessionWithPR = value as Partial<ChatSessionWithPR>;
+	return !!asChatSessionWithPR.pullRequest;
+}
+
 export function registerCommands(
 	context: vscode.ExtensionContext,
 	reposManager: RepositoriesManager,
@@ -854,10 +859,10 @@ export function registerCommands(
 				issueModel = argument.pullRequestModel;
 			} else if (argument instanceof PRNode) {
 				issueModel = argument.pullRequestModel;
-			} else if (argument && 'pullRequest' in argument) {
+			} else if (isChatSessionWithPR(argument)) {
 				issueModel = argument.pullRequest;
 			} else {
-				issueModel = argument as IssueModel;
+				issueModel = argument;
 			}
 		}
 
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -18,6 +18,7 @@ import { ITelemetry } from './common/telemetry';
 import { asTempStorageURI, fromPRUri, fromReviewUri, Schemes, toPRUri } from './common/uri';
 import { formatError } from './common/utils';
 import { EXTENSION_ID } from './constants';
+import { ChatSessionWithPR } from './github/copilotApi';
 import { CopilotRemoteAgentManager, ICopilotRemoteAgentCommandArgs } from './github/copilotRemoteAgent';
 import { FolderRepositoryManager } from './github/folderRepositoryManager';
 import { GitHubRepository } from './github/githubRepository';
@@ -836,7 +837,7 @@ export function registerCommands(
 		}),
 	);
 
-	async function openDescriptionCommand(argument: RepositoryChangesNode | PRNode | IssueModel | undefined) {
+	async function openDescriptionCommand(argument: RepositoryChangesNode | PRNode | IssueModel | ChatSessionWithPR | undefined) {
 		let issueModel: IssueModel | undefined;
 		if (!argument) {
 			const activePullRequests: PullRequestModel[] = reposManager.folderManagers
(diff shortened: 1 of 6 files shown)


src/github/copilotRemoteAgent.ts
```diff
@@ -110,6 +110,11 @@ export class CopilotRemoteAgentManager extends Disposable {
 		return await getCopilotApi(this.credentialStore, this.telemetry);
 	}
 
+	public async getUserLogin(): Promise<string | undefined> {
+		const capi = await this.copilotApi;
+		return await capi?.getUserLogin();
+	}
```
alexr00: `getUserLogin` doesn't appear to be used anywhere.

src/github/copilotApi.ts
```diff
@@ -173,6 +179,28 @@ export class CopilotApi {
 		return sessions.sessions;
 	}
 
+	public async getUserLogin(): Promise<string> {
+		return (await this.octokit.call(this.octokit.api.users.getAuthenticated)).data.login;
```
alexr00: You can get this from the `CredentialStore`. In the constructor, instead of passing in just the `token`, we could pass in the whole `CredentialStore`.

alexr00: https://github.com/microsoft/vscode-pull-request-github/blob/aaa863e4a255348556c3724095a1888933cee902/src/github/credentials.ts#L404-L405

src/github/copilotApi.ts
```diff
@@ -173,6 +179,28 @@ export class CopilotApi {
...
+	}
+
+	public async getAllCodingAgentPRs(repositoriesManager: RepositoriesManager): Promise<PullRequestModel[]> {
+		const username = await this.getUserLogin();
+		if (!username) {
+			Logger.error('Failed to get GitHub username from auth provider', CopilotApi.ID);
+			return [];
+		}
+		const query = `is:open is:pr assignee:${username} archived:false`;
```
alexr00: You should be able to do `is:open author:copilot-swe-agent[bot] involves:<put the user login here> is:pr`, then you don't need to filter by author at the end.

src/github/copilotApi.ts
```diff
@@ -173,6 +179,28 @@ export class CopilotApi {
...
+		const allItems = await Promise.all(
+			repositoriesManager.folderManagers.map(async fm => {
+				const result = await fm.getPullRequests(0, undefined, query);
```
alexr00:
```suggestion
				const result = await fm.getPullRequests(PRType.Query, undefined, query);
```

src/github/copilotApi.ts
alexr00: Actually, since you're going through the folder manager, you can do this:
`is:open author:copilot-swe-agent[bot] involves:${user} is:pr`
And you shouldn't need to go through all folder managers since passing in a query shouldn't be filtering to a specific repo.

src/commands.ts
```diff
@@ -853,8 +854,10 @@ export function registerCommands(
 				issueModel = argument.pullRequestModel;
 			} else if (argument instanceof PRNode) {
 				issueModel = argument.pullRequestModel;
+			} else if (argument && 'pullRequest' in argument) {
```
alexr00: The `in` check is dangerous for future refactoring, I've been trying to remove them. Something like this is safer:
```
function isChatSessionWithPR(value: any): value is ChatSessionWithPR {
	const asChatSessionWithPR  = value as Partial<CatSessionWithPR>;
	return !!asChatSessionWithPR.pullRequest;
}

Comment (alexr00) [reply src/commands.ts]: That would also remove the need for the `as IssueModel` below.

File: src/github/copilotRemoteAgent.ts (hunk 1 above)
Comment (osortega) [reply src/github/copilotRemoteAgent.ts]: Good catch! Removed it

File: src/github/copilotApi.ts (hunk 2 above)
Comment (osortega) [reply src/github/copilotApi.ts]: @alexr00 I'm passing now the credentialStore besides the token since there's quite a few callers that check for the return value of `getCopilotApi` that depends on being able to get a token.
If I were to refactor all the callers then it might become a bigger PR

File: src/github/copilotApi.ts (hunk 3 above)
Comment (osortega) [reply src/github/copilotApi.ts]: Ahhh perfect! Thanks for the tip

File: src/github/copilotApi.ts (hunk 4 above)
Comment (osortega) [reply src/github/copilotApi.ts]: Changed it! Thanks

File: src/commands.ts (hunk 5 above)
Comment (osortega) [reply src/commands.ts]: Makes sense! Changed it

File: src/github/copilotApi.ts [hunk 6]
```diff
@@ -201,6 +230,23 @@ export class CopilotApi {
 		}
 		return await logsResponse.text();
 	}
+
+	private getHub() {
```
rzhao271: Nit: what is the return type?

src/github/copilotApi.ts
```diff
@@ -201,6 +230,23 @@ export class CopilotApi {
...
+			authProvider = AuthProvider.githubEnterprise;
+		} else if (this.credentialStore.isAuthenticated(AuthProvider.github)) {
+			authProvider = AuthProvider.github;
+		} else {
+			return;
+		}
+
+		const github = this.credentialStore.getHub(authProvider);
+		if (!github) {
+			return;
```
rzhao271: Is this separate case necessary?

rzhao271 (approved):
LGTM. Added some comments for polish

src/github/copilotApi.ts
osortega: Ahh good catch! Fixed it

rzhao271 (dismissed):

alexr00 (dismissed):

PR #7373, opened 2025-07-22
experiment chat editor integration

rebornix:


src/github/copilotRemoteAgent.ts
```diff
@@ -391,7 +394,24 @@ export class CopilotRemoteAgentManager extends Disposable {
 		});
 
 		this._onDidChangeChatSessions.fire();
-		vscode.commands.executeCommand('vscode.open', webviewUri);
+		if (vscode.workspace.getConfiguration('chat').get('agentSessionsViewLocation') !== 'disabled') {
```
joshspicer: This core setting will be undefined on old clients with this new extension. Update engine or change this check?

rebornix: @joshspicer good catch (deleted my old comment)

PR #7393, opened 2025-07-24
Create deno.yml

EvaCoreBot:


alexr00:
Thanks for the PR, but we aren't using deno at the moment.

PR #7479, opened 2025-07-31
Fix microsoft/vscode#259022 always fallback when tool call content is empty

rebornix:


lszomoru:
@rebornix, PR build is 🔴

rzhao271 (approved):

PR #7528, opened 2025-08-07
Copilot remote agent tests

osortega:
Adding tests for copilot remote agent functionality

### DIFF
--- a/src/test/common/sessionParsing.test.ts
+++ b/src/test/common/sessionParsing.test.ts
@@ -11,6 +11,7 @@ import {
 	toFileLabel,
 	SessionResponseLogChunk
 } from '../../../common/sessionParsing';
+import { diffHeaders, diffNoAts, simpleDiff } from './fixtures/gitdiff/sessionParsing';
 
 describe('sessionParsing', function () {
 	describe('parseSessionLogs()', function () {
@@ -209,17 +210,7 @@ another non-data line`;
 
 	describe('parseDiff()', function () {
 		it('should parse diff content correctly', function () {
-			const diffContent = `diff --git a/src/file.ts b/src/file.ts
-index 1234567..abcdefg 100644
---- a/src/file.ts
-+++ b/src/file.ts
-@@ -1,4 +1,4 @@
- export function hello() {
--  console.log('hello');
-+  console.log('hello world');
- }`;
-
-			const result = parseDiff(diffContent);
+			const result = parseDiff(simpleDiff);
 
 			assert(result);
 			assert.strictEqual(result.fileA, '/src/file.ts');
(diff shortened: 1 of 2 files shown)


src/lm/tools/copilotRemoteAgentTool.ts
```diff
@@ -127,7 +127,10 @@ export class CopilotRemoteAgentTool implements vscode.LanguageModelTool<CopilotR
 			};
 			const buffer: Buffer = Buffer.from(JSON.stringify(preferredRendering));
 			const data: Uint8Array = Uint8Array.from(buffer);
-			lmResult.push(new vscode.LanguageModelDataPart2(data, 'application/pull-request+json', [vscode.LanguageModelPartAudience.User]));
+
+			// API might not be available for tests, this guarantees we are still able to test
+			const userAudience = (vscode as any).LanguageModelPartAudience?.User ?? 1;
```
osortega: @alexr00 Had to do this because the value was not available during runtime for the tests, not sure if there's a better way to make sure the value is available

src/test/common/sessionParsing.test.ts
```diff
@@ -0,0 +1,298 @@
...
+			const result = parseToolCallDetails(toolCall, '');
+
+			assert.strictEqual(result.toolName, 'Read repository');
+			assert.strictEqual(result.invocationMessage, 'Read repository');
+		});
+	});
+
+	describe('parseDiff()', function () {
+		it('should parse diff content correctly', function () {
+			const diffContent = `diff --git a/src/file.ts b/src/file.ts
```
osortega: Disabled the hygiene script for this since I couldn't figre out a way to disable it for the line.
@alexr00 is there a way to do that?

dbaeumer (dismissed):

src/lm/tools/copilotRemoteAgentTool.ts
alexr00: There is a way! Took 5 tries to get this with Copilot. It really didn't want to do this: https://github.com/microsoft/vscode-pull-request-github/pull/7530

src/test/common/sessionParsing.test.ts
alexr00: The best thing to do would be to add these strings as test fixtures under ./test/common/fixtures. Maybe in a file ./test/common/fixtures/sessionParsing.ts and just export them all as strings from there.

alexr00 (changes requested):
See https://github.com/microsoft/vscode-pull-request-github/pull/7528#discussion_r2259984980

src/test/common/sessionParsing.test.ts
osortega: Perfect! That worked

alexr00 (approved):

PR #7533, opened 2025-08-08
Remove some `any`s, add a rule disallowing some `any`s, and use a const string for swe agent

alexr00:


.vscode/settings.json
```diff
@@ -4,6 +4,14 @@
 		".eslintrc.json": "jsonc",
 		".eslintrc.*.json": "jsonc"
 	},
+	"json.schemas": [
```
aeschli, line 7: This association is defined by the eslint extension, and comes in through that.

aeschli (dismissed):

bpasero (dismissed):
Give `any` no chance!

.vscode/settings.json
alexr00: It doesn't quite catch the `.eslintrc.*.json` format that we use here. This ESLint includes:

image

aeschli (approved):

PR #7539, opened 2025-08-08
Fix too many refreshes when closing a PR

### LINKED ISSUE #7537: Closing a PR causes a flurry of search API calls
We're listening for the close event in more than one place 😞

alexr00:
Fixes #7537

### DIFF
--- a/src/view/prsTreeModel.ts
+++ b/src/view/prsTreeModel.ts
@@ -40,6 +40,7 @@ export class PrsTreeModel extends Disposable {
 
 	private _cachedPRs: Map<FolderRepositoryManager, Map<string | PRType.LocalPullRequest | PRType.All, ItemsResponseResult<PullRequestModel>>> = new Map();
 	private readonly _repoEvents: Map<FolderRepositoryManager, vscode.Disposable[]> = new Map();
+	private _getPullRequestsForQueryLock: Promise<void> = Promise.resolve();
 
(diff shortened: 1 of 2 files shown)


src/view/prsTreeModel.ts
```diff
@@ -238,26 +239,36 @@ export class PrsTreeModel extends Disposable {
...
-			PRType.Query,
-			{ fetchNextPage },
-			query,
-		);
-		cache.set(query, prs);
+		let release: () => void;
+		const lock = new Promise<void>(resolve => { release = resolve; });
+		const prev = this._getPullRequestsForQueryLock;
+		this._getPullRequestsForQueryLock = prev.then(() => lock);
+		await prev;
```
alexr00, line 246: We get two events when a PR is closed: timeline updated and state changed. Since this method is async, we could call the search API two times when this happens as the cache has not yet been updated. To prevent this, we lock `getPullRequestsForQuery` such it can only be executed once at a time.

src/view/treeNodes/pullRequestNode.ts
```diff
@@ -54,11 +54,6 @@ export class PRNode extends TreeNode implements vscode.CommentingRangeProvider2
 		super(parent);
 		this.registerSinceReviewChange();
 		this.registerConfigurationChange();
-		this._register(this.pullRequestModel.onDidChange((e) => {
-			if (e.title || e.state) {
-				this.refresh(this);
-			}
-		}));
```
alexr00, line 61: We listen to this event and refresh based on it in another place. Doing so here as well results in double the refreshes.

PR #7641, opened 2025-08-22
Checkout PR at the end of a completed chat session

joshspicer:
closes https://github.com/microsoft/vscode/issues/260227

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -651,11 +651,27 @@ export function registerCommands(
 		return { folderManager, pr };
 	};
 
-	context.subscriptions.push(vscode.commands.registerCommand('pr.checkoutFromDescription', async (context: OverviewContext | undefined) => {
-		if (!context) {
+	context.subscriptions.push(vscode.commands.registerCommand('pr.checkoutFromDescription', async (ctx: OverviewContext | { path: string } | undefined) => {
+		if (!ctx) {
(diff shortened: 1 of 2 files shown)


src/commands.ts
```diff
@@ -651,11 +651,27 @@ export function registerCommands(
...
-		const resolved = await resolvePr(context);
+
+		if ('path' in ctx) {
+			const { path } = ctx;
+			const prNumber = Number(Buffer.from(path.substring(1), 'base64').toString('utf8'));
+			if (Number.isNaN(prNumber)) {
+				return vscode.window.showErrorMessage(vscode.l10n.t('Unable to parse pull request number.'));
+			}
+			const folderManager = reposManager.folderManagers[0];
+			const pullRequest = await folderManager.fetchById(folderManager.gitHubRepositories[0], Number(prNumber));
```
alexr00, line 666: This is not going to work for multi-repo. You can easily end up getting the wrong PR. In fully multi repo (usually multiroot workspace) setup, you could get a PR from an unrelated repo.

If there are multiple folder managers and github repos, it would be better to actually find the correct repo to fetch from. You can probably use `getManagerForFile`:

https://github.com/microsoft/vscode-pull-request-github/blob/aa02ca643af80ca4d43590a98435a67629d69393/src/github/repositoriesManager.ts#L144-L145

Issue #7266, opened 2025-07-09
Coding Agent: Allow user to decide whether change goes to fork or upstream


joshspicer:
Coding agent will prefer the user's fork (if they have it set up as a remote), as determined by existing GHPR APIs. 

@eleanorjboyd shared that the automatic behavior here was undesired.  We should somehow make this configurable.

eleanorjboyd:
Heres a video showing the whole thing: 

https://github.com/user-attachments/assets/8f866cc7-2afe-41c6-bf2e-5ac6fc7eb141

joshspicer:
Do you have your fork set as a remote (`git remote -v`)?

eleanorjboyd:
```
(base) eleanorboyd@Eleanors-MacBook-Pro-2 vscode-python-environments % git remote -v
karthiknadig    https://github.com/karthiknadig/vscode-python-environments.git (fetch)
karthiknadig    https://github.com/karthiknadig/vscode-python-environments.git (push)
origin  git@github.com:eleanorjboyd/vscode-python-environments.git (fetch)
origin  git@github.com:eleanorjboyd/vscode-python-environments.git (push)
upstream        git@github.com:microsoft/vscode-python-environments.git (fetch)
upstream        git@github.com:microsoft/vscode-python-environments.git (push)
```

joshspicer:
This is technically working as designed (closing https://github.com/microsoft/vscode-copilot/issues/18806) but agreed options are good

micahgodbolt:
For most large msft projects we set upstream for the msft repo and origin for our fork, which means new branches are pushed to our fork instead of the main repo. We then create PRs from our fork.

This means that using this extension every time we check out a branch it doesn't use the local branch, it pulls pr/name/prnumber. This means if we make changes and push, those changes are pushed to a new pr/name/prnumber branch instead of the branch originally used to create the PR.

Is there any way to fix this?

joshspicer:
I think configuring `githubPullRequests.remotes` to only include the remote you're interested in would work around this.

Next pre-release you'll be prompted (if your repo has >1 valid GitHub remotes) to select your preference. Ther will also be a new command to toggle between remotes.  This is cached per-workspace

<img width="914" height="359" alt="Image" src="https://github.com/user-attachments/assets/2ab06529-f9f4-4057-a8f2-0b346bc83cff" />

I think this will help @micahgodbolt - please let me know once it's out!

eleanorjboyd:
[...]

PR #7645, opened 2025-08-22
Let user select which github remote to target coding agent sessions against

joshspicer:
closes https://github.com/microsoft/vscode-pull-request-github/issues/7266

### DIFF
--- a/package.json
+++ b/package.json
@@ -1741,6 +1741,17 @@
 				"category": "%command.pull.request.category%"
 			},
 			{
+				"command": "pr.preferredCodingAgentGitHubRemote",
+				"title": "%command.pr.preferredCodingAgentGitHubRemote.title%",
+				"icon": "$(gear)",
+				"category": "%command.pull.request.category%"
+			},
+			{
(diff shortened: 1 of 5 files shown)


Tyriar (approved):

TylerLeonhardt (approved):

package.json
```diff
@@ -1740,6 +1740,17 @@
...
+				"command": "pr.preferredCodingAgentGitHubRemote",
+				"title": "%command.pr.preferredCodingAgentGitHubRemote.title%",
+				"icon": "$(gear)",
+				"category": "%command.pull.request.category%"
+			},
+			{
+				"command": "pr.resetCodingAgentPreferences",
+				"title": "%command.pr.resetCodingAgentPreferences.title%",
+				"category": "%command.pull.request.category%"
+			},
```
alexr00, line 1753: Should the commands be accessible in the command palette even when the user is not signed in? Might be worth adding them in the command palette menu contribution and setting the `when` clause such that they only show when the user is signed in.

PR #7743, opened 2025-09-03
Populate thinking progress to chat editor

rebornix:


justschen (dismissed):

justschen:
@rebornix some errors here

justschen (approved):

src/github/copilotRemoteAgent/chatSessionContentBuilder.ts
```diff
@@ -388,6 +388,10 @@ export class ChatSessionContentBuilder {
 			const toolDetails = parseToolCallDetails(toolCall, deltaContent);
 			toolPart.toolName = toolDetails.toolName;
 
+			if (toolPart.toolName === 'think') {
```
alexr00, line 391: It looks like this `think` string is used in many places. It would be great to have some string constants for all the tools we know about.

PR #7772, opened 2025-09-08
More welcome view states in the Chat Sessions view

alexr00:
Fixes https://github.com/microsoft/vscode/issues/259285


alexr00 (commented):
The issue was that we had a few state where no welcome view would show:
- No folder open
- Folder open, but no git repos
- Folder opne, but no GitHub repos
This change addes welcome views for those states.

PR #7776, opened 2025-09-08
Fix apostrophes

alexr00:
Fixes microsoft/vscode#264527


src/github/copilotRemoteAgent/chatSessionContentBuilder.ts
```diff
@@ -76,7 +76,7 @@ export class ChatSessionContentBuilder {
 				// if this is the first response, then also add the PR card
 				if (sessionIndex === 0) {
 					const uri = await toOpenPullRequestWebviewUri({ owner: pullRequest.remote.owner, repo: pullRequest.remote.repositoryName, pullRequestNumber: pullRequest.number });
-					const plaintextBody = marked.parse(pullRequest.body, { renderer: new PlainTextRenderer(), }).trim();
+					const plaintextBody = marked.parse(pullRequest.body, { renderer: new PlainTextRenderer(), smartypants: true }).trim();
```
mjbvz, line 79: This is an ok workaround but likely not complete and in future marked releases the `smartypants` option is removed. I think we should fix this in the `PlainTextRenderer` instead

PR #7785, opened 2025-09-09
fix #7783 (candidate)

### LINKED ISSUE #7783: `_No logs available for this session_` race condition
On some machines the user ends up here because the coding agent session is still !== `in_progress`. This was fixed for some entrypoints (a new chat session), but not for the 'cloud button'

https://github.com/microsoft/vscode-pull-request-github/blob/b7ed43e0ea2ca80a66eb4819e7207668c0680d69/src/github/copilotRemoteAgent/chatSessionContentBuilder.ts#L101-L102

related fix https://github.com/microsoft/vscode-pull-request-github/pull/7675

joshspicer:
closes https://github.com/microsoft/vscode-pull-request-github/issues/7783

Moves `await this.waitForQueuedToInProgress(...)` to be awaited after any new session is created, instead of just in certain cases

### DIFF
--- a/src/test/mocks/mockExtensionContext.ts
+++ b/src/test/mocks/mockExtensionContext.ts
@@ -21,9 +21,13 @@ export class MockExtensionContext implements ExtensionContext {
 		store(key: string, value: string): Thenable<void> {
 			throw new Error('Method not implemented.');
 		}
+		keys(): Thenable<string[]> {
+			throw new Error('Method not implemented.');
+		}
 		delete(key: string): Thenable<void> {
 			throw new Error('Method not implemented.');
 		}
+
 		onDidChange!: Event<SecretStorageChangeEvent>;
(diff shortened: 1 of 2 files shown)


src/test/mocks/mockExtensionContext.ts
```diff
@@ -21,9 +21,13 @@ export class MockExtensionContext implements ExtensionContext {
 		store(key: string, value: string): Thenable<void> {
 			throw new Error('Method not implemented.');
 		}
+		keys(): Thenable<string[]> {
```
joshspicer, line 24: Necessary change for CI now that SecretStorage.keys() API is finalized

alexr00 (approved):
Seems reasonable

joshspicer:
Discussed and we will include this in a GHPR recovery build

PR #7829, opened 2025-09-22
Filter out repeat sessions

joshspicer:
Before: 
Screenshot 2025-09-22 at 9 39 19 AM

After: 
image

Capi seems to now return >1 copy of a session. 
image


osortega (approved):

osortega:
Looks like we don't need to do this anymore, would be good to verify

src/github/copilotRemoteAgent/chatSessionContentBuilder.ts
```diff
@@ -30,9 +30,13 @@ export class ChatSessionContentBuilder {
 		capi: CopilotApi,
 		timelineEventsPromise: Promise<TimelineEvent[]>
 	): Promise<Array<vscode.ChatRequestTurn | vscode.ChatResponseTurn2>> {
-		const sortedSessions = sessions.slice().sort((a, b) =>
-			new Date(a.created_at).getTime() - new Date(b.created_at).getTime()
-		);
+		const sortedSessions = sessions
```
osortega, line 33: Looks like we don't need to do this anymore, would be good to verify

PR #7849, opened 2025-09-24
implement vscode#268089

joshspicer:
ref https://github.com/microsoft/vscode/issues/268063

Implement and depends on `chatSessionsProvider@2`, added in https://github.com/microsoft/vscode/pull/268089

CD will fail until `chatSessionsProvider@2` is available

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
 		"activeComment",
 		"chatParticipantAdditions",
 		"chatParticipantPrivate",
-		"chatSessionsProvider",
+		"chatSessionsProvider@2",
 		"codiconDecoration",
 		"codeActionRanges",
 		"commentingRangeHint",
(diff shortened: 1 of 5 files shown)


rebornix (approved):

package.json
```diff
@@ -14,7 +14,7 @@
 		"activeComment",
 		"chatParticipantAdditions",
 		"chatParticipantPrivate",
-		"chatSessionsProvider",
+		"chatSessionsProvider@2",
```
roblourens, line 17: The api version can only be used in extensions that have it enabled in distro, under `extensionsEnabledWithApiProposalVersion`. I assume it will just be ignored and not break anything

roblourens: Right https://github.com/microsoft/vscode/blob/422af1583191adbd102f2c3b259c73738a1ead5b/src/vs/platform/extensionManagement/common/extensionGalleryService.ts#L905-L907

joshspicer: nice, thanks for checking

PR #7872, opened 2025-09-26
Fix for refresh button

osortega:
This fixes the refresh button so that we show the progress in the view.
Before we were triggering and forgetting about it which is confusing on the tree view since it doesn't instantly do the refresh


alexr00:
@osortega and I discussed, and really I need to reconcile `PrsTreeModel` and `CopilotStateModel` for this to be fixed properly 🙈. I will tackle this in debt week.

PR #7876, opened 2025-09-27
Update to use eslint9

mjbvz:
Updates the project to use eslint 9. This switches to use a flat config instead of multiple config files. I also needed to update some of the eslint dependencies

I generally tried to keep the existing logic and rules with this change. We can revisit these later

I did end up having to make a few code changes for new errors that showed up after updating


alexr00:
@mjbvz CI is failing

mjbvz:
Look like the pipeline is using an old node version. Seeing if I can bump it to node 17 or newer

azure-pipeline.nightly.yml
```diff
@@ -31,6 +31,8 @@ extends:
 
     l10nSourcePaths: ./src
 
+    nodeVersion: "20.x"
```
mjbvz, line 34: Adding these to make sure we're using a consistent node version across pipelines. The default in the template was already node 20

PR #7890, opened 2025-09-29
sort so origin is first (…

joshspicer:
closes https://github.com/microsoft/vscode/issues/265367

### DIFF
--- a/src/github/copilotRemoteAgent.ts
+++ b/src/github/copilotRemoteAgent.ts
@@ -397,6 +397,9 @@ export class CopilotRemoteAgentManager extends Disposable {
 			return;
 		}
 
+		// Sort so that origin is first if it exists
+		ghRemotes.sort((a, b) => (a.remoteName === 'origin' ? -1 : b.remoteName === 'origin' ? 1 : 0));
+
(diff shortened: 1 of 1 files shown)


src/github/copilotRemoteAgent.ts
```diff
@@ -397,6 +397,9 @@ export class CopilotRemoteAgentManager extends Disposable {
 			return;
 		}
 
+		// Sort so that origin is first if it exists
+		ghRemotes.sort((a, b) => (a.remoteName === 'origin' ? -1 : b.remoteName === 'origin' ? 1 : 0));
```
Tyriar, line 401: I don't think this is a stable sort, is that a problem?

Screenshot 2025-09-30 at 8 57 20 am

```
[{ data: 1, remote: 'origin' }, { data: 2, remote: 'origin' }].sort((a, b) => (a.remote === 'origin' ? -1 : b.remote === 'origin' ? 1 : 0));
```

You would need to also cover the case when both are origin in order to be stable.

joshspicer: I don't think it's a problem, it's just changing the order to be presented in a quickpick (moving origin to the top for convenience)

PR #7900, opened 2025-10-01
Ignore repos that aren't in one of the workspace folders

### LINKED ISSUE #7896: Ignore worktrees that aren't the workspace folder if they duplicate a repo
Got an example from @digitarald where we end up doing everything 4 times because there are 3 worktrees in a repo.

alexr00:
Fixes #7896

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -21,6 +21,7 @@ import { BRANCH_PUBLISH, EXPERIMENTAL_CHAT, FILE_LIST_LAYOUT, GIT, IGNORE_SUBMOD
 import { initBasedOnSettingChange } from './common/settingsUtils';
 import { TemporaryState } from './common/temporaryState';
 import { Schemes } from './common/uri';
+import { isDescendant } from './common/utils';
 import { EXTENSION_ID, FOCUS_REVIEW_MODE } from './constants';
 import { createExperimentationService, ExperimentationTelemetry } from './experimentationService';
 import { CopilotRemoteAgentManager } from './github/copilotRemoteAgent';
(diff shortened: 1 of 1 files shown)


Tyriar (approved):

Wetmelon:
If I only open subfolders in a worktree, this no longer opens any pull requests associated with the checked-out branch, bricking the entire extension.  Is there a setting I can use to disable this behavior?  Will roll back to an older version as a workaround.

Think Monorepo with several out-of-tree git worktrees, and I only want to open a few subfolders of one of the worktrees to keep my workspace small.

EDIT:  Confirmed that rolling back to older version restores the expected functionality👍

alexr00:
@Wetmelon thanks for pointing this out. Maybe the behavior we need is to show repos that are within one of the workspace folders _or_ contain one of the workspace folders.

frontier159:
@alexr00 yeh I agree with @Wetmelon -- I frequently open a specific package sub-directly, which isn't the root.

Can it not just auto add the `git rev-parse --show-toplevel` directory into the list to show?

frontier159:
Oh looks like it might be addressed in https://github.com/microsoft/vscode-pull-request-github/pull/8005?

pWoLiAn:
2025-11-03 08:55:35.569 [info] [Activation] Repo file:///root/vega/framework is not in a workspace folder, ignoring.

extension is ignoring valid submodules due to this change

alexr00:
What is the value of your settings `git.detectSubmodules` and `githubPullRequests.ignoreSubmodules`?

pWoLiAn:
default values i.e. git.detectSubmodules=true and githubPullRequests.ignoreSubmodules=false

PR #7932, opened 2025-10-03
Include more details in error messages and handle permission errors

pwang347:
| Before | After |
| -- | -- |
|  image |  image |


src/github/copilotRemoteAgent.ts
```diff
@@ -683,6 +683,26 @@ export class CopilotRemoteAgentManager extends Disposable {
 		}
 		const { owner, repo, remote, repository, ghRepository, baseRef } = repoInfo;
 
+		// Check if user has permission to access the repository
+		try {
+			await ghRepository.octokit.api.repos.get({ owner, repo });
```
joshspicer, line 724: We already have a way to see if coding agent is assignable by the user, which is probably the better check here?

https://github.com/microsoft/vscode-pull-request-github/blob/d03583336e8f10aa2a3c79f45588d34925653ab9/src/github/copilotRemoteAgent.ts#L307

pwang347: what do you think about generating separate errors for both? I think if there's no visibility to the repository whatsoever (e.g., wrong github account) it might be useful to see that in addition to knowing whether you have permission to assign Copilot

src/github/copilotRemoteAgent.ts
```diff
@@ -689,6 +689,37 @@ export class CopilotRemoteAgentManager extends Disposable {
...
+		} catch (error) {
+			if (error.status === 404 || error.status === 403) {
+				const currentUser = await ghRepository.octokit.api.users.getAuthenticated().then((u) => u.data.login).catch(() => 'unknown');
```
alexr00: You can get the current user from `this.credentialStore`.

joshspicer (dismissed):

PR #7933, opened 2025-10-03
Use word dict for git branch naming

pwang347:
image

Fixes https://github.com/microsoft/vscode/issues/264492


TylerLeonhardt:
@pwang347 👋 for PRs of code that you aren't the owner for you should add the owner as the reviewer since they will ultimately be the one to own the code in the future 🙂 Once you have dedicated ownership of something, then you can send PRs without reviewers which will land in our `#codereview` channel for anyone to review.

pwang347:
@TylerLeonhardt got it, thanks for the heads up!

joshspicer:
btw (for branches we create) we do that here https://github.com/microsoft/vscode-pull-request-github/blob/62ffd00b5f363d69d4bb8b0612807ccd2dcb6b27/src/github/copilotRemoteAgent/gitOperationsManager.ts#L16-L17

If the branch wasn't created from VS Code, _or_ if we don't specify it when [posting the job](https://github.com/microsoft/vscode-pull-request-github/blob/62ffd00b5f363d69d4bb8b0612807ccd2dcb6b27/src/github/copilotApi.ts#L77) then it will get a random branch from GH.

joshspicer:
Note that the branch name needs to have the `copilot/` prefix for reasons

joshspicer (approved):
looks good to me, thank you!  I will indeed keep Alex for reivewing the dependency :)

src/github/copilotRemoteAgent/gitOperationsManager.ts
```diff
@@ -13,7 +14,10 @@ export class GitOperationsManager {
 
 	async commitAndPushChanges(repoInfo: RepoInfo) {
 		const { repository, remote, baseRef } = repoInfo;
-		const asyncBranch = `copilot/vscode${Date.now()}`;
+		const config = vscode.workspace.getConfiguration('git');
```
alexr00: We have a file with all the settings constants: 

https://github.com/microsoft/vscode-pull-request-github/blob/c8bc7248b305f2cce381fdddf117ddccd0208183/src/common/settingKeys.ts#L66-L67

src/github/copilotRemoteAgent/gitOperationsManager.ts
```diff
@@ -13,7 +14,10 @@ export class GitOperationsManager {
...
+		const branchWhitespaceChar = config.get<string>('branchWhitespaceChar')!;
```
alexr00: This can also be added to `settingsKeys.ts`.

src/github/copilotRemoteAgent/gitOperationsManager.ts
```diff
@@ -138,4 +142,47 @@ export class GitOperationsManager {
 			}
 		}
 	}
+
+	// Adapted from https://github.com/microsoft/vscode/blob/e35e3b4e057450ea3d90c724fae5e3e9619b96fe/extensions/git/src/commands.ts#L3007
+	private async generateRandomBranchName(repository: Repository, prefix: string, separator: string): Promise<string> {
+		const config = vscode.workspace.getConfiguration('git');
+		const branchRandomNameDictionary = config.get<string[]>('branchRandomName.dictionary')!;
```
alexr00: Same comment as above RE settings constants

alexr00: There are many bad examples of using `!` in this repo, but for new code we should try to avoid it. You can specify a default value for `get`ing a config by passing in a second parameter. 

If there is no dictionary, we should gracefully fall back to something else, rather than throwing as will happen here.

alexr00: I think it's fair to fall back to the old behavior if the the dictionary is undefined or empty.

src/github/copilotRemoteAgent/gitOperationsManager.ts
```diff
@@ -138,4 +142,47 @@ export class GitOperationsManager {
...
+			return '';
+		}
+
+		// 5 attempts to generate a random branch name
+		for (let index = 0; index < 5; index++) {
+			const randomName = prefix + '/' + uniqueNamesGenerator({
+				dictionaries,
+				length: dictionaries.length,
+				separator
+			});
```
alexr00: In general, it's safer to use template strings rather than string concatonation with +.


```suggestion
			const randomName = `${prefix}/${uniqueNamesGenerator({
				dictionaries,
				length: dictionaries.length,
				separator
			})}`;
```

src/github/copilotRemoteAgent/gitOperationsManager.ts
pwang347: updated, thank you!

src/github/copilotRemoteAgent/gitOperationsManager.ts
pwang347: updated, let me know if the new changes look good to you!

PR #7934, opened 2025-10-03
Add inline coding agent confirmation

joshspicer:
ref https://github.com/microsoft/vscode/issues/260228


src/github/copilotRemoteAgent.ts
```diff
@@ -175,25 +217,13 @@ export class CopilotRemoteAgentManager extends Disposable {
...
-				return {};
-			}
-
-			const uri = await toOpenPullRequestWebviewUri({ owner: pullRequest.remote.owner, repo: pullRequest.remote.repositoryName, pullRequestNumber: pullRequest.number });
-			const plaintextBody = marked.parse(pullRequest.body, { renderer: new PlainTextRenderer(true), smartypants: true }).trim();
+			stream.confirmation(
+				vscode.l10n.t('Delegate to coding agent'),
+				DELEGATE_MODAL_DETAILS,
+				{ step: 'create' },
+				['Delegate', 'Cancel']
```
alexr00, line 224: Are these shown to the user? If so, they should also be localized.

PR #7962, opened 2025-10-09
only show TODO code lens on lines that start with a comment token

joshspicer:
fix for https://github.com/microsoft/vscode-pull-request-github/pull/7954 being far too verbose.

### DIFF
--- a/package.json
+++ b/package.json
@@ -609,7 +609,6 @@
 					"default": [
 						"TODO",
 						"todo",
-						"BUG",
 						"FIXME",
 						"ISSUE",
 						"HACK"


connor4312:
I suggest adding ` * ` (with surrounding spaces_ for JS/TS docblock comments too, and `///` for Rust and C# docblock styles

joshspicer:
nice 

image

package.json
```diff
@@ -616,6 +616,21 @@
...
+					"items": {
+						"type": "string",
+						"description": "%githubIssues.createIssueCommentPrefixes.items%"
+					},
+					"default": [
+						"//",
+						"#",
+						"--",
+						" * ",
+						"///"
```
TylerLeonhardt: This seems suspicious. It's something that would be language specific too.

```
<!-- TODO: ...... -->
```

```
/*
TODO: some multi-line
comment with context
*/
```

```
<#
TODO: PowerShell block comments are weird
I know
#>
```

https://github.com/microsoft/vscode/blob/55f06131905f07dde60f301aa56be6e2f52155a3/extensions/make/language-configuration.json#L2-L7

The comment syntax is declared by the language support, I feel like we should somehow leverage that if we can.

TylerLeonhardt: https://github.com/microsoft/vscode/blob/55f06131905f07dde60f301aa56be6e2f52155a3/src/vs/editor/common/languages/languageConfigurationRegistry.ts#L35-L47

we have this in core... and we have a number of commands in core that are about retrieval of symbols:
https://code.visualstudio.com/api/references/commands#commands

I wonder if this warrants some way to get the LanguageConfiguration via a Command. Or at least just the comment stuff...

@hediet @alexdima I see your name around this code, maybe you have ideas

alexr00 (changes requested):
I don't think we should have two setting that both control whether the code action/codelens show.

Can we instead just remove "BUG" from the list of defaults?

joshspicer:
@alexr00 I still think it's incorrect to match outside of a comment. Do you disagree?

@TylerLeonhardt thanks for the idea/pointers

package.json
joshspicer: Thanks for the pointers, this looks a lot more correct.

I justified that this was ok since it's configurable via a setting and did not know we could possibly leverage core for this information.

alexr00:
We can check if there's a comment at that position: 

https://github.com/microsoft/vscode-pull-request-github/blob/6de5b34e36a697cd7eb9f82c133e126ce656ca70/src/issues/util.ts#L519-L520

joshspicer:
isComment works great - thank you @alexr00!

image

src/issues/issueTodoProvider.ts
```diff
@@ -114,7 +114,11 @@ export class IssueTodoProvider implements vscode.CodeActionProvider, vscode.Code
 
 		const codeLenses: vscode.CodeLens[] = [];
 		for (let lineNumber = 0; lineNumber < document.lineCount; lineNumber++) {
-			const line = document.lineAt(lineNumber).text;
+			const textLine = document.lineAt(lineNumber);
+			const { text: line, firstNonWhitespaceCharacterIndex } = textLine;
+			if (!(await isComment(document, new vscode.Position(lineNumber, firstNonWhitespaceCharacterIndex)))) {
```
joshspicer, line 123: Can't decide what is more efficient/correct to do first, the comment check or the `findTodoInLine()` check.  @alexr00 preference?

alexr00: IIRC, `isComment` ends up being a round trip to the renderer, so `findTodoInLine` is probably cheaper time-wise to do first.

PR #7974, opened 2025-10-10
Fix is updating check

alexr00:


src/github/pullRequestOverview.ts
```diff
@@ -217,6 +217,7 @@ export class PullRequestOverviewPanel extends IssueOverviewPanel<PullRequestMode
 	}
 
 	protected override async updateItem(pullRequestModel: PullRequestModel): Promise<void> {
+		this._isUpdating = true;
```
hediet, line 220: What bit me quite a lot in the past with this pattern is reentrancy, especially if it is an async method (thats why we had inline completions loading indicator spinning forever ;) ).
I recommend to either throw if `_isUpdating` is already set to true, or to use a counter `updatingCounter`, that you increment/decrement.

alexr00: I will throw, thanks for the idea!

PR #8001, opened 2025-10-15
Use pencil for drafts

alexr00:
Fixes microsoft/vscode#271439


alexr00:
@Tyriar we could, but probably not this milestone. I would rather not let perfect be the enemy of improvement.

alexr00:
https://github.com/microsoft/vscode/issues/271523

PR #8009, opened 2025-10-16
Support emoji completions in review comments

### LINKED ISSUE #7178: Add emoji completions for `:smile:` style emoji
It’s already super useful to review code in VSCode. It would be even better if we could directly add multiple media types—such as emojis and support for uploading images and videos—without needing to switch to other tools or return to the browser.

Image
Image

🥺🙏

kycutler:
4e3fee72-6676-4976-ab94-5c4bb19161e3

Closes #7178

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.120.1
+
+### Fixes
+
+- Extension cannot find git repo when VS Code didn't open the git root directory. https://github.com/microsoft/vscode-pull-request-github/issues/7964
+
 ## 0.120.0
 
 ### Changes
--- a/src/@types/vscode.proposed.chatParticipantPrivate.d.ts
+++ b/src/@types/vscode.proposed.chatParticipantPrivate.d.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-// version: 10
+// version: 11
 
 declare module 'vscode' {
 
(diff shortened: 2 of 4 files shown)


src/view/emojiCompletionProvider.ts
```diff
@@ -0,0 +1,65 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { ensureEmojis, getEmojis } from '../common/emoji';
+import { Schemes } from '../common/uri';
+
+export class EmojiCompletionProvider implements vscode.CompletionItemProvider {
+	private static readonly ID: string = 'EmojiCompletionProvider';
```
alexr00: Looks like this ID isn't used anywhere. In other classes, this is used when calling the logger, so that we can see where the log is coming from.

src/view/emojiCompletionProvider.ts
```diff
@@ -0,0 +1,65 @@
...
+	private _emojiCompletions: vscode.CompletionItem[] = [];
+
+	constructor(private context: vscode.ExtensionContext) {
```
alexr00: nit: inconsistent use of privates starting with `_`. This codebase is bad about this, but we can try for consistency within new files. Let's go with using `_` for privates.

src/common/emoji.ts
```diff
@@ -36,3 +36,10 @@ export function emojify(message: string) {
 		return emojiMap?.[code] || s;
 	});
 }
+
+export function getEmojis(): Record<string, string> {
+	if (emojiMap === undefined) {
+		return {};
+	}
+	return emojiMap;
+}
```
alexr00: Instead of having a separate `getEmojis`, lets just have `ensureEmojis` return the map. Then you don't need to make two calls when all you want is the map.

alexr00 (changes requested):
A few small changes, plus one thing that doesn't work they way I'd expect: If I type a full emoji expression (ex. `:running:`, I don't expect the picker to reset.

Image

src/common/emoji.ts
kycutler: Good call, updated

kycutler:
> A few small changes, plus one thing that doesn't work they way I'd expect: If I type a full emoji expression (ex. `:running:`, I don't expect the picker to reset.

Good catch! Updated so it will only trigger if the leading `:` appears after whitespace or beginning of line.

kycutler:
Good feedback! Improved (and simplified) the logic quite a bit -- now it also handles cases like placing the cursor within an existing emoji text and pressing Ctrl+Space.

alexr00 (approved):
Works great!

Looks like there's an unrelated build errror, due to an API change that wasn't pick up in the extension over night 🙄. After I make the fix, you can just use the "update branch" button on github.com to pull in changes without losing my review approval.

PR #8017, opened 2025-10-17
Fix build break due to API change

alexr00:


mjbvz:
Thanks for the fix!

We want to migrate from using `id` to a uri  (`resource`) to identify chat sessions. This will allow us to open other stuff like custom editors instead of always opening our native chat session view: https://github.com/microsoft/vscode/pull/271723

For now we will shim in the `resource` if it's not provided. At some point though we'll need a bigger push to switch over the API in a breaking way. I'll let you know when that happens so we can coordinate

PR #8022, opened 2025-10-20
fix: ignore null in timeline event

ipcjs:
Some PR timeline arrays can contain null values, which causes downstream parsing errors
Ignoring null entries fixes the parsing error.

for example: https://github.com/TranscodeGroup/docker/pull/23. 
截屏2025-10-20 17 56 26

### DIFF
--- a/src/github/graphql.ts
+++ b/src/github/graphql.ts
@@ -262,7 +262,7 @@ export interface TimelineEventsResponse {
 	repository: {
 		pullRequest: {
 			timelineItems: {
-				nodes: (MergedEvent | Review | IssueComment | Commit | AssignedEvent | HeadRefDeletedEvent)[];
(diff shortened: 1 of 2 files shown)


ipcjs:

alexr00:
@ipcjs _that's_ where the `__typename` error was coming from? Nice find. Based on the query, that shouldn't be possible:

https://github.com/ipcjs/vscode-pull-request-github/blob/688855c6807e1c505ea2b81585e69360cb3a9781/src/github/queriesShared.gql#L296-L319

GitHub must have changed something. How'd you figure this out?

ipcjs:
This PR(https://github.com/TranscodeGroup/docker/pull/23) was added to a private Project, maybe that's related?


截屏2025-10-20 18 20 24

PR #8038, opened 2025-10-24
update chatSessionType  -> resourceScheme

joshspicer:
this is the way to do it post-https://github.com/microsoft/vscode/pull/272013

### DIFF
--- a/package.json
+++ b/package.json
@@ -3529,33 +3529,33 @@
 			"chat/chatSessions": [
 				{
 					"command": "pr.openChanges",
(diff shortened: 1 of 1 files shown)


rebornix (approved):

mjbvz:
Thanks @joshspicer! Going to take a look to see if we can support the previous syntax too for codex

mjbvz:
Hm seems like `chatSessionType` still works for me in a simple extension. Will keep testing as I think `chatSessionType` is probably more clear but glad we have this workaround for now

PR #8039, opened 2025-10-24
Contribute open changes button to copilot-cloud-agent

joshspicer:


bhavyaus (approved):

rebornix (approved):

src/commands.ts
```diff
@@ -727,7 +727,22 @@ export function registerCommands(
...
-			} else {
+			}
+			else if (contextHasPath(pr)) {
+				// looks like '/123' where 123 is the PR number
+				const { path } = pr;
+				const prNumber = Number(path.startsWith('/') ? path.substring(1) : path);
+				if (Number.isNaN(prNumber)) {
+					return vscode.window.showErrorMessage(vscode.l10n.t('Unable to parse pull request number.'));
+				}
+				const folderManager = reposManager.folderManagers[0];
```
alexr00, line 749: Do we have any other info about which repo it is? Choosing the first folderManager and the first github repo will sometimes result in the wrong repo getting picked.

alexr00 (dismissed):
Conflicts.

src/commands.ts
joshspicer: I agree. The part right now only provides the URI (and now managed from another extension, so it's harder to map that). Taking a look if we can expose more information that is available to the editor's chat part

PR #8044, opened 2025-10-24
do not exclude markdown when detecting comment for Delegate codelens

joshspicer:
### Before
image

### After
image


joshspicer:
cc/ @alexr00  I am not sure why `isComment` excludes markdown (forcing it to return true) in the first place.  I did this change to avoid a behavior change in other places, but may want to change the default behavior of `isComment`?

alexr00:
It's because we always want to show the issue completion provider and issue link provider (which is actually disabled) in markdown. But, this can be moved into the issue completion provider.

PR #8105, opened 2025-11-06
Adopt scm/repository

alexr00:


package.json
```diff
@@ -3146,6 +3146,18 @@
 					"group": "1_modification@5"
 				}
 			],
+			"scm/repository": [
+				{
+					"command": "pr.create",
+					"when": "scmProvider =~ /^git|^remoteHub:github/ && scmProviderRootUri in github:reposNotInReviewMode",
+					"group": "navigation"
+				},
+				{
```
lszomoru: Do you really want to add it in both places?

alexr00: I thought that was best pactice. Anything that is added to the inline menu should also show in the context menu.

lszomoru: Given that menu items in the inline group can be hidden and then it will move to the `...` I would only add it inline

alexr00: GHPR has been following this best practice for just about every command (and where it doesn't, it's a mistake). If the best practice has changed, then it should be adopted for all menus in GHPR. For now, I will add it to both and we can discuss if the best practice has changed.

alexr00: Ok, I'm seeing that extension tree menus behave differently here. You cannot right click and hide actions in extension trees.

alexr00: I'll remove this from the context menu, and we can discuss in standup what should be done about the inline actions in extension tree.

lszomoru (approved):

PR #8126, opened 2025-11-10
Comments don't show when PR is on non-default repo

### LINKED ISSUE #8050: Comments don't show when PR is on non-default repo
When I make a PR to a repo that is also the default repo (the repo the "from" repo in the PR was forked from), I can see the comments in the associated PR "inline" in the editors (and also in the comments window) when that branch is checked out. Great!

However, this does not work when I'm submitting a PR to a repo that is not the default in the sense above. To be specific, we have repo "A" which is the the orginal repo for some project. There is a GH fork of this repo, repo "B" (under an organization), and then my own fork of "B", call it "C" under my user. Specifically "A" is [this](https://github.com/scylladb/seastar), "B" is [this](https://github.com/redpanda-data/seastar) and "C" is [this](https://github.com/travisdowns/seastar).

My workload is that I have all three configured as remotes, and I always push my feature branches to "C" and from there I open pull requests against "A" or "B" but most often "A". When I open them against "B", comments for the PRs show fine, but against "A" they do not. I can still see the PRs in the "Created by me" list in the GH PR extension sidebar, but not by default (I have to click "continue fetching from other remotes" once). I can see the contents of the PR in that view, open diffs, etc, but the comments do not show inline in the editable view of the same files even when the PR branch is checked out locally.

If I right click "Checkout pull request" from the GH PR menu, then the comments _do_ appear inline, but now I'm not on my local branch any more but a `pr/*` style branch like `pr/travisdowns/3068`.

- Extension version: 0.120.1
- VSCode Version: 1.105.1
- OS: Ubuntu 22.04 and 24.04
- Repository Clone Configuration (single repository/fork of an upstream repository):
- GitHub Product (GitHub.com/GitHub Enterprise version x.x.x): github.com

Steps to Reproduce:

1. As above

alexr00:
Fixes #8050

### DIFF
--- a/src/view/pullRequestCommentControllerRegistry.ts
+++ b/src/view/pullRequestCommentControllerRegistry.ts
@@ -20,9 +20,15 @@ interface PullRequestCommentHandlerInfo {
 	dispose: () => void;
 }
 
+interface PRCommentingRangeProviderInfo {
+	provider: vscode.CommentingRangeProvider2;
+	refCount: number;
+	dispose: () => void;
+}
+
 export class PRCommentControllerRegistry extends Disposable implements vscode.CommentingRangeProvider, CommentReactionHandler {
 	private _prCommentHandlers: { [key: number]: PullRequestCommentHandlerInfo } = {};
-	private _prCommentingRangeProviders: { [key: number]: vscode.CommentingRangeProvider2 } = {};
+	private _prCommentingRangeProviders: { [key: number]: PRCommentingRangeProviderInfo } = {};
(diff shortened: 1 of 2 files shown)


src/view/pullRequestCommentControllerRegistry.ts
```diff
@@ -20,9 +20,15 @@ interface PullRequestCommentHandlerInfo {
 	dispose: () => void;
 }
 
+interface PRCommentingRangeProviderInfo {
+	provider: vscode.CommentingRangeProvider2;
+	refCount: number;
+	dispose: () => void;
+}
```
Tyriar, line 27: `extends IDisposable`?

PR #8153, opened 2025-11-17
Fix only one reviewer on PR page

alexr00:


alexr00:
@copilot, add tests for the two changed functions.

PR #8173, opened 2025-11-19
Create devcontainer.json

ljhlh49-dot:


ljhlh49-dot (commented):
Run

alexr00:
Thanks for the PR, but we won't be taking this change right now.

PR #8177, opened 2025-11-20
 Fix method call

alexr00:


mjbvz:
Oh it's because of the record overload :/

```ts
function t(message: string, args: Record<string, any>): string;
```

I guess function's match `Record<string, any>`.

PR #8321, opened 2025-12-19
Disable buggy sticky scroll

alexr00:


webviews/editorWebview/index.css
```diff
@@ -48,7 +48,7 @@ textarea:focus,
 }
 
 .title {
-	position: sticky;
+	/* position: sticky; */
```
Tyriar, line 51: Should these comments be removed or add TODOs to make it clear it's coming back?

PR #8335, opened 2026-01-05
Improve commit line unwrapping

alexr00:
See https://github.com/microsoft/vscode-pull-request-github/pull/8255#issuecomment-3708315953

### DIFF
--- a/src/github/folderRepositoryManager.ts
+++ b/src/github/folderRepositoryManager.ts
@@ -3074,11 +3074,94 @@ function unwrapCommitMessageBody(body: string): string {
 	// Pattern to detect fenced code block markers
 	const FENCE_PATTERN = /^[ \t]*```/;
 
+	const getLeadingWhitespaceLength = (text: string): number => text.match(/^[ \t]*/)?.[0].length ?? 0;
+	const hasHardLineBreak = (text: string): boolean => / {2}$/.test(text);
+	const appendWithSpace = (base: string, addition: string): string => {
+		if (!addition) {
+			return base;
+		}
+		return base.length > 0 && !/\s$/.test(base) ? `${base} ${addition}` : `${base}${addition}`;
+	};
+
 	const lines = body.split('\n');
 	const result: string[] = [];
 	let i = 0;
 	let inFencedBlock = false;
-	let inListContext = false;
+	const listIndentStack: number[] = [];
+
+	const getNextNonBlankLineInfo = (
+		startIndex: number,
+	): { line: string; indent: number; isListItem: boolean } | undefined => {
+		for (let idx = startIndex; idx < lines.length; idx++) {
+			const candidate = lines[idx];
+			if (candidate.trim() === '') {
+				continue;
+			}
+			return {
+				line: candidate,
+				indent: getLeadingWhitespaceLength(candidate),
+				isListItem: LIST_ITEM_PATTERN.test(candidate),
+			};
+		}
+		return undefined;
+	};
+
+	const getActiveListIndent = (lineIndent: number): number | undefined => {
+		for (let idx = listIndentStack.length - 1; idx >= 0; idx--) {
+			const indentForLevel = listIndentStack[idx];
+			if (lineIndent >= indentForLevel + 2) {
+				listIndentStack.length = idx + 1;
+				return indentForLevel;
+			}
+			listIndentStack.pop();
+		}
+		return undefined;
+	};
+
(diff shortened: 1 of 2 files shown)


src/github/folderRepositoryManager.ts
```diff
@@ -3074,11 +3074,94 @@ function unwrapCommitMessageBody(body: string): string {
...
+		if (!currentLine) {
+			return false;
+		}
+
+		const trimmed = currentLine.trim();
+		if (!trimmed) {
+			return false;
+		}
+
+		if (hasHardLineBreak(baseLine) || hasHardLineBreak(currentLine)) {
```
dmitrivMS, line 3132: nit: it's a little inconsistent that some regex-es are tested as "variable.test(...)" and some are wrapped into local functions.

dmitrivMS (approved):

src/test/github/folderRepositoryManager.test.ts
```diff
@@ -222,4 +222,12 @@ describe('titleAndBodyFrom', function () {
 		assert.strictEqual(result?.title, 'title');
 		assert.strictEqual(result?.body, 'Wrapped paragraph across lines.\n\n- Item 1\n  - Nested item\n    More nested content\n- Item 2\n\nAnother wrapped paragraph here.');
```
justingrant, line 223: Same here: I think the nested lines should be merged. 

```suggestion
		assert.strictEqual(result?.body, 'Wrapped paragraph across lines.\n\n- Item 1\n  - Nested item More nested content\n- Item 2\n\nAnother wrapped paragraph here.');
```

src/test/github/folderRepositoryManager.test.ts
justingrant, line 215: Something weird is happening with the display of this suggestion where the original code is shown as a blank line. But anyways the list continuation line should not have a line-break before it.

```suggestion
		assert.strictEqual(result?.body, 'This is wrapped text that should be joined.\n\n- List item with continuation\n- Another item');
```

src/test/github/folderRepositoryManager.test.ts
```diff
@@ -222,4 +222,12 @@ describe('titleAndBodyFrom', function () {
...
 	});
+
+	it('handles nested lists', async function () {
```
justingrant, line 226: @alexr00 Sorry, I wasn't quick enough to review this PR before merging, but I'd suggest adding a few more test cases, including some using numbered list items which are harder to implement than unordered lists because the numeric part can be variable-length.  Also, the number of spaces is actually quite lenient in GitHub. Finally, code blocks and additional paragraphs are also supported in GH lists, so should probably have test cases for those.

I didn't review this PR code closely, but I did see a few `2`'s in the PR code, so wanted to make sure that you weren't assuming that the list indentation was always 2 spaces per level and/or that the list marker was always one character. 

Here's some test cases, adapted from https://chatgpt.com/share/695c1f08-b928-8004-88c1-79417405b729.  All the cases above should have their lines joined where there's a single `\n`. I apologize if some of these duplicate codepaths that were tested in other tests!

I suspect it'd be good to also make nested versions of these tests... I'll leave that as an exercise for copilot! :-) 

1. Basic numeric list
   continuation.
   Third line

1. Additional spaces are
    OK for a continuation (unless it's 4 spaces which would be a code block).
    Third line

* Additional spaces are
    OK for a continuation (unless it's 4 spaces which would be a code block).
    Third line

10. Multi-digit numbers should also
    work for a continuation.
    Third line

1. Multi-paragraph lists are also supported.

   Second paragraph in the same list item.
   Third line

* Multi-paragraph lists are also supported.

  Second paragraph in the same list item.
  Third line

1. Item with code:
   
       code line
       code line

* Item with code:
   
      code line
      code line

1. Fewer spaces are also OK
 for a list continuation (as long as there's at least one space)

* Fewer spaces are also OK
 for a list continuation (as long as there's at least one space)

1. Fewer spaces are also OK
  for a list continuation (as long as there's at least one space)

* Fewer spaces are also OK
  for a list continuation (as long as there's at least one space)

alexr00: Thank you for the test cases!

src/test/github/folderRepositoryManager.test.ts
alexr00: I think since they have preceding whitespace they shouldn't be merged, right? The preceding whitespace indicates an intentional new line, where as no preceding whitespace should be unwrapped.

alexr00: Actually, I see github.com does unwrap these.

Issue #8351, opened 2026-01-06
Switch branch quick pick is not showing all branches


sandy081:
Testing #8340

version - `0.125.2026010604`

I was trying to switch the main branch and it does not show me all branches  - see the screenshot below

<img width="1281" height="255" alt="Image" src="https://github.com/user-attachments/assets/11519e02-d488-4f83-9045-e4ed77b05f45" />

where are the GItHub shows correctly

<img width="935" height="582" alt="Image" src="https://github.com/user-attachments/assets/272a6065-4125-45e0-9235-5ed43b6403a9" />

PR #8365, opened 2026-01-07
Switch branch quick pick is not showing all branches

alexr00:
Fixes microsoft/vscode-pull-request-github#8351

### DIFF
--- a/src/common/async.ts
+++ b/src/common/async.ts
@@ -36,13 +36,13 @@ export function throttle<T>(fn: () => Promise<T>): () => Promise<T> {
 	return trigger;
 }
 
-export function debounce(fn: () => any, delay: number): () => void {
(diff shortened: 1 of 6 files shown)


src/common/async.ts
```diff
@@ -36,13 +36,13 @@ export function throttle<T>(fn: () => Promise<T>): () => Promise<T> {
 	return trigger;
 }
 
-export function debounce(fn: () => any, delay: number): () => void {
+export function debounce<T extends (...args: any[]) => any>(fn: T, delay: number): (...args: Parameters<T>) => void {
```
dmitrivMS, line 39: This feels dangerous - what if someone calls it with different args, it will only be called once - could the consumer expect it to be called with every unique set of args?

bpasero (approved):

PR #8583, opened 2026-03-10
Link to local file for permalinks in webview

### LINKED ISSUE #8571: Local-ized Links in PR and Issue Overviews
Currently, editor-embedded comments with permalinks get `replacePermalink`ed to make the links instead navigate to the local version of the file (with a `view on GitHub` backup). It would be nice if permalinks in PR/Issue descriptions and comments on the overview pages did the same.

Not only would this improve consistency, as currently clicking the same link in two different places has two different behaviors, but it also facilitates new workflows and usage patterns like having a TOC of PR file links in the sidebar and navigating through them as you review a PR.

Happy to help with an implementation if you agree this would be useful.

Daniel-Aaron-Bloom:
This pull request adds the functionality to link to a file within the project in a comment on the PR/issue overview pages. This feature creates parity with #5558, so links appear and function identically in both locations. It also enables users to quickly jump around using a list of referenced file locations in the PR description or a comment.

Fixes #8571

This PR also adds a few `await this._waitForReady;` in `_replyMessage` and `_throwError`, which I think are the correct behavior, and also without them this extension sometimes fails to respond.

### DIFF
--- a/src/test/common/utils.test.ts
+++ b/src/test/common/utils.test.ts
@@ -53,7 +53,6 @@ describe('utils', () => {
 	});
 
 	describe('processPermalinks', () => {
-		const repoOwner = 'microsoft';
 		const repoName = 'vscode';
 		const authority = 'github.com';
 		const sha = 'a'.repeat(40);
@@ -65,7 +64,7 @@ describe('utils', () => {
 
 		it('should add data attributes when file exists locally', async () => {
 			const html = makePermalink('src/file.ts', 10);
-			const result = await utils.processPermalinks(html, repoOwner, repoName, authority, async () => true);
+			const result = await utils.processPermalinks(html, repoName, authority, async () => true);
 
 			assert(result.includes('data-local-file="src/file.ts"'));
 			assert(result.includes('data-start-line="10"'));
@@ -77,7 +76,7 @@ describe('utils', () => {
 
 		it('should set end line when range is specified', async () => {
 			const html = makePermalink('src/file.ts', 10, 20);
-			const result = await utils.processPermalinks(html, repoOwner, repoName, authority, async () => true);
+			const result = await utils.processPermalinks(html, repoName, authority, async () => true);
 
 			assert(result.includes('data-start-line="10"'));
 			assert(result.includes('data-end-line="20"'));
@@ -85,35 +84,35 @@ describe('utils', () => {
 
 		it('should not modify links when file does not exist locally', async () => {
 			const html = makePermalink('src/file.ts', 10);
-			const result = await utils.processPermalinks(html, repoOwner, repoName, authority, async () => false);
+			const result = await utils.processPermalinks(html, repoName, authority, async () => false);
 
 			assert.strictEqual(result, html);
 		});
 
 		it('should not modify non-permalink links', async () => {
 			const html = '<a href="https://example.com">example</a>';
-			const result = await utils.processPermalinks(html, repoOwner, repoName, authority, async () => true);
+			const result = await utils.processPermalinks(html, repoName, authority, async () => true);
 
 			assert.strictEqual(result, html);
 		});
 
 		it('should not modify links to a different repo', async () => {
 			const html = `<a href="https://github.com/other/repo/blob/${sha}/src/file.ts#L10">link</a>`;
-			const result = await utils.processPermalinks(html, repoOwner, repoName, authority, async () => true);
+			const result = await utils.processPermalinks(html, repoName, authority, async () => true);
 
 			assert.strictEqual(result, html);
 		});
 
 		it('should skip already processed links', async () => {
 			const html = `<a data-permalink-processed="true" href="https://github.com/microsoft/vscode/blob/${sha}/src/file.ts#L10">link</a>`;
-			const result = await utils.processPermalinks(html, repoOwner, repoName, authority, async () => true);
+			const result = await utils.processPermalinks(html, repoName, authority, async () => true);
 
 			assert.strictEqual(result, html);
 		});
 
 		it('should process multiple links independently', async () => {
 			const html = makePermalink('src/exists.ts', 1) + makePermalink('src/missing.ts', 2);
-			const result = await utils.processPermalinks(html, repoOwner, repoName, authority, async (path) => path === 'src/exists.ts');
+			const result = await utils.processPermalinks(html, repoName, authority, async (path) => path === 'src/exists.ts');
 
 			assert(result.includes('data-local-file="src/exists.ts"'));
 			assert(!result.includes('data-local-file="src/missing.ts"'));
@@ -121,14 +120,14 @@ describe('utils', () => {
 
 		it('should return original HTML when fileExistsCheck throws', async () => {
 			const html = makePermalink('src/file.ts', 10);
-			const result = await utils.processPermalinks(html, repoOwner, repoName, authority, async () => { throw new Error('fail'); });
+			const result = await utils.processPermalinks(html, repoName, authority, async () => { throw new Error('fail'); });
 
 			assert.strictEqual(result, html);
 		});
 
 		it('should handle links without surrounding text', async () => {
 			const html = makePermalink('src/file.ts', 5);
-			const result = await utils.processPermalinks(html, repoOwner, repoName, authority, async () => true);
+			const result = await utils.processPermalinks(html, repoName, authority, async () => true);
 
 			assert(result.includes('link text'));
 			assert(result.includes('data-local-file="src/file.ts"'));


Daniel-Aaron-Bloom:

src/common/webview.ts
```diff
@@ -74,6 +74,7 @@ export class WebviewBase extends Disposable {
 			seq: originalMessage.req,
 			res: message,
 		};
+		await this._waitForReady;
```
alexr00, line 77: Why wait here?

src/common/webview.ts
```diff
@@ -82,6 +83,7 @@ export class WebviewBase extends Disposable {
 			seq: originalMessage?.req,
 			err: error,
 		};
+		await this._waitForReady;
```
alexr00, line 86: Why wait here?

webviews/editorWebview/app.tsx
```diff
@@ -11,6 +11,86 @@ import { PullRequest } from '../../src/github/views';
...
+interface PermalinkAnchor {
+	element: HTMLAnchorElement;
+	url: string;
+	file: string;
+	startLine: number;
+	endLine: number;
+}
+
+function findUnprocessedPermalinks(
+	root: Document | Element,
```
alexr00: The approach of finding the permalinks in the webview involves multiple back-and-forths with the extension. I would suggest a different approach:
- In the extension (probably in issueOverview.ts or pullRequestOverview.ts)
    - Find all the permalinks
	- Check for file existance
	- Replace permalinks with some very easy to find "tag" which contains all the relevant info
- In the webview
    - Find this easy to find "tag"
	- Parse the relevant info from it
	- Replace with an `<a>` who's on click handler uses the message passing (context.tsx) to open the relevant file

src/common/webview.ts
Daniel-Aaron-Bloom: For the same reason we wait in `_postMessage` above

> 	```ts
> protected async _postMessage(message: any) {
> 		// Without the following ready check, we can end up in a state where the message handler in the webview
> 		// isn't ready for any of the messages we post.
> 		await this._waitForReady;
> 		this._webview?.postMessage({
> 			res: message,
> 		});
> 	}
> ```

src/common/webview.ts
Daniel-Aaron-Bloom: Same as above

webviews/editorWebview/app.tsx
Daniel-Aaron-Bloom: I can give that a shot.

src/common/webview.ts
Daniel-Aaron-Bloom: In practice without this, the extension simply fails to work about half the time in a dev container (never got a failure outside of the container). The await calls to get the file mapping simply never resolve (presumably because the message got dropped)

webviews/editorWebview/app.tsx
Daniel-Aaron-Bloom: Done, and I added support for diff links too.

src/common/webview.ts
Daniel-Aaron-Bloom: With the refactor, this became less of an issue, but it still seems like the right thing to do.

Daniel-Aaron-Bloom:
Yup. Much cleaner and more efficient. I've also added diff link support to open the diff view.

src/github/issueOverview.ts
```diff
@@ -321,10 +321,13 @@ export class IssueOverviewPanel<TItem extends IssueModel = IssueModel> extends W
 			this._item = issue as TItem;
 			this.setPanelTitle(this.buildPanelTitle(issueModel.number, issueModel.title));
 
+			// Process permalinks in bodyHTML before sending to webview
+			issue.bodyHTML = await this.processLinksInBodyHtml(issue.bodyHTML);
```
alexr00: Overwriting the original `bodyHTML` might come back to bite us later: What if some other part of the extension needs the original HTML? We've just overwritten it and lost it. 

Instead of setting the `bodyHTML` here, can we pass a separate `bodyHTML` when we post the message to `pr.initialize1?

alexr00: Same for the timeline events. We shouldn't overwrite the `bodyHTML`, but we can copy the timeline events.

src/github/issueOverview.ts
```diff
@@ -618,7 +658,7 @@ export class IssueOverviewPanel<TItem extends IssueModel = IssueModel> extends W
 			if (allAssignees) {
 				const newAssignees: IAccount[] = allAssignees.map(item => item.user);
 				await this._item.replaceAssignees(newAssignees);
-				const events = await this._getTimeline();
+				const events = await this.processTimelineEvents(await this._getTimeline());
```
alexr00: Instead of calling `await this.processTimelineEvents(await this._getTimeline())` everywhere, could we move the `processTimelineEvents` into `_getTimeline`?

src/github/issueOverview.ts
alexr00: Same in pullRequestOverview.ts.

src/github/pullRequestOverview.ts
```diff
@@ -638,6 +676,33 @@ export class PullRequestOverviewPanel extends IssueOverviewPanel<PullRequestMode
...
+		try {
+			const { file, startLine } = message.args;
+			const fileChanges = await this._item.getFileChangesInfo();
+			const change = fileChanges.find(
+				fileChange => fileChange.fileName === file || fileChange.previousFileName === file,
+			);
+
+			if (!change) {
+				Logger.warn(`Could not find file ${file} in PR changes`, PullRequestOverviewPanel.ID);
+				return;
```
alexr00, line 696: We should fall back to opening the original link in the browser.

src/github/utils.ts
```diff
@@ -334,6 +334,116 @@ async function transformHtmlUrlsToExtensionUrls(body: string, githubRepository:
...
+	rootUri: vscode.Uri
+): Promise<string> {
+	try {
+		const repoName = escapeRegExp(githubRepository.remote.repositoryName);
+		const repoOwner = escapeRegExp(githubRepository.remote.owner);
+		const authority = escapeRegExp(githubRepository.remote.gitProtocol.url.authority);
+
+		// Process blob permalinks (exclude already processed links)
+		const blobPattern = new RegExp(
+			`<a\\s+(?![^>]*data-permalink-processed)([^>]*?href="https?:\/\/${authority}\/${repoOwner}\/${repoName}\/blob\/[0-9a-f]{40}\/([^"#]+)#L(\\d+)(?:-L(\\d+))?"[^>]*?)>([^<]*?)<\/a>`,
```
alexr00: Can we use named captures here?

webviews/editorWebview/app.tsx
```diff
@@ -41,6 +41,34 @@ export function Root({ children }) {
...
 
+	useEffect(() => {
+		const handleLinkClick = (event: MouseEvent) => {
+			const target = event.target as HTMLElement;
+			const anchor = target.closest('a[data-local-file]');
+			if (anchor) {
+				const file = anchor.getAttribute('data-local-file');
+				const startLine = anchor.getAttribute('data-start-line');
+				const endLine = anchor.getAttribute('data-end-line');
+				const linkType = anchor.getAttribute('data-link-type');
```
alexr00: Can we move this metadata extraction next to where the anchor tag is constructed? That way, all the knowledge of what these attributes are is in the same place.

src/common/webview.ts
alexr00: Interesting, thanks!

webviews/editorWebview/app.tsx
Daniel-Aaron-Bloom: Done. I had to move both to `src/common/utils.ts` to deal with import restrictions in the webview. Let me know if there's a better place for it.

src/github/issueOverview.ts
Daniel-Aaron-Bloom: I think I've done as you've suggested, but I went a little further and pushed the replacement to be inside `getInitializeContext`, which require making that `async`, which unlocked doing the same for `processTimelineEvents`. Processing them inside seemed a bit cleaner than outside, plus it takes care of both PRs and Issues in once place without materially impacting timing (`await` inside vs outside shouldn't be a big deal in this case)

src/common/utils.ts
```diff
@@ -1014,4 +1014,165 @@ export function truncate(value: string, maxLength: number, suffix = '...'): stri
...
+	fileExistsCheck: (filePath: string) => Promise<boolean>
+): Promise<string> {
+	try {
+		const escapedRepoName = escapeRegExp(repoName);
+		const escapedRepoOwner = escapeRegExp(repoOwner);
+		const escapedAuthority = escapeRegExp(authority);
+
+		// Process blob permalinks (exclude already processed links)
+		const blobPattern = new RegExp(
+			`<a\\s+(?![^>]*data-permalink-processed)([^>]*?href="https?:\/\/${escapedAuthority}\/${escapedRepoOwner}\/${escapedRepoName}\/blob\/[0-9a-f]{40}\/(?<filePath>[^"#]+)#L(?<startLine>\\d+)(?:-L(?<endLine>\\d+))?"[^>]*?)>(?<linkText>[^<]*?)<\/a>`,
```
alexr00: Should a link like this work (I know it's a private repo, but the structure of the link): https://github.com/alexr00/playground/blob/f86be02708e48ed4648b224771995c7573213946/readme.md?#L6

I'm not seeing it get replaced for a PR in that repo.

Daniel-Aaron-Bloom: It should work, I'm using it in private repos right now (and it's being very helpful). It only doesn't replace the permalinks if the file doesn't exist locally (so you have to check out a PR branch before opening the view) and diff links if the files isn't present in the diff.

These two should show up in this repo, although at the time I made this comment the latter did not because of the different repoOwner (it has since been updated and now both appear to work):

https://github.com/microsoft/vscode-pull-request-github/pull/8583/changes#diff-0660e0b0f883e874d471ed55dbd1bc1a9e009f9422952d55ac7979b8daa23141R77

https://github.com/Daniel-Aaron-Bloom/vscode-pull-request-github/blob/854b78ed243e91c2e431dca323c33a06c2fe343e/common/views.ts#L14

Daniel-Aaron-Bloom: Screenshot 2026-03-18 at 8 58 48 AM

All 4 are clickable and work properly for me.

Issue #8191, opened 2025-12-02
Support enterprise uri only for vscode-pull-request-github


Changqing-JING:
Now I have a problem, my github repo is at enterprise server "https://example-github-enterprise.com"

And my copilot is at another enterprise server:
https://xxx.ghe.com

Then it lead to a problem, to use github pull request plugin, I need:
`"github-enterprise.uri": "https://example-github-enterprise.com"`
to use copilot, I need:
`"github-enterprise.uri": "https://xxx.ghe.com"`

But there is the github-enterprise.uri field is a global config which impacts both vscode-pull-request-github and copilot.
Is this possible to add a enterprise.uri field inside the config of vscode-pull-request-github.

marcinbukowczan64-prog:
[bitwarden_export_20260207190840.json](https://github.com/user-attachments/files/25300661/bitwarden_export_20260207190840.json)
[](url)https://www.facebook.com/ewa.laszczak.5231
#

lordgreg:
This is exactly the same issue I am currently having. 

There is no way I can use two enteprise accounts for two use-cases:
1. copilot uses ghe url1
2. pull-requests uses ghe url2

IamTheKaaZZ:
Having exactly the same issue, I cannot have both the Copilot instance running and the extension because they need different enterprice-uri links.

holgerkampffmeyervector:
We at Vector have at least 500 users who would like to see this!

IamTheKaaZZ:
Bumping this again! We have many users that would love to use this feature change!

IamTheKaaZZ:
Any updates on this?

Sardiogo:
I’d strongly support this feature.

This would be a meaningful improvement for enterprise customers. In real company environments, different GitHub related services often need independent configuration, and a single global github-enterprise.uri becomes a deployment constraint. An extension specific setting here would remove that blocker and make adoption much easier in enterprise setups.

IamTheKaaZZ:
any updates on this?

alexr00:
@TylerLeonhardt did you have any thoughts/plans on what we should do for this?

TylerLeonhardt:
This isn't something the GH PR extension can/should handle itself. It requires changes in GitHub Auth in core. 

duplicate of https://github.com/microsoft/vscode/issues/281545

PR #8647, opened 2026-04-07
Fix split enterprise auth for PRs and Copilot

IamTheKaaZZ:
Add githubPullRequests.customEnterpriseUri for PR-specific enterprise auth while preserving github-enterprise.uri as the default enterprise path.

Use extension-managed PAT auth only for the custom PR enterprise host. Keep Copilot-facing flows on the legacy enterprise auth provider. Add explicit sign-in commands for generic and custom enterprise flows. Refresh repository detection and enterprise host cache on auth/config changes. Add regression tests for custom enterprise auth, legacy Copilot auth, and enterprise sign-in routing.

Fixes this issue: https://github.com/microsoft/vscode-pull-request-github/issues/8191

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ There are several settings that can be used to configure the extension.
 
 As mentioned above, `githubPullRequests.remotes` is used to specify what remotes the extension should try to fetch pull requests from.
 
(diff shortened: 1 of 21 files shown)


IamTheKaaZZ:

alexr00:
Thank you for the PR, but we should solve this in VS Code core.

PR #8677, opened 2026-04-18
feat: add setting to control editor close behavior after marking file as viewed

### LINKED ISSUE #5092: Option to *not* close file marked as viewed via keybinding

sungwoonpark0502:
Fixes #5092

## Problem
When `pr.markFileAsViewed` is triggered via keybinding or editor menu bar, the active editor closes automatically. This is surprising to users who want the file to stay open after marking it as viewed.

## Solution
Add a new configuration setting `githubPullRequests.closeAfterMarkAsViewed` (boolean, default: `false`).

- When `false` (default): editor stays open after marking as viewed -- matches tree view behavior
- When `true`: restores the previous auto-close behavior

## Changes
- `package.json`: register new `githubPullRequests.closeAfterMarkAsViewed` setting
- `package.nls.json`: add description string
- `src/commands.ts`: wrap the tab-close logic in a check for the new setting

## Test plan
- [ ] With setting `false` (default): trigger `pr.markFileAsViewed` via keybinding -- editor should stay open
- [ ] With setting `true`: trigger `pr.markFileAsViewed` via keybinding -- editor should close
- [ ] Tree view behavior unchanged (uses `FileChangeNode` path, not affected by this setting)

### DIFF
--- a/package.json
+++ b/package.json
@@ -862,6 +862,11 @@
 					"default": "author",
 					"description": "%githubPullRequests.pullRequestAvatarDisplay.description%"
 				},
+				"githubPullRequests.closeAfterMarkAsViewed": {
+					"type": "boolean",
+					"default": false,
+					"description": "%githubPullRequests.closeAfterMarkAsViewed.description%"
+				},
 				"githubIssues.alwaysPromptForNewIssueRepo": {
 					"type": "boolean",
 					"default": false,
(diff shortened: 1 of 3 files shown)


sungwoonpark0502:
@microsoft-github-policy-service terminate

PR #8679, opened 2026-04-20
Fix: Re-parent pulled-up directory children to ensure proper checkbox refresh

### LINKED ISSUE #8646: Reopening - Parent folder checkbox not auto-checked when all children are marked as viewed
Reopening #8584 

- Extension version: 0.134.0
- VSCode Version:1.114.0
- OS: Windows

Steps to Reproduce:
I'm on v0.134.0 and it seems I still have the issue.
In the screenshot below, `projects.e2e-spec.ts` is the only file in the `projects` folder. When ticking `projects.e2e-spec.ts`, the checkbox in `projects` used to be automatically ticked as well, which is the intended behaviour since all files in the folder are now checked.

Image

MaxDNG:
Fixes #8646

This pull request addresses a bug where directory tree nodes that are promoted to the parent node (when the root directory label is empty) were not correctly updating their parent references. The changes ensure that child nodes have their `parent` property set correctly when they are moved up in the tree. This improves the consistency and correctness of the tree structure in the UI.

**Tree node parenting fixes:**

* In `commitNode.ts`, when flattening root directory nodes, each child’s `parent` is now set to the current node before being promoted to the parent.
* In `filesCategoryNode.ts`, after flattening root directory nodes, each child’s `parent` is set to the current node to maintain correct parent relationships.
* In `pullRequestNode.ts`, when promoting children of a root directory node, each child’s `parent` is set appropriately before being added to the parent.

### DIFF
--- a/src/test/view/treeNodes/directoryTreeNode.test.ts
+++ b/src/test/view/treeNodes/directoryTreeNode.test.ts
@@ -206,4 +206,57 @@ describe('DirectoryTreeNode', function () {
 			assert.strictEqual(rootDir.checkboxState?.state, vscode.TreeItemCheckboxState.Unchecked);
 		});
 	});
+
+	describe('parent pointer after phantom-root pull-up', function () {
+		// When a tree has multiple top-level directories, a temporary DirectoryTreeNode
+		// with label='' (phantom root) is used to build the tree, then its children are
+		// pulled up to the actual container. If children are not re-parented, getParent()
+		// still returns the phantom DirectoryTreeNode and processCheckboxUpdates fires
+		// refresh() on an invisible node, so checkbox state never updates visually.
+
+		it('getParent() returns undefined after child is re-parented to the container', function () {
+			const container = createMockParent();
+			const phantomRoot = new DirectoryTreeNode(container, '');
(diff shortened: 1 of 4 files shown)


MaxDNG:

Eiji-Kudo:
Confirmed working locally on top of current `main` (v0.142.0)!

https://github.com/user-attachments/assets/49affa9f-44db-4ccb-a9fa-fa4f5ee0db87




Cherry-picked the two commits from this PR onto `origin/main`, built and installed the VSIX. After reload, parent folder checkboxes now auto-check correctly when all child files are marked as viewed, and the "Refresh" workaround from #8646 is no longer needed.

Hope this gets merged soon — thanks for tracking down the root cause (the missing `parent` re-assignment after the phantom-root pull-up).

alexr00 (approved):
Thank you for the fix!

alexr00 (approved):
/AzurePipelines run

TylerLeonhardt (approved):

Eiji-Kudo:
@alexr00 @TylerLeonhardt Thank you so much for reviewing this!
So much appreciated!

PR #8686, opened 2026-04-21
Add base and compare branch names to Copilot PR title and description generation

### LINKED ISSUE #8662: Missing Source Branch Name in `pullRequestDescriptionGeneration`
### Description 
When using the `pullRequestDescriptionGeneration` feature, the generated content does not seem to have access to the source (merge) branch name. 

### Reproduction 
Tried multiple prompts to explicitly extract the branch name. For example: 
`generate pull request description with the following content: 
write the current branch name (the branch going to be merged to base)` 

### Expected Behavior 
The generated description should include the actual source branch name (the branch being merged into the base branch). 

### Actual Behavior 
- Returns a generic or suggested branch name 
- Or ignores the request entirely 

### Notes 
- Prompt is intentionally simple and explicit 
- Behavior suggests the source branch name is not available in the generation context 

### Suggested Improvement 
Expose the source (merge) branch name in the `pullRequestDescriptionGeneration` context so it can be reliably used in prompts.

shaypet:
Adds `baseBranch` and `compareBranch` to the context passed to `TitleAndDescriptionProvider.provideTitleAndDescription`, and forwards them from all three call sites where title and description generation is triggered.

Previously the model had no knowledge of which branches were involved, causing it to hallucinate branch names in the generated description. With this change, the actual branch names are passed to the provider so they can be included in the prompt.

### Changes

- `src/api/api.d.ts`: Added `baseBranch?` and `compareBranch?` to the `TitleAndDescriptionProvider` context interface.
- `src/github/createPRViewProvider.ts`: Passes `this.model.baseBranch` and `this.model.compareBranch` when calling `provideTitleAndDescription` from the Create PR view.
- `src/github/pullRequestOverview.ts`: Passes `this._item.base.ref` and `this._item.head?.ref` when regenerating a description from the PR overview.
- `src/commands.ts`: Passes `args.baseBranch` and `args.compareBranch` when calling through the `github.createPullRequest.generate` command API.

Fixes #8662

The corresponding change to the Copilot Chat extension that uses these values to render them in the prompt is at microsoft/vscode-copilot-chat#5083.

### DIFF
--- a/src/api/api.d.ts
+++ b/src/api/api.d.ts
@@ -256,7 +256,7 @@ export interface IGit {
 }
 
 export interface TitleAndDescriptionProvider {
-	provideTitleAndDescription(context: { commitMessages: string[], patches: string[] | { patch: string, fileUri: string, previousFileUri?: string }[], issues?: { reference: string, content: string }[], template?: string }, token: CancellationToken): Promise<{ title: string, description?: string } | undefined>;
+	provideTitleAndDescription(context: { commitMessages: string[], patches: string[] | { patch: string, fileUri: string, previousFileUri?: string }[], issues?: { reference: string, content: string }[], template?: string, baseBranch?: string, compareBranch?: string }, token: CancellationToken): Promise<{ title: string, description?: string } | undefined>;
 }
 
 export interface ReviewerComments {
--- a/src/api/api.d.ts
+++ b/src/api/api.d.ts
@@ -256,7 +256,7 @@ export interface IGit {
 }
 
 export interface TitleAndDescriptionProvider {
-	provideTitleAndDescription(context: { commitMessages: string[], patches: string[] | { patch: string, fileUri: string, previousFileUri?: string }[], issues?: { reference: string, content: string }[], template?: string, baseBranch?: string, compareBranch?: string }, token: CancellationToken): Promise<{ title: string, description?: string } | undefined>;
+	provideTitleAndDescription(context: { commitMessages: string[], patches: string[] | { patch: string, fileUri: string, previousFileUri?: string }[], issues?: { reference: string, content: string }[], template?: string, compareBranch?: string }, token: CancellationToken): Promise<{ title: string, description?: string } | undefined>;
 }
 
 export interface ReviewerComments {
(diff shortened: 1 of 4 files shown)


shaypet:

src/api/api.d.ts
```diff
@@ -256,7 +256,7 @@ export interface IGit {
 }
 
 export interface TitleAndDescriptionProvider {
-	provideTitleAndDescription(context: { commitMessages: string[], patches: string[] | { patch: string, fileUri: string, previousFileUri?: string }[], issues?: { reference: string, content: string }[], template?: string }, token: CancellationToken): Promise<{ title: string, description?: string } | undefined>;
+	provideTitleAndDescription(context: { commitMessages: string[], patches: string[] | { patch: string, fileUri: string, previousFileUri?: string }[], issues?: { reference: string, content: string }[], template?: string, baseBranch?: string, compareBranch?: string }, token: CancellationToken): Promise<{ title: string, description?: string } | undefined>;
```
alexr00: Why is the name of the base branch useful when generating a title + description for a PR?

shaypet: My original intent was mainly around the compare (merge/source) branch, not the base branch.

The idea for this change started specifically with the compare branch, for two main reasons:

- In personal use, I give branches informative names and use them to help generate a more accurate PR title and description.

- At work, we include the JIRA ticket ID in the branch name. When I tried to generate a PR description that links to the ticket (by extracting it from the branch name via the pullRequestDescriptionGeneration prompt), I got guesses/hallucinations instead of the actual value.

### **Regarding the base branch:** 
my assumption was that, just like we pass the merge branch, all associated information should already be available in the prompt context.

Personally, I don’t see a strong need for it, but since I found value in exposing the compare branch, I thought others might benefit from having the base branch available as well.

If the community doesn’t see a real use for it, I can simply remove it from this PR.

alexr00: We don't have strict API review for API exposed by this extension, but, generally speaking, when adding API we only want to add what we know is needed. I would prefer we do not include `base` as we don't have a use case for it.

The reasoning behind including the merge branch is clear to me, and I support it's addition.

alexr00 (changes requested):

src/api/api.d.ts
shaypet: changes where made. base branch removed from context in both repo (this and copilot)

i would appreciate if you can help with accelerating https://github.com/microsoft/vscode-copilot-chat/pull/5083 review.

thank you.

alexr00 (approved):

jrieken (approved):

PR #8761, opened 2026-05-29
feat: `contributes.yamlValidation`

Malix-Labs:
I do no know if there would be another better URL than from https://json.schemastore.org/, feel free to suggest

### DIFF
--- a/package.json
+++ b/package.json
@@ -869,6 +869,16 @@
 				}
 			}
 		},
+		"yamlValidation": [
+			{
+				"fileMatch": ".github/ISSUE_TEMPLATE/!(config).{yml,yaml}",
+				"url": "https://json.schemastore.org/github-issue-forms.json"
+			},
+			{
+				"fileMatch": ".github/ISSUE_TEMPLATE/config.{yml,yaml}",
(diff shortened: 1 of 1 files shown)


alexr00 (approved):
Cool, thank you!

Giuspepe (approved):

Malix-Labs:
yw @alexr00 <3

also please read https://github.com/microsoft/vscode-pull-request-github/pull/8761#issue-4549952764

\+ I strongly recommend you test it in your end to see if it works by default without needing any more extensions, since my setup is highly customized and have many extensions and non-default settings

there might be a better URL I'm not (or even cannot be) aware of

alexr00:
Looks like we already allow schemastore.org: 

https://github.com/microsoft/vscode/blob/25afc315fe0f791309594255cc27a73b43a2ae95/extensions/json-language-features/package.json#L137

Malix-Labs:
@alexr00 yes indeed, but since Microsoft owns both GitHub Issues and thus its schema, isn't there a a static file in this repository / another Microsoft repository which could provide this yml schema as a path instead of relying on that external CDN?

Issue #4315, opened 2022-12-11
Commit wise reviews


ldelossa:
Hello, 

Currently there is poor support for the common use case of commit-wise reviewing. 

In a commit-wise review, each commit should be viewable and each file within a commit should be able to have comments applied to them. 

In addition, when a commit is being viewed, the local repository should be checked out to that commit object. This allows for the editor not to force read-only on the file. Forcing read-only stops the use of any LSP tools and the ability to navigate the changes that occurred within a PR. 

I'd like to see better support for commit-wise reviews with the above taken into account.

davidrinnan:
Especially given that in a scenario where: 
- branch A from master, commits changes, creates PR-A 
- branch B from branch A, commits changes, creates PR-B 

Reviewer review PR-B, approves and merges with master. 

It will now look like PR-A has been merged without being reviewed. Even though I have reviewed everything in PR-B which includes PR-A.

TheElectronWill:
GitHub's web UI now supports this (kinda): https://github.blog/changelog/2025-12-11-review-commit-by-commit-improved-filtering-and-more-in-the-pull-request-files-changed-public-preview/#%f0%9f%94%8d-review-commit-by-commit

Is there any plan to move forward with this feature on the extension side? I may be able to contribute :)

Issue #2793, opened 2021-06-23
Can't add pull request review inline comments in files under Commits


simonmichael:
Issue Type: <b>Bug</b>

Enter Review Mode for a pull request. Expand the Commits tree in the sidebar, open a commit, open one of its files, scroll to a diff hunk. Mouse over the + in the right pane's gutter, or the - in the left pane's gutter. When you do this in a file under the Files tree, you can click to create an inline comment. Under the Commits tree, it is not clickable. 

This makes it difficult to review by commit.

Extension version: 0.27.1
VS Code version: Code 1.57.1 (Universal) (507ce72a4466fbb27b715c3722558bb15afa9f48, 2021-06-17T13:28:32.912Z)
OS version: Darwin arm64 20.3.0
Restricted Mode: No

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M1 (8 x 24)|
|GPU Status|2d_canvas: enabled<br>gpu_compositing: enabled<br>metal: disabled_off<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>rasterization: enabled<br>skia_renderer: disabled_off_ok<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|2, 3, 3|
|Memory (System)|16.00GB (0.09GB free)|
|Process Argv|--crash-reporter-id 02db353d-4821-41f7-a6df-92ed86181e77|
|Screen Reader|no|
|VM|0%|
</details>

ELLIOTTCABLE:
This one is super-important to my workflow. For large, complex PRs, viewing changes "intermingled" is a nightmare — as long as your collaborators are disciplined, it's much easier to view the changes *chronologically*, which allows you to follow the development of the changes and stay in sync with the author's intentions along the way.

tizzyapunkt:
Right now, it's at least possible to add comments while viewing the latest commit and to the files this commit touched. Might be coincidence though.

PeterBenc:
This one together with #3301 is crucial for my workflow, I am kind of sad to see in `on deck` or in `backlog`. Not sure how others review PRs but if you can't review by commits they might as well be squashed. I am literally waiting for this one feature to migrate from the web UI. :disappointed:

ldelossa:
Yeah. This is basically what I'm waiting for to actually use this as a code review tool. 

[...]

PR #8867, opened 2026-07-30
Allow commenting on files from earlier commits of a pull request

artur-czuba-iyuno:
Refs #4315, #2793

Files opened from the Commits node get a `review` URI carrying the sha of that commit. `findMatchedFileChangeForReviewDiffView` only ever compares that sha against the pull request head and the merge base, so any other commit fails to match, no commenting ranges are produced, and the gutter stays empty. This is why commenting has always happened to work on the newest commit and nowhere else, as observed in #2793.

Comments are created through `addPullRequestReviewThread`, which takes only `path`, `line` and `side` — no commit id — so a new thread is always anchored to the head of the pull request. Simply relaxing the sha comparison would place comments at the same line number in the head diff, which is silently wrong for any file that changed after the commit being viewed.

This change therefore allows commenting on a commit's file only when the file is unchanged between that commit and the head. In that case the document's line numbers are by definition the head's line numbers, so no position translation is needed and the comment lands exactly where it was written — matching what the GitHub web UI does when reviewing changes.

* `provideCommitCommentingRanges` provides ranges from `diffBetween(base.sha, commit, file)`, so every offered line is part of the pull request diff and is accepted by the mutation.
* When the file has changed since the commit, commenting is disabled and the file gets a warning icon in the Commits tree. Hovering explains why, names the most recent commit that touched the file, and offers a link to open the version of the file that can be commented on. The git lookup happens in `resolveTreeItem`, so it only runs for the file actually hovered; detecting which files changed costs one `diffBetween` per commit node, not one per file.
* The thread reconciliation in `_pendingCommentThreadAdds` is made URI-aware. Without it the newly created thread fails to match the incoming one and gets duplicated.

### Testing

New tests in `src/test/view/reviewCommentController.test.ts` cover the three cases: ranges are provided for a file that has not changed since the commit, commenting is disabled for a file that has changed, and a document of the head commit is left to the existing code path.

Verified manually against a checked out pull request, including the warning icon, its hover, and the link to the latest version of the file.

### Not covered here

Anchoring a thread to a specific commit via the REST API (`commit_id`), which would also enable commenting on files that changed later and let GitHub mark such threads as outdated. That is the larger part of #4315 and would add a second code path for creating comments next to the GraphQL one. Happy to follow up with it if that is the direction you would prefer.

Added files under a commit need #8866 as well, otherwise the head side of their diff is the base document and this code path does not apply to them. The two changes are independent.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -396,6 +396,34 @@ export function registerCommands(
 		),
 	);
 
+	// Used from the tooltip of a file that has changed since the commit being viewed, to get to the
+	// version of that file that can actually be commented on.
+	context.subscriptions.push(
+		vscode.commands.registerCommand(
+			'pr.openCurrentFileChange',
+			async (args: { rootPath: string; fileName: string } | undefined) => {
+				if (!args?.fileName) {
+					return;
+				}
+
+				const reviewManager = reviewsManager.reviewManagers.find(
+					manager => manager.repository.rootUri.path === args.rootPath,
+				);
+				const fileChange = reviewManager?.reviewModel.localFileChanges.find(
+					change => change.fileName === args.fileName,
+				);
+				if (!fileChange) {
+					return vscode.window.showInformationMessage(
+						vscode.l10n.t('{0} is not part of the changes of the checked out pull request.', args.fileName),
+					);
+				}
+
+				await fileChange.resolve();
+				return vscode.commands.executeCommand(fileChange.command.command, ...(fileChange.command.arguments ?? []));
+			},
+		),
+	);
+
 	context.subscriptions.push(
 		vscode.commands.registerCommand('pr.deleteLocalBranch', async (e: PRNode) => {
 			const folderManager = reposManager.getManagerForIssueModel(e.pullRequestModel);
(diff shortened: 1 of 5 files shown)


Angie20182 (commented):
Request Id: 39e943e2-aed1-44e2-a137-a3e4b7000200
Correlation Id: e94a2ca2-30b0-1000-a157-f18c22c7e480
Timestamp: 2026-07-30T10:11:44Z
Message: AADSTS50020: User account 'loveemily22@gmail.com' from identity provider 'live.com' does not exist in tenant 'Microsoft' and cannot access the application '00000003-0000-0ff1-ce00-000000000000'(Office 365 SharePoint Online)

alexr00:
Thank you for the PR!

> This change therefore allows commenting on a commit's file only when the file is unchanged between that commit and the head. In that case the document's line numbers are by definition the head's line numbers, so no position translation is needed and the comment lands exactly where it was written — matching what the GitHub web UI does when reviewing changes.

This sounds confusing for the user. We go from "comments not allowed in commit diffs" to "comments sometimes allowed in commit diffs". I would prefer to leave as is.

Issue #4618, opened 2023-03-13
 Option to Choose Base for Diff


sedghi:
Currently, when performing a review in GitHub, the first round shows the changed files against the upstream base. However, in subsequent rounds of review, the diff is based on the previous commit, making it difficult to compare changes against the latest upstream base.

Is there anywhere that I can choose the base for the diff?

alexr00:
You can show the "Changes since last review" using the action in the tree view, but there's currently no way to choose a specific base.

sedghi:
Thanks for the reply.
Is there a reason why I don't see that option?

![image](https://user-images.githubusercontent.com/7490180/224722239-300034a3-21b4-4ffb-9ce0-4d2c323f15af.png)

alexr00:
"Changes since last review" changes the base for the whole PR you have checked out (though an individual file option is a good idea), so it's on the root of the tree in your screenshot. For non-checked out PRs it's on the description node

![image](https://user-images.githubusercontent.com/38270282/224725068-715c842d-d071-4602-bc2f-24da1a87d409.png)

sedghi:
Oh I didn't see it there, and honestly didn't expect it there either. Thanks! it works great as I see now, but option to choose the base would be awesome

bogdando:
That's a nice to have, indeed! When there is an update pushed over a pull request, resulting in messages like:

``force-pushed the foo branch from <sha1> to <sha2> ... Compare``
It would be a much welcome feature to get that difference in the addon interface

bogdando:
regarding implementation, that could be done by tracking multiple local branches for each PR, like gerrit revisions for a change-set, then picking the previous revision (another pr-xxx branch) as a new base for the diff

benny5555:
I have a proposed implementation in #8968 and would like to check the UI direction here first, since that is the part most likely to need adjusting.

**Design**

- A **Select Diff Range...** action (`pr.selectDiffRange`) on pull request nodes and the description node (inline `$(git-compare)` icon plus context menu entry) and in the "Changes in Pull Request" view title. It is not in the command palette.
- It opens a quick pick with:
  - **All changes** (merge base to head, the default)
  - **Since my last review** (the existing "Show Changes Since Last Review" behaviour)
  - **Since my last comment**: the commit my newest review comment was written against
  - **Since the earliest unresolved thread**
  - **Custom...**: a second quick pick listing the pull request's commits, newest first, plus "Merge base"
[...]

PR #8967, opened 2026-09-19
Resolve the folder a pull request belongs to when open folders share a remote

### LINKED ISSUE #8969: Approve and resolve act on the wrong folder when a repository and its worktrees are open in one window
- Extension version: 0.166.1
- VSCode Version: 1.138.0 (Windows client, WSL 2 remote)
- OS: Windows 11 host, Ubuntu on WSL 2 running the extension host
- Repository Clone Configuration: single repository, with eight git worktrees under `worktrees/` opened as additional repositories via `git.autoRepositoryDetection: "subFolders"` and `git.scanRepositories`
- GitHub Product: GitHub.com

When several open folders are checkouts of the same GitHub repository (a repository plus its worktrees, each on a different pull request branch), two actions go to the wrong folder:

**1. Approving from the pull request overview fails although the checkout is in sync.**
Opening a description resolves the folder manager by owner and repository name (`RepositoriesManager.getManagerForRepository`) and takes the first folder with that remote, not the folder the clicked node belongs to. The overview panel is bound to that folder, so `PullRequestModel.approve` compares another worktree's checked out `HEAD` with the pull request's remote head and rejects with:

> Submitting review failed. The remote head of the pull request branch has changed. Please pull the latest changes from the remote branch before approving.

Which folder is "first" depends on repository discovery order, so the same setup can work in one window and fail after a reload. Approving from the sidebar's active pull request view works, because `ReviewManager` passes its own folder manager.

**2. Resolving a comment thread on the left side of a diff appears to do nothing.**
Every `ReviewCommentController.hasCommentThread` returns `true` for any `review:` scheme thread, and `resolveCommentHandler` picks the handler with the longest root path, so a worktree's controller handles a thread from the main repository. It calls `resolveReviewThread` on its own pull request model, which does not know the thread. The GitHub mutation still runs (the thread is resolved on github.com), but no change event fires and the editor never updates.

Steps to Reproduce:

1. Open a repository root as the workspace folder. Add worktrees under `worktrees/2`, `worktrees/3`, ... and list them in `git.scanRepositories` so the git extension opens them; check out a different open pull request branch in each.
2. In the "Changes in Pull Request" view of a worktree that is not the first discovered folder, click the description node and press **Approve** in the overview. Observe the "remote head ... has changed" error even though `git status` shows the branch in sync with its upstream.
3. In the same window, open a file diff from the main repository's changes view and click **Resolve Conversation** on a thread shown in the left editor. Observe that nothing changes in the editor, while the thread is resolved on github.com.

Fix proposed in #8967: resolve the folder from the pull request model (the folder whose GitHub repository owns the model, then the folder that has the pull request checked out, then the remote match), and have review controllers only claim `review:` threads whose recorded root path is their own repository.

benny5555:
## Summary

- `RepositoriesManager.getManagerForIssueModel` now prefers the folder whose GitHub repository owns the model, then the folder that has the pull request checked out, and only then falls back to the first folder with a matching remote. `FolderRepositoryManagerResolver` exposes the same lookup, and the open-description and chat-session commands use it.
- `ReviewCommentController.hasCommentThread` only claims `review:` scheme threads whose recorded root path is the controller's own repository, instead of every `review:` thread in the window.

## Problem

When a repository and several of its worktrees are open in one window, every folder has the same GitHub remote and two things go wrong:

1. Opening a pull request description resolved the folder manager by owner and repository name and took the first match, so the overview panel could be bound to a different worktree than the node that was clicked. Approving from that panel then compared the other worktree's checked out `HEAD` with the remote head and failed with "The remote head of the pull request branch has changed. Please pull the latest changes from the remote branch before approving." even though the clicked worktree was in sync. Which folder wins depends on repository discovery order, so the failure is intermittent across window reloads.
2. Every `ReviewCommentController` claimed every `review:` scheme comment thread, and the handler with the longest root path won. Resolving or replying to a thread on the left side of a diff therefore went through another folder's pull request model: the GitHub mutation ran, but that model did not know the thread, so nothing in the editor updated and the action appeared to do nothing.

Reproduce with a single-folder workspace at the repository root, `git.autoRepositoryDetection: "subFolders"` and `git.scanRepositories` listing `worktrees/*`, each worktree on a different pull request branch. Open a pull request description from a worktree's "Changes in Pull Request" view and click Approve, or resolve a comment thread shown on the left side of a file diff.

## Validation

- `npm run test`: 567 passing, including the new `RepositoriesManager.getManagerForIssueModel` and `ReviewCommentController.hasCommentThread` tests
- Manually in a window with a repository and eight worktrees: approving from the overview and resolving left-side threads now act on the folder that was clicked

Fixes #8969


sylvesterkaczmarek (commented):
The root-path comparison is unconditionally case-insensitive. On Linux /repo/Foo and /repo/foo can be two different repositories, so both controllers can claim the same review URI here. Can this use platform/filesystem-aware path equality instead of lowercasing on every OS, with a case-sensitive regression?

benny5555:
Addressed the Copilot note about malformed `review:` queries in 7173ae9e: a query that does not parse is not one this extension created and is no longer claimed by any controller, and a query without a root path falls back to checking the file path (which review uris keep) against the controller's repository instead of being claimed by every controller. Tests cover both cases.

benny5555:

PR #8970, opened 2026-09-19
Paginate the compare API for large diffs of a narrowed commit range

benny5555:
## Summary

- `compareCommits` takes whether the compared range covers the whole pull request. For a narrowed range (today: "Show Changes Since Last Review") with 100 or more files it now requests a full 100-file page explicitly and pages through the compare API, up to GitHub's 300-file maximum with a logged warning, instead of falling back to `pulls.listFiles`.
- On a GitHub "Server Error" it rethrows for a narrowed range instead of returning the whole pull request's files.
- The merged pull request shortcut, which also lists the whole pull request, only applies when the whole range is shown.

## Problem

`pulls.listFiles` returns the files of the whole pull request. Using it as the large-diff fallback for a compare that starts at the reviewed commit shows changes outside the selected range, so "Show Changes Since Last Review" silently reverted to all changes for pull requests with 100 or more files in the range. The same happened on merged pull requests regardless of size.

This is the first of a set of small pull requests extracted from #8968 (select a diff range). It is independent of the others.

## Validation

- `npm run test` (extension host suite)


benny5555:

PR #8971, opened 2026-09-19
Keep editors and comment threads in sync when the compared base changes

benny5555:
## Summary

Toggling "Show Changes Since Last Review" changes the base commit that file diffs are compared against, but several places kept state from the previous base:

- The raw file change cache was not cleared on toggle, so consumers such as the overview kept the previous file list.
- Open `pr:` diffs were only reopened when the head commit changed; a base-only change left stale diffs open.
- The previous "Changes in Pull Request" multi diff editor stayed open next to the newly opened one.
- `PullRequestCommentController` cached editor threads by file name and side. A document reopened with a different base commit in its uri hit the cache and got no comment threads.
- `ReviewCommentController` created left-side threads against the merge base once and never rebuilt them, so reopened left-side documents showed no comments. It now rebuilds them when the merge base changes. While there, thread creation is awaited (it was fire-and-forget) and threads created by an initialization that was superseded meanwhile are disposed instead of being left in detached arrays.

This is one of a set of small pull requests extracted from #8968 (select a diff range). It is independent of the others.

## Validation

- `npm run test` (extension host suite)
- Manually: toggle "Show Changes Since Last Review" with file diffs open; the diffs reopen against the new base and the existing threads stay visible on both sides.


benny5555:

PR #8972, opened 2026-09-19
Add a selectable diff range to the pull request model

benny5555:
## Summary

Replaces the `showChangesSinceReview` boolean on `PullRequestModel` with a `DiffRange`: a preset (all changes, since my last review, since my last comment, since the earliest unresolved thread, or a custom base commit), an optional base commit and an optional head commit. The boolean getter and setter stay as a facade over the "since my last review" preset, so the existing menus, context values and the overview "View Changes" card are unchanged.

- `getRawFileChangesInfo` picks the compare base from the range. Presets that depend on review data are resolved lazily from the review thread cache through the pure resolver in `src/github/diffRange.ts`, which has no `vscode` dependency and is covered by headless tests (`src/test/scripts/diffRange.test.ts`, 19 cases).
- A range generation guards against a compare response for a range that was replaced while the request was in flight.
- A warning is recorded on the range when GitHub compared from the merge base because the selected base is not an ancestor of the head.
- `ReviewManager` keys its "nothing changed" early-out on the range instead of the boolean.

**No user interface in this change.** The command, tree label, left-side outdated threads and setting follow in separate stacked pull requests. UI design discussion: #4618.

## Stack

This is part of a stack extracted from #8968. It is based on #8970 and #8971 (the commit list includes them); **the commit to review is the last one**.

3. **this**: diff range model
4. command and quick pick
5. tree label
6. outdated threads on the base side

## Validation

- `npm run compile:test && npm run test:scripts` (94 passing)


benny5555:

PR #8973, opened 2026-09-19
Add the "Select Diff Range..." command and quick pick

benny5555:
## Summary

Adds `pr.selectDiffRange`, offered inline and in the context menu of pull request and description nodes and in the "Changes in Pull Request" view title (not in the command palette). It opens a quick pick with the range presets (all changes, since my last review, since my last comment, since the earliest unresolved thread) and a "Custom..." entry that picks the base commit from the pull request's commits, newest first, plus "Merge base". Presets without a candidate commit are listed under an "Unavailable" separator with the reason. `pr.showDiffAll` and `pr.showDiffSinceLastReview` accept pull request nodes as well as the description node.

Localized labels for presets, unavailable reasons and warnings live in `src/common/diffRangeLabels.ts`; the quick pick in `src/view/diffRangeQuickPick.ts`. UI design discussion: #4618.

## Stack

Part of the stack extracted from #8968; based on #8972. **The commit to review is the last one.**

1. #8970 compare pagination for narrowed ranges
2. #8971 editors and comment threads follow a base change
3. #8972 diff range model
4. **this**: command and quick pick
5. tree label
6. outdated threads on the base side
7. #8968 default setting (top of the stack)

## Validation

- `npm run compile`, `npm run lint`, `npm run hygiene`, `npm run check:commands`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8974, opened 2026-09-19
Show the selected diff range on the pull request and description nodes

benny5555:
## Summary

The description node and pull request nodes show the active range in their description, e.g. `5ad2698..d2a9384 · since earliest unresolved thread`, with a tooltip warning when the base is not an ancestor of the head. A `:customRange` context value marks ranges other than the default and "since my last review", so the existing "Show All Changes" action appears inline to reset the range. Reopening the first diff after a range change applies to every non-default range. UI design discussion: #4618.

## Stack

Part of the stack extracted from #8968; based on #8973. **The commit to review is the last one.**

3. #8972 diff range model
4. #8973 command and quick pick
5. **this**: tree label
6. outdated threads on the base side

## Validation

- `npm run compile:test` (updated `repositoryChangesNode` test)


benny5555:

PR #8975, opened 2026-09-19
Show outdated review threads on the base side of a selected diff range

benny5555:
## Summary

An outdated thread has no line on the current head, so it was never shown on a file diff, only on the separate "commit~" entries. With a non-default diff range the left side of the diff is the commit such a thread was written against (or a later one), so the thread has a well defined position there and shows why a line was later changed.

Both comment controllers add a copy of each outdated thread to the left-side document at the thread's original lines. In review mode the lines are mapped through the local `git diff` when the range base differs from the commit the thread was written against; threads that cannot be mapped are skipped. Changes to and removal of a thread are applied to the copies too. The default all-changes view is unchanged.

## Stack

Part of the stack extracted from #8968; based on #8974. **The commit to review is the last one.**

3. #8972 diff range model
4. #8973 command and quick pick
5. #8974 tree label
6. **this**: outdated threads on the base side

## Validation

- `npm run compile`, `npm run lint`, `npm run hygiene`
- Manually on a pull request with 10 threads written against an earlier commit: with "since the earliest unresolved thread" selected, the threads appear in the left editor at their original lines, and resolving one updates both copies.


benny5555:

